PR #8, opened 2019-03-05
Runtime api upgrade

bschwinn:


pbaize (approved):

tgoc99 (changes requested):
Could you do quick update to ReadMe as well with the update?  Just point at different docs, use "Channels" instead of "service", etc.

pjbroadbent (approved):

PR #13, opened 2019-03-13
[SERVICE-393] Updating FDC3 Demo UI

ottah:
- Removed `I` and `e` prefixes from interfaces and enums.
- Added `tslint-react` for JSX/TSX linting.
- Rewrote react components to use modern standards and simplify complexity.


src/demo/apps/DialerApp.tsx
```diff
@@ -1,114 +1,91 @@
...
+interface AppProps {
     phoneNumber?: string;
 }
 
-interface IAppState {
-    inCall: boolean;
-    phoneNumber: string;
-    pendingCall: ContactPayload|null;
-}
+// tslint:disable-next-line:variable-name
```
tomrobertsOF: Are we really going to need a ts-lint exception before every component declaration in the whole service?

src/demo/apps/DialerApp.tsx
```diff
@@ -1,114 +1,91 @@
...
-        } else {
-            this.setState({pendingCall: null});
-        }
-    }
-
-    private handleIntent(context: ContactPayload, startCall: boolean): void {
-        const phoneNumber: string = context.id.phone!;
+        // Cleanup
+        return () => {
+            dial.unsubscribe();
```
tomrobertsOF, line 72: My lack of React knowledge is coming out, but when would this cleanup happen? Is it sensible to create/destroy the listeners every time? 

They are stored in memory, so would get cleared on a page navigate/reload anyway, and I can't think of a reason we would still be on the page but not want the listeners.

src/demo/ContextMenuPopup.tsx
```diff
@@ -0,0 +1,113 @@
...
+export const ContextMenuPopup: React.FunctionComponent<ContextMenuPopupProps> = (props) => {
+    const [menuItems, setMenuItems] = React.useState<MenuItem[]>([]);
+    let windowHeight = 2;
+
+    React.useEffect(() => {
+        const handleMenuMessage = (message: ContextMenuMessage) => {
+            if (message.action === "update") {
+                setMenuItems(message.items);
+                const ofWindow: fin.OpenFinWindow = fin.desktop.Window.getCurrent();
+                ofWindow.isShowing((showing) => {
```
rypete: We should look to switch over any v1 API calls to v2

src/demo/apps/DialerApp.tsx
ottah: Can add the following to the `tslint.json`?:
```
"variable-name": {
            "options": [
                "allow-pascal-case"
            ]
        },
```

But that will allow pascal case everywhere.

src/demo/ContextMenuPopup.tsx
ottah: I'll change calls to the v1 API in the other files but leave the ContextMenu as they currently are.  I will replace the context menu with a different one I'm writing based on the POC you sent me.

pjbroadbent: Where's the other PoC?

src/demo/apps/DialerApp.tsx
pjbroadbent: Can we make these declarations functions, rather than constants? As in `export function DialerApp(props: AppProps) {`?

Firstly, because... well, they are actually functions (rather than variables/constants). But should also remove the need for the linter exception?..

src/demo/apps/DialerApp.tsx
ottah: The `React.useEffect` hook is called when the component is mounted (`ComponentDidMount`), the cleanup happens when the component is unmounted (`ComponentWillUnmount`). In this case it doesn't matter as the window closes the listeners are destroyed. But if the component was alongside another component and removed/hidden and the window remained open, it would remain subscribed even when the component isn't visible.

src/demo/ContextMenuPopup.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    break;
+                case ContextMenuItemType.BUTTON:
+                    component = <button key={item.id} className={"w3-button" + (item.hasChildren ? " children" : "")} onClick={handleClick}>{item.caption}</button>;
+                    break;
+                default:
+                    component = <></>;
+                    itemHeight = 0;
+            }
+
+            windowHeight += itemHeight;
```
bryangaleopenfin: Can I check this is safe and/or 'The React Way'? Do we know for certain that this method will get called every time windowHeight is reset to 2 at the start of the component function?

ottah: You're right, it doesn't get reset to 2. Fixed it.

src/demo/apps/ChartsApp.tsx
```diff
@@ -1,62 +1,59 @@
...
+            throw new Error("Invalid context received");
+        }
+    }
 
+    React.useEffect(() => {
         document.title = "Charts";
-        this.state = {
-            symbolName: props.symbolName || "AAPL"
-        };
+    });
```
bryangaleopenfin: As much for my understanding of React as anything else - would it make sense to pass in [] as the second parameter here, as we're only going to want to execute this once. There's a few similar cases where we're setting the document title

src/demo/apps/DialerApp.tsx
rypete: It's the "react" way to do fat arrow functions for FC.  However, as Phil noted it, making them old school functions will fix the linting errors as we have seen before.  We should probably go this route to avoid sprinkling tslint disables everywhere. 

Something to note is that you will lose the `React.FunctionComponent<...>` type.  You will need to assign a type directly to `props` and if you use children, it'll be `props: React.PropsWithChildren<PropType>`.  Maybe more will be needed to get `useEffect` to behave.

src/demo/components/common/ColorLinker.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+
+
+interface ColorLinkerProps {
+  selected: string;
+  onColorChange?: (color: string) => void;
```
bryangaleopenfin, line 6: Is it expected that this doesn't get used at all?

src/demo/components/common/ContextMenu.tsx
```diff
@@ -148,46 +148,46 @@ interface IContextMenuProps {
  * known at the time the page is created, the context menu definition can use a promise to specify menu items (or a 
  * subset of menu items) asynchronously.
  */
-export class ContextMenu extends React.Component<IContextMenuProps> {
+export class ContextMenu extends React.Component<ContextMenuProps> {
```
bryangaleopenfin: How come we're keeping this one as class-style rather than function-style?

src/demo/components/dialer/Number.tsx
```diff
@@ -2,61 +2,33 @@ import * as React from 'react';
...
-    inCall?: boolean;
-    number?: string;
-    handleChange?: (num: string) => void;
-}
-
-interface INumberState {
+interface NumberProps {
     inCall: boolean;
     number: string;
+    onValueChange?: (num: string) => void;
```
bryangaleopenfin, line 8: Is it worth standardizing the naming of callbacks included in the props? We sometimes have `handleFoo` and sometimes `onFoo` right now

webpack.config.js
```diff
@@ -108,21 +108,21 @@ const manifestPlugin = new CopyWebpackPlugin([{
  * 
  * This embeds the package version into the source file as a string constant.
  */
-const versionPlugin = new webpack.DefinePlugin({PACKAGE_VERSION: `'${version}'`});
+const versionPlugin = new webpack.DefinePlugin({ PACKAGE_VERSION: `'${version}'` });
```
bryangaleopenfin, line 111: How come the formatting change (spaces inside braces) here?

src/provider/index.ts
```diff
@@ -117,7 +117,7 @@ export interface IQueuedIntent {
      * application - selector will be null until the intent reaches the front of
      * the queue.
      */
-    selector: fin.OpenFinApplication|null;
+    selector: fin.OpenFinApplication | null;
```
bryangaleopenfin, line 120: The spaces around `|` is inconsistent with layouts - we worrying about that later?

bryangaleopenfin (approved):
Looks good to me. Have some questions/concerns, but assume they can be resolved without problem.

There's some inconsistency in formatting with spacing inside braces, and spacing around | when declaring union types (at least when compared to layouts), but not sure if I should worry about that for a later linting pass.

src/demo/apps/DialerApp.tsx
pjbroadbent: Looking at the definition of `FunctionComponent`, it's a function that returns a `ReactElememt`.

Should be strongly-typable as so:
```
export function DialerApp(props: AppProps): React.ReactElement<AppProps> {
    // . . .
}
```

src/demo/components/common/ColorLinker.tsx
ottah: The color linker component was just a test for when the context channel gets added. Not being used yet.

src/demo/components/common/IntentButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  DEFAULT,
+  SPIN,
+  ERROR
+}
+
+export function IntentButton(props: IntentButtonProps): React.ReactElement {
+  const {action, title, iconClassName} = props;
+  const [buttonState, setButtonState] = React.useState(ButtonStateType.DEFAULT);
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
```
pjbroadbent, line 19: Also need to `stopPropogation` on these events. Clicking one of these currently will both fire an intent and broadcast the context (which happens when clicking on a `SymbolsRow`/`ContactsRow`.

ottah: Yep fixed that in my in another branch I'll be merging in tonight.

src/demo/apps/BlotterApp.tsx
```diff
@@ -1,48 +1,33 @@
...
+    {"name": "NVDA"},
+    {"name": "PYPL"},
+    {"name": "QCOM"},
+    {"name": "ROKU"},
+    {"name": "TSLA"}
+];
+
+export function BlotterApp(): React.ReactElement {
+    React.useEffect(() => {
         document.title = "Blotter";
```
MichaelMCoates: From here, not sure if it's needed: https://reactjs.org/docs/hooks-effect.html 

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the inputs array always works.

src/demo/apps/BlotterApp.tsx
```diff
@@ -1,48 +1,33 @@
...
+    {"name": "NFLX"},
+    {"name": "NVDA"},
+    {"name": "PYPL"},
+    {"name": "QCOM"},
+    {"name": "ROKU"},
+    {"name": "TSLA"}
+];
+
+export function BlotterApp(): React.ReactElement {
+    React.useEffect(() => {
```
MichaelMCoates, line 28: From here, not sure if it's needed: https://reactjs.org/docs/hooks-effect.html 

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the inputs array always works.

src/demo/components/common/ColorLinker.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function ColorLinker(props: ColorLinkerProps): React.ReactElement {
+  const {selected} = props;
+  const colors: string[] = ['red', 'blue', 'orange', 'purple'];
+  const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const {value} = event.currentTarget;
+    console.log('change');
```
MichaelMCoates: leftover `console.log`

src/demo/components/contacts/ContactsRow.tsx
```diff
@@ -1,76 +1,43 @@
...
-        this.sendIntent(event.currentTarget, true);
+        return new fdc3.Intent(fdc3.Intents.START_CALL, getContext()).send();
+    };
 
-        event.stopPropagation();
-
-        if(this.props.handleSelect) {
-            this.props.handleSelect(null);
+    const handleClick = () => {
+        if (handleSelect) {
```
MichaelMCoates, line 35: `stopPropagation`

src/demo/components/dialer/Number.tsx
```diff
@@ -2,61 +2,33 @@ import * as React from 'react';
...
-    }
-
-    private handleClear(event: React.MouseEvent<HTMLElement>): void {
-        const handler: ((key: string) => void)|undefined = this.props.handleChange,
-            input: HTMLInputElement = document.getElementsByClassName("number-input")[0] as HTMLInputElement;
-
-        if (handler) {
-            handler("");
+    };
+    const handleClear = (event: React.MouseEvent<HTMLElement>) => {
```
MichaelMCoates, line 21: `stopPropagation`

src/demo/index.tsx
```diff
@@ -19,54 +19,61 @@ import { LauncherApp } from './apps/LauncherApp';
...
 
-if (window.hasOwnProperty("fin")) {
-    uuid = fin.desktop.Application.getCurrent().uuid;
-
-    //Colours are used to simulate having multiple apps capable of handling the same intent.
-    //In this sample they are re-skins of the same application, but in a real scenario they would be completely unrelated applications.
-    color = uuid.split("-")[2];
+function App(): React.ReactElement {
+    const hasFin = window.hasOwnProperty("fin");
+    let uuid = fin.Window.me.uuid;
```
MichaelMCoates: would this error if `hasFin` is `undefined`?

src/provider/components/AppCard.tsx
```diff
@@ -1,49 +1,30 @@
...
-    private doubleClickHandler(event: React.MouseEvent<HTMLDivElement>): void {
-        const handler = this.props.openHandler;
-
-        if (handler) {
-            handler(this.props.app);
-        }
-    }
+export function AppCard(props: AppCardProps): React.ReactElement {
+    const {app, selected, openHandler, selectHandler} = props;
+    const clickHandler = (event: React.MouseEvent<HTMLElement>) => {
```
MichaelMCoates, line 15: `stopPropagation`

src/demo/index.tsx
ottah: Removed the check as it is unneeded.

src/demo/components/common/Dialog.tsx
```diff
@@ -2,46 +2,36 @@ import * as React from 'react';
...
-                        ))}
-                    </div>
-                </div>
-            );
-        } else {
-            return null;
+export function Dialog(props: DialogProps): React.ReactElement {
+    const {show, title, body, options, handleOption} = props;
+    const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+        if (handleOption) {
```
MichaelMCoates, line 17: `stopPropagation`

src/demo/components/dialer/Dialer.tsx
```diff
@@ -2,45 +2,50 @@ import * as React from 'react';
...
+interface KeyPadButtonProps {
+    value: string;
+    handleKeyPress?: (value: string) => void;
+}
 
-        if (handler) {
-            handler((event.target as HTMLButtonElement).value);
+function KeyPadButton(props: KeyPadButtonProps): React.ReactElement {
+    const {value, handleKeyPress} = props;
+    const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
```
MichaelMCoates, line 44: `stopPropagation`

src/provider/components/AppCard.tsx
MichaelMCoates: Does this not need a `stopPropagation`?

tomrobertsOF (approved):

MichaelMCoates (approved):

pjbroadbent (approved):

PR #14, opened 2019-03-13
[SERVICE-391] Update service client

tomrobertsOF:
**Do not merge until PR13 is merged. Likely lots of conflicts and should be easier to resolve in this one.**

- Updated the client to match more closely with the improvements made to the layouts service. 
- Also includes an overhaul and modularisation of the channel messaging on both the client and provider.
- Demos and provide have been updated to properly handle the client changes but otherwise left unchanged.


src/client/connection.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * @hidden
```
bryangaleopenfin, line 2: This is entirely my ignorance, but what does this do?

ottah (approved):

src/client/connection.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * The event emitter to emit events received from the service.  All addEventListeners will tap into this.
+ */
+export const eventEmitter = new EventEmitter();
+
+/**
+ * Promise to the channel object that allows us to connect to the client
+ */
+export let channelPromise: Promise<ChannelClient>;
+// Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
```
bryangaleopenfin, line 40: Worth an explicit `Todo:` or a reference to a ticket?

src/client/main.ts
```diff
@@ -0,0 +1,280 @@
...
+}
+
+/**
+ * App metadata is Desktop Agent specific - but should support a name property.
+ */
+export interface AppMetadata {
+    id: number;
+    name: string;
+    title: string;
+    manifest_url: string;
```
bryangaleopenfin, line 53: Assume it's not our fault we're breaking from camelCase convention here?

src/demo/apps/ChartsApp.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import * as fdc3 from '../../client/index';
+import * as fdc3 from '../../client/main';
 
 import '../../../res/demo/css/w3.css';
 
 import { Chart } from '../components/charts/Chart';
-import { SecurityPayload, Payload } from '../../client/context';
+import { SecurityContext, Context } from '../../client/context';
```
bryangaleopenfin: Spaces inside {} - are we worrying about linting for now?

src/provider/MetadataStore.ts
```diff
@@ -105,7 +105,7 @@ export class MetadataStore {
      */
     public update(appData: IApplication, app: Application): void {
         if (!this.appData.hasOwnProperty(appData.id)) {
-            this.appData[appData.id] = {uuid: app.identity.uuid, name: app.identity.name!, directoryId: appData.id};
+            this.appData[appData.id] = {uuid: app.identity.uuid, name: app.identity.uuid, directoryId: appData.id};
```
bryangaleopenfin, line 108: Obviously this looks wrong but I assume it's deliberate? Worth a comment here?

bryangaleopenfin (approved):
Approved with comments/Qs

src/client/connection.ts
rypete: Hides the file from `typedoc`.  I'm guessing it was left in from another project, but, should remain as it will be needed eventually when docs come.

src/client/main.ts
```diff
@@ -0,0 +1,280 @@
...
+ * // raise the intent against a particular app
+ * await agent.raiseIntent(appIntent.intent.name, context, appIntent.apps[0].name);
+ * ```
+ */
+export async function findIntent(intent: string, context?: Context): Promise<AppIntent> {
+    return tryServiceDispatch(APITopic.FIND_INTENT, {intent, context});
+}
+
+/**
+ * Find all the avalable intents for a particular context.
```
rypete: available*

rypete (approved):
Looking good

src/provider/MetadataStore.ts
tomrobertsOF: An appIdentity doesn't have a name. We need a full identity to send it messages. Main window of an app has name===uuid

src/demo/apps/ChartsApp.tsx
tomrobertsOF: Current versions of GTS don't work properly with tsx files. Will need to be dealt with, but not part of this story.

src/client/main.ts
tomrobertsOF: I just copied this from a different location, but probably comes from the runtime.

src/client/connection.ts
tomrobertsOF: I'd say it's not strictly a TODO, but just an explanation for why we need that specific guard. I don't think there is any harm in keeping it there.

src/client/connection.ts
tomrobertsOF: It was copied from the layouts project, but I left it there intentionally since that file is all the client stuff we don't want in the docs.

src/client/main.ts
tomrobertsOF: Those comments were all copy/pasted from the FDC3 spec docs :P 

I'll update, and maybe should tell them too.

src/demo/apps/ChartsApp.tsx
pjbroadbent: SERVICE-365

src/client/main.ts
pjbroadbent: This comes from the FDC3 App Directory specs.

tomrobertsOF: Thanks Phil! 👍

MichaelMCoates (approved):
Looks great, waiting for 393.

pjbroadbent (approved):

PR #17, opened 2019-03-18
[SERVICE-396] Update FDC3 app directory

bryangaleopenfin:
This PR updates our IApplication interface (now Application) to match https://fdc3.finos.org/appd-spec

- Now that we have both `appId` and `name`, this PR changes things to primarily use `appId` internally
- This PR eliminates one of our `AppMetadata` interfaces, since it was essentially a duplicate of `IApplication`


src/client/directory.ts
```diff
@@ -2,40 +2,63 @@
...
+    name: string;
+    value: string;
+}
+
+export interface DirectoryIntent {
+    name: string;
+    displayName?: string;
+    contexts: string[];
+
+    // Sepcification is ambigous on type of customConfig, so leaving as 'any'
```
pjbroadbent: Typo: `Sepcification`.

Seems it's for vendor-specific extensions, which would mean it's basically up to us to define. Will have to think about if we have any use for this.

src/provider/FDC3.ts
```diff
@@ -195,42 +195,57 @@ export class FDC3 {
...
+
+            let isJson = false;
+            try {
+                JSON.parse(appInfo.manifest);
+                isJson = true;
+            } catch (e) {
+            }
+
+            if (isJson) {
+                reject(new Error('Inline JSON manifests not supported'));
```
rypete, line 221: Could this be moved up in place of the `isJson` stuff above?

```
JSON.parse(appInfo.manifest);
reject(new Error('Inline JSON manifests not supported'));
```

src/client/directory.ts
```diff
@@ -2,40 +2,63 @@
...
- * Type definition that is used wherever the API expects the name of an application to be passed.
+ * Type definition that is used wherever the API expects the ID of an application to be passed.
  *
  * It is not possible for TypeScript to verify that only valid application names are passed to the FDC3 API. This type
  * definition is more a hint to the callee, over a plain "string" argument.
  */
-export type AppIdentifier = string;
+export type DirectoryAppId = string;
+
+export interface DirectoryAppImage {
```
pjbroadbent: Maybe go with `AppDir` as the prefix?

Maps a little closer to what the directory is named within FDC3, and a bit shorter.

src/client/directory.ts
```diff
@@ -2,40 +2,63 @@
...
+    value: string;
+}
+
+export interface DirectoryIntent {
+    name: string;
+    displayName?: string;
+    contexts: string[];
+
+    // Sepcification is ambigous on type of customConfig, so leaving as 'any'
+    customConfig: any;
```
ottah: Perhaps would be better `customConfig: { [key: string]: any}`? Or another `DirectoryNameValuePair`?

ottah (approved):
Looks good

src/client/directory.ts
ottah: I agree with this, Directory- is a bit confusing for some of the interfaces/types.

src/client/directory.ts
MichaelMCoates: `ambigous`

src/provider/FDC3.ts
bryangaleopenfin: 👍

src/client/directory.ts
bryangaleopenfin: The spec really gives us no guarantees, so this could be a `string` or a `number`, at least in theory. My sense is that this should stay as `any`

src/client/directory.ts
bryangaleopenfin: 👍

src/client/directory.ts
bryangaleopenfin: Discussed with Phil - just dropping the prefix entirely

rypete (approved):

ottah:
:thumbsup:

PR #19, opened 2019-03-19
[SERVICE-405] Added pre-commit package

ottah:
- Not currently running any precommit checks because of gts/CLANG error


package.json
```diff
@@ -67,5 +68,8 @@
     "testMatch": [
       "**/test/provider/**/*.(ts|tsx)"
     ]
+  },
+  "precommit": {
+    "run": []
```
pjbroadbent: Can you add `check` here so that this is ready to go - we'll just hold off on merging until #16 is in.

rypete (approved):
Approved, but makes sense to update this  then merge it after the linting PR is in.

pjbroadbent (approved):

pjbroadbent:
With SERVICE-395/#23, linter is now fully setup and passes.

Can merge this in as soon as conflicts are resolved - @TheoOttahOF can you take a look?

PR #20, opened 2019-03-19
[SERVICE-397] Build out framework for FDC3 testing

tomrobertsOF:
Important note: this PR necessitated a runtime version update to v10 in order to get puppeteer working.

- Added the foundation of a test harness. Uses puppeteer to execute code inside the OpenFin windows themselves
- Project now has separate process for integration and unit tests (both using jest)
- Added simple test files of both types to test the test harnesses


test/demo/testUtils.test.ts
```diff
@@ -0,0 +1,35 @@
+import 'jest';
+import {OFPuppeteerBrowser, TestWindowContext} from './utils/ofPuppeteer';
+
+
+
+// Testing test utils...
+describe('basic puppeteer functionality', () => {
+  jest.setTimeout(100000);
```
pjbroadbent: Looks like there's basically one of these in every file. What's the purpose of these, assume they help with debugging?

Can this instead be set as a cli-arg in `runner.js`, or in one of the Jest config files?

test/demo/contextListener.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    // Register a listener. We don't await as the promise won't resolve until
+    // the listener is triggered
+    const callbackSpy = jest.fn();
+    const callbackResponse =
+        fdc3Remote.addContextListener(testIdentity2, callbackSpy);
+    await new Promise(
+        res => setTimeout(
+            res,
+            1000));  // Slight delay to ensure callback is properly registered
```
pjbroadbent: Will we need these timeouts after every client call? Is there a way to build it into the framework?

test/demo/utils/ofPuppeteer.ts
```diff
@@ -0,0 +1,117 @@
...
+
+// This gets mounted by jest as part of our setup
+declare const global: NodeJS.Global&{__BROWSER__: puppeteer.Browser};
+
+export type TestWindowContext = Window&{
+  fin: Fin;
+  OpenfinFDC3: typeof import('../../../src/client/main')
+};
+
+export class OFPuppeteerBrowser {
```
pjbroadbent, line 15: Have only skimmed this so far, but if we're going to be wrapping Puppeteer in a util, can we make it more "OpenFin-y"?

As in, rather than using puppeteer's page and browser concepts, name things around "windows" and "apps"? Looking at the current util's, it's not clear what a page refers to exactly - a window?..

test/demo/testUtils.test.ts
tomrobertsOF: You'd think it would be easy to specify this in config, but apparently not. It was in every file because of an oversight when copy-pasting some boiler-plate.

I've changed it to only be in tests which we expect to take longer (which at the moment is just the context listener test since it creates apps).

test/demo/contextListener.test.ts
tomrobertsOF: Not every client call, just the addContextListener calls. Let's talk about this face-to-face later today.

test/demo/utils/ofPuppeteer.ts
tomrobertsOF: Part of the issue with that is that there isn't really a good correlation between the two that I've been able to figure out yet. We should discuss offline and maybe chat with someone from runtime about what the mappings are.

res/test/app-directory.json
```diff
@@ -0,0 +1,100 @@
+[{
```
bryangaleopenfin, line 1: Flagging that you or I will need to update this to fit with the changes in PR-17

res/test/test-app-main.json
```diff
@@ -0,0 +1,22 @@
...
+        "defaultWidth": 500
+    },
+    "services": [
+        {
+            "name": "fdc3",
+            "manifestUrl": "http://localhost:3923/test/test-provider.json"
+        }
+    ],
+    "runtime": {
+        "arguments": "--enable-aggressive-domstorage-flushing",
```
bryangaleopenfin: What does this do? Worth me adding to the Layouts test apps too? (Or to rephrase - where is it worth using this flag?)

src/provider/MetadataStore.ts
```diff
@@ -104,6 +104,9 @@ export class MetadataStore {
      * @param app An OpenFin application that has been created from 'appData'
      */
     public update(appData: IApplication, app: Application): void {
+        if (typeof appData.id !== 'number') {
```
bryangaleopenfin: Flagging that this will conflict now that  we use `appData.appId`, and is a `string`

src/demo/components/common/ContextMenu.ts
```diff
@@ -1,258 +1,245 @@
-import {WindowOption} from 'openfin/_v2/api/window/windowOption';
```
bryangaleopenfin, line 1: Safe to assume changes in this file are just linting?

test/runner.js
```diff
@@ -0,0 +1,77 @@
+const execa = require('execa');
```
bryangaleopenfin, line 1: Understand this not supporting the full set up command line options that runner.js in layouts does, but feel like `--file-name` support is needed

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -0,0 +1,59 @@
...
+export async function resolve(executionTarget: Identity, intent: IntentType, context?: Context): Promise<IApplication[]> {
+    return ofBrowser.executeOnWindow(executionTarget, async function(this: TestWindowContext, intent: IntentType, context?: Context): Promise<IApplication[]> {
+        return this.OpenfinFDC3.findIntent(intent, context).then(appIntent => appIntent.apps);
+    }, intent, context);
+}
+
+export async function broadcast(executionTarget: Identity, context: Context): Promise<void> {
+    return ofBrowser.executeOnWindow(executionTarget, async function(this: TestWindowContext, context: Context): Promise<void> {
+        return this.OpenfinFDC3.broadcast(context);
+    }, context).then(() => new Promise<void>(res => setTimeout(res, 100))); // Broadcast is fire-and-forget. Slight delay to allow for service to handle
```
bryangaleopenfin, line 36: If this is mirroring the client module, do we want to give a delay that broadcast doesn't?

test/demo/contextListener.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        // Check to see that the main window started correctly
+        expect(fin.Application.wrapSync(testManagerIdentity).isRunning()).resolves.toBe(true);
+
+        // Launch two test applications
+        await fdc3Remote.open(testManagerIdentity, testIdentity1.uuid);
+        await fdc3Remote.open(testManagerIdentity, testIdentity2.uuid);
+
+        // Check that the apps have started properly
+        expect(fin.Application.wrapSync(testIdentity1).isRunning()).resolves.toBe(true);
```
bryangaleopenfin: If we're using `expect(...).resolves`, should that be prefixed with await?

test/demo/contextListener.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        expect(fin.Application.wrapSync(testIdentity2).isRunning()).resolves.toBe(true);
+
+        // Register a listener. We don't await as the promise won't resolve until
+        // the listener is triggered
+        const callbackSpy = jest.fn();
+        const callbackResponse = fdc3Remote.addContextListener(testIdentity2, callbackSpy);
+        await new Promise(res => setTimeout(res,
```
bryangaleopenfin, line 32: Feels inconsistent with `broadcast` that this `timeout` isn't going inside `fdc3Remote.addContextListener`. Noting the comment, though is there no way to resolve this inconsistency?

test/provider/MetadataStore.unittest.ts
```diff
@@ -0,0 +1,79 @@
+import 'jest';
+
+import {Application} from 'openfin/_v2/main';
+
+import {IApplication} from '../../src/client/directory';
+import {AppMetadata, MetadataStore} from '../../src/provider/MetadataStore';
```
bryangaleopenfin, line 6: Flagging that this will conflict - name change of `IApplication`

bryangaleopenfin (approved):
Looks great! The only thing that jumps out as wrong-wrong (I think?) is the missing awaits when putting `expect`-ing on a how a `Promise` resolves. The rest is pretty subjective or stuff I want to better understand

test/demo/contextListener.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as fdc3Remote from './utils/fdc3RemoteExecution';
+import {getFinConnection} from './utils/fin';
+
+
+describe('advanced puppeteer functionality', () => {
+    const testManagerIdentity = {uuid: 'test-app', name: 'test-app'};
+    const testIdentity1 = {uuid: 'test-app-1', name: 'test-app-1'};
+    const testIdentity2 = {uuid: 'test-app-2', name: 'test-app-2'};
+
+    it('can can register and trigger a context listener', async () => {
```
bryangaleopenfin, line 11: Weird question, but what does `it` mean to you? Like, in your head are you saying "it", or "I.T". I don't get this Jest's choice of abbreviation 🤷‍♂️

(And on the other tests you're stating the `it` and `describe` messages with capital letters)

res/test/app-directory.json
tomrobertsOF: I guess whichever one is merged second gets the conflicts

res/test/test-app-main.json
tomrobertsOF: This is just a carry over from the other app manifests. I remember Phil looked into it and it's not really needed. I've removed it from everywhere it appears in this project.

src/demo/components/common/ContextMenu.ts
tomrobertsOF: Yea, didn't touch this file.

test/demo/utils/fdc3RemoteExecution.ts
tomrobertsOF: Difference here is after calling this util we would likely then check that the listener has been called. Alternative would be to have this delay pretty much everywhere we call this util in a test file. If you look at Phil's comments above he explicitly wanted not that.

test/demo/contextListener.test.ts
tomrobertsOF: Good catch. Will update

test/demo/contextListener.test.ts
tomrobertsOF: I discussed with Phil at length yesterday. This is the only way that we can see to do it for now due to the weird workaround. We need to get the value back from the listener, which without "exposeFunction" can only be done by returning a nested promise which resolves when the listener is triggered. Since await nested promises awaits the whole stack, we can't (at least as far as I can see) know when the remote code has actually finished running. As far as adding the delay to the util, we would then lose the ability to get the value, unless i've missed some trick we could do with promises

test/demo/contextListener.test.ts
tomrobertsOF: Defnitely some grammatical inconsistencies on my part. Lets discuss offline as a team and decide on some standard.

test/runner.js
rypete: Yes, the cli args would be nice to have

tomrobertsOF: Yea, i cut them out when getting it working and forgot to re-add. I'll copy most of it back in from layouts.

rypete (approved):

rypete:
Great job.  I would expect this to evolve a bit over time as it has before, but I think this is a great starting place to get things going.

test/demo/contextListener.test.ts
bryangaleopenfin: Yep. Another thing to take into account is how Jenkins interprets these strings in the results xml file - it treats '.' as a package separator, so that's why the hierarchy when looking at unit test results in Jenkins is a bit odd - there's the odd test that has a '.' in its name

tomrobertsOF: Ah, ok. I hadn't noticed that. Definitely needs fixing.

pjbroadbent (approved):

tomrobertsOF:
I have a couple more changes incoming to address PR comments, so please hold off on merging for a little bit.

res/test/test-app-main.json
pjbroadbent: This ensures any changes made to `localStorage` get written to disk immediately. Was used in layouts to make S&R testing easier. Not sure yet if first iteration of FDC3 will be using local/session storage for anything.

src/provider/MetadataStore.ts
tomrobertsOF: Resolved

test/runner.js
tomrobertsOF: After some discussions, we'll be adding these back in as part of the "Integrate service tooling into FDC3" story, so will leave as-is for now.

test/provider/MetadataStore.unittest.ts
tomrobertsOF: Resolved

tomrobertsOF:
> Great job. I would expect this to evolve a bit over time as it has before, but I think this is a great starting place to get things going.

Thanks buddy!

tomrobertsOF:
Ok, changes are in and everything seems to be passing now. Feel free to merge when happy.

PR #23, opened 2019-03-25
[SERVICE-395] Deploy updated FDC3 service to CDN

rypete:
** Passing CI **

The goods: 
* Swapped out GTS for ESLint
* Adds .d.ts generation
* Adds .vscode/settings.json to configure vsc eslint extension for the workspace

Bug: There is a [bug](https://github.com/eslint/typescript-eslint-parser/issues/531) with `@typescript-eslint/parser` which causes VSC to throw a small error window every time you use `const`.  To work around this for the time being, I've used `eslint.run: onSave`.  This is not an issue when using the CLI (read `npm run fix`) to lint.


bryangaleopenfin (approved):

src/demo/apps/DialerApp.tsx
```diff
@@ -79,9 +79,22 @@ export function DialerApp(props: AppProps): React.ReactElement {
...
-            {!inCall && <Dialer handleKeyPress={onDialerEntry} />}
+            {!inCall && <Dialer handleKeyPress={
+                (() => {
+                    if (Math.random() < 0.5) {
+                        return onDialerEntry;
+                    } else {
+                        return () => {};
+                    }
+                })()
+            } />}
```
pjbroadbent: Can we remove this? Looks like I was testing something, forgot to clean up again :-)

rypete: Fixed

pjbroadbent (approved):

PR #25, opened 2019-03-26
[SERVICE-392] Provider architechture

pjbroadbent:
Refactored the provider into a more modular architecture. Not all functionality of the current PoC has been ported over at this point - will need to make a call on whether it makes more sense to focus on preserving the current level of functionality, or getting the new structure in place ASAP.

Summary of refactoring changes:
- Split the "back end" parts of the provider into a 'model' and 'controller' components.
- Moved "front end" react code into 'view' sub-dir.
- Wiring between components managed via dependency injection.
- Extracted `fin` API usage into separate component for easier unit test setup & stubbing.
- Tweaks to API definitions in `APIHandler`, replaced double payload/return maps with a single map of tuples.
- Removed `MetadataStore`, the state stored in the model means these mappings will no longer be necessary.
- The `PreferencesStore` is currently unreferenced but not yet removed, need to decide if we'll keep this functionality.
- Moved much of the common/non-MVC files into `common` folder. The only files at the top-level `provider` directory are the `index.ts` "bootstrap", and the client API wiring.

Summary of other changes:
- All demo/provider windows now appear in the taskbar.
- Added service declaration to all demo apps.
- Removed unused app manifest (was a duplicate of the main demo app manifest).
- Fixed react bug that caused launcher to continuously fire `resolve` requests.
- Fixed missing 'selected' indicator on resolver buttons.
- Prevented blotter app from initially selecting the first row.
- Added `fdc3.` prefix to all FDC3-defined intent/context strings.
- Slight tweaks to intent/context definitions - adding general "intent" definition, removing 'Base' suffix from context type.

Some of the new components are still very "bare" - `AppDirectory.ts`, for example. The intention is that this PR will form an initial "skeleton" that will be built-out over several future stories.

There are some minor miscellaneous changes in this PR too. To see just the refactor-based changes, look at commit 448df47.


src/client/connection.ts
```diff
@@ -57,8 +57,11 @@ if (fin.Window.me.uuid !== SERVICE_IDENTITY.uuid || fin.Window.me.name !== SERVI
 
 /**
  * Wrapper around service.dispatch to help with type checking
+ *
+ * @param action String identifying the call being made to the provider
+ * @param payload Object containing additional arguments
  */
-export async function tryServiceDispatch<T extends APITopic>(action: T, payload: TopicPayloadMap[T]): Promise<TopicResponseMap[T]> {
+export async function tryServiceDispatch<T extends APITopic>(action: T, payload: API[T][0]): Promise<API[T][1]> {
```
bryangaleOF: Come back to

src/client/context.ts
```diff
@@ -53,10 +29,25 @@ export interface ContextBase {
...
-}
\ No newline at end of file
+    [key: string]: unknown;
+}
+
+// Built-in contexts
+export interface ContactContext extends Context {
+    type: 'fdc3.contact';
+    name: string;
+    id: {[key: string]: string}&{email?: string; twitter?: string; phone?: string};
```
bryangaleopenfin, line 47: \*eyeroll\* at Twitter being a first-order _thing_ in the spec

src/client/internal.ts
```diff
@@ -20,29 +24,17 @@ export const SERVICE_CHANNEL = 'of-fdc3-service-v1';
...
-    [APITopic.FIND_INTENT]: AppIntent;
-    [APITopic.FIND_INTENTS_BY_CONTEXT]: AppIntent[];
-    [APITopic.BROADCAST]: void;
-    // This is not strictly to spec as target should be able to return data.
-    // Needs a not-insignificant amount of provider changes to allow the two-way
-    // transmission of data before it can be put back to IntentResponse
-    // TODO: Revisit after SERVICE-392
-    [APITopic.RAISE_INTENT]: void;
+export type API = {
+    [APITopic.OPEN]: [OpenPayload, void];
```
bryangaleopenfin, line 46: No way for this to be {payload:OpenPayload, response:void}, so we're not relying on magic indexes, right?

src/demo/components/common/IntentButton.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
+import {IntentResolution} from '../../../client/main';
```
bryangaleopenfin, line 3: Do we want to be consistent about whether we're doing this or `import * as fdc3 from '../../../client/main'`?

src/demo/components/contacts/ContactsRow.tsx
```diff
@@ -40,7 +40,7 @@ export function ContactsRow(props: ContactRowProps): React.ReactElement {
 
     const getContext = (): ContactContext => {
         return {
-            type: "contact",
+            type: 'fdc3.contact',
```
bryangaleopenfin, line 44: Do these belong in an enum somewhere, similar to `Intents`?

src/client/internal.ts
bryangaleopenfin: And I guess we can't use `APISpecification` here?

src/provider/common/Injectables.ts
```diff
@@ -0,0 +1,28 @@
...
+    APP_DIRECTORY,
+    CONTEXT_HANDLER,
+    ENVIRONMENT,
+    INTENT_HANDLER,
+    MODEL,
+    SELECTOR
+}
+
+type InjectableMap = {
+    [K in keyof typeof Injectable]: K extends string ? K : never;
```
bryangaleopenfin, line 18: Feel like I'll need you to explain the purpose of the `never` here

src/provider/common/Injectables.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Declare all injectable types/values here.
+ * 
+ * This enum determines the keys that can be used within `@inject` tags. The mapping of these tags to default concrete types is defined in 
+ * `inversify.config.ts`. These mappings can be programmatically overridden by calling the methods of the {@link Injector} util.
```
bryangaleopenfin: Should `inversify.config.ts` be in this PR? (Or the comment changed to make it sound less like this file exists?)

src/provider/common/Injector.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    private static _container: Container = (() => {
+        const container = new Container();
+        const promises: Promise<unknown>[] = [];
+
+        Object.keys(Bindings).forEach(k => {
+            const key: Keys = k as any;
+
+            if (typeof Bindings[key] === 'function') {
+                container.bind(Inject[key]).to(Bindings[key] as any).inSingletonScope();
```
bryangaleopenfin, line 66: I feel confused how `Inject[key]` isn't just `key`

src/provider/common/Signal.ts
```diff
@@ -0,0 +1,262 @@
+// tslint:disable-next-line:no-any
```
tomrobertsOF, line 1: Should we just split this out into a common util since we're using it both here and in layouts?

src/provider/controller/IntentHandler.ts
```diff
@@ -0,0 +1,74 @@
...
+    @inject(Inject.SELECTOR)        private _selector!: SelectorHandler;
+    @inject(Inject.CONTEXT_HANDLER) private _contexts!: ContextHandler;
+
+    private _promise: Promise<IntentResolution>|null;
+
+    constructor() {
+        this._promise = null;
+    }
+
+    public async raise(intent: Intent): Promise<IntentResolution> {
```
tomrobertsOF, line 26: Maybe handled elsewhere and I just haven't seen it yet, but shouldn't this accept a target too?

src/provider/index.ts
```diff
@@ -1,132 +1,114 @@
...
-
-    /**
-     * The service should always use the current selection, whenever an intent of
-     * this type is fired.
-     */
-    ALWAYS_FOR_INTENT = 'ALWAYS_FOR_INTENT'
+import 'reflect-metadata';
+
+import {RaiseIntentPayload, APITopic, OpenPayload, FindIntentPayload, FindIntentsByContextPayload, BroadcastPayload} from '../client/internal';
+import {AppIntent, IntentResolution, Application, Intent} from '../client/main';
```
tomrobertsOF: Not sure if it's a good idea for provider/index to reference client/main. Could we split those shared types into a separate file or something? Could maybe have them in a hidden file and re-exported in main to keep the docs nice?

src/provider/index.ts
```diff
@@ -1,132 +1,114 @@
-import {AppId, Application, AppName} from '../client/directory';
-import {IntentType} from '../client/intents';
-import {RaiseIntentPayload} from '../client/internal';
-import {Context} from '../client/main';
-
-import {FDC3} from './FDC3';
-import {AppMetadata} from './MetadataStore';
-
-console.log('the provider has landed.');
```
tomrobertsOF, line 9: I'll be honest, this deletion makes me a bit sad :P

src/provider/index.ts
```diff
@@ -1,132 +1,114 @@
...
-     * Only specified when success is false.
-     */
-    reason?: string;
-
-    /**
-     * Determines the future behaviour of this intent
-     */
-    defaultAction: DefaultAction;
+@injectable()
+export class Main {
```
tomrobertsOF, line 19: Can we find a better name than "Main"? Maybe something like "FDC3Service", but open to any suggestions?

src/provider/controller/AsyncInit.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+@injectable()
+export abstract class AsyncInit {
+    private _initialized: Promise<this>;
+
+    constructor() {
+        this._initialized = this.init().then(() => this);
+    }
+
+    public get initialized(): Promise<this> {
```
bryangaleopenfin, line 19: We're doing `Promise<this>` rather than `Promise<void>` so we can do chaining?

src/provider/index.ts
```diff
@@ -1,132 +1,114 @@
...
+
+    private async findIntentsByContext (payload: FindIntentsByContextPayload): Promise<AppIntent[]> {
+        return [];
+    }
+
+    private async broadcast(payload: BroadcastPayload): Promise<void> {
+        await this._contexts.broadcast(payload.context);
+    }
+
+    private async raiseIntent(payload: RaiseIntentPayload): Promise<IntentResolution> {
```
tomrobertsOF, line 105: Shouldn't this also handle payload.target in some way?

src/provider/controller/IntentHandler.ts
```diff
@@ -0,0 +1,74 @@
...
+import {SelectorHandler, SelectorResult} from "./SelectorHandler";
+import {IntentResolution, Application } from "../../client/main";
+import {ContextHandler} from "./ContextHandler";
+import {FindFilter, Model} from "../model/Model";
+import {AppDirectory} from "../model/AppDirectory";
+import {AppWindow} from "../model/AppWindow";
+
+@injectable()
+export class IntentHandler {
+    @inject(Inject.APP_DIRECTORY)   private _directory!: AppDirectory;
```
bryangaleopenfin: At risk of bikeshedding, would prefer we standardized on putting @inject on its own line, but ok with being overruled

src/provider/model/AppDirectory.ts
```diff
@@ -0,0 +1,23 @@
+import {Application} from "../../client/directory";
+import {injectable} from 'inversify';
+
+@injectable()
+export class AppDirectory {
+    private _directory: Application[] = require('../../../res/provider/app-directory.json');
```
tomrobertsOF: I may be wrong, but would this cause the app directory JSON itself (not just the URL) to be hard-coded inside the main bundle once webpacked? If so, is that intended and/or likely to change in the future?

src/provider/model/AppDirectory.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    public async getAppByName(name: string): Promise<Application|null> {
+        return this._directory.find((app: Application) => {
+            return app.name === name;
+        }) || null;
+    }
+
+    public async getAppsByIntent(intentType: string): Promise<Application[]> {
```
tomrobertsOF, line 16: Is there any sense in also having a `getAppsByContext` for when we implement `findIntentsByContext` or is that likely to be handled differently?

src/provider/model/AppWindow.ts
```diff
@@ -0,0 +1,75 @@
...
+    public get contexts(): ReadonlyArray<ContextChannel> {
+        return this._contexts;
+    }
+
+    public focus(): Promise<void> {
+        return this._window.setAsForeground();
+    }
+}
+
+export interface Context {
```
tomrobertsOF: Are these two interfaces meant to be here? Seems a strange place.

src/provider/view/components/AppCard.css
```diff
@@ -1,19 +1,19 @@
...
-    border-color: limegreen;
-}
-
-.app-card img {
-    float: left;
-    padding: 3px;
-    margin: 0 12px;
-    width: 56px;
-    height: 56px;
+.w3-button.app-card {
```
tomrobertsOF, line 1: Long live w3-css!!

tomrobertsOF (commented):
Several comments and questions, but overall looks great so far.

src/provider/index.ts
```diff
@@ -1,132 +1,114 @@
...
+            contexts: this._contexts,
+            intents: this._intents
+        });
+
+        // Wait for creation of any injected components that require async initialization
+        await Injector.initialized;
+
+        // Current API
+        this.apiHandler.registerListeners<API>({
+            [APITopic.OPEN]: this.onOpen.bind(this),
```
bryangaleopenfin: Why `onOpen` rather than just `open` as with the other listeners?

src/provider/model/Environment.ts
```diff
@@ -0,0 +1,74 @@
...
+     * Arguments: (identity: Identity)
+     */
+    public readonly windowClosed: Signal1<Identity> = new Signal1();
+
+    public async createApplication(appInfo: Application): Promise<AppWindow> {
+        const app = await fin.Application.startFromManifest(appInfo.manifest);
+        return new AppWindow(app.identity, appInfo);
+    }
+
+    public wrapApplication(appInfo: Application, identity: Identity): AppWindow {
```
bryangaleopenfin, line 49: Something feels off with this, though I don't have a solution - since `Environment` isn't FDC3 specific in any way, it feels like this should be returning something like an `EnviromentWindow`, rather than `AppWindow` which is very FDC3 specific. And it feels odd we've got this abstraction layer that is still exposing "fin"-level things like `Application`

I'm not saying we should have `EnviromentWindow`, just something feels vaguely not quite right in how we're abstracting things. Probably an in person conversation

src/provider/model/Model.ts
```diff
@@ -0,0 +1,119 @@
...
+
+            if (preferredWindows.length > 0) {
+                return preferredWindows;
+            }
+        }
+
+        return windows;
+    }
+
+    public async findOrCreate(appInfo: Application, options?: FindOptions): Promise<AppWindow> {
```
bryangaleopenfin: This is an odd one, as we can potentially not find an application, depending on the options, but not be able to create it as it's already running. Right?

Should we only allow preferences rather than requirements here?

src/provider/model/Model.ts
```diff
@@ -0,0 +1,119 @@
...
+            await matchingWindow.focus();
+            return matchingWindow;
+        } else {
+            return this._environment.createApplication(appInfo);
+        }
+    }
+
+    private async onWindowCreated(identity: Identity, manifestUrl: string): Promise<void> {
+        const apps = await this._directory.getAllApps();
+        const appInfo = apps.find(app => app.manifest.startsWith(manifestUrl));
```
bryangaleopenfin, line 89: Definitely feels strange to be using `manifestUrl` when we have an `appId`. I get we can't use that directly here. Is this the problem you were talking about, where we need access to windows before they explicitly register themselves with the service?

src/provider/model/Model.ts
```diff
@@ -0,0 +1,119 @@
...
+    @inject(Inject.APP_DIRECTORY)
+    private _directory!: AppDirectory;
+
+    @inject(Inject.ENVIRONMENT)
+    private _environment!: Environment;
+
+    private _windows: AppWindow[];
+    private _channels: ContextChannel[];
+
+    constructor(@inject(Inject.ENVIRONMENT) environment: Environment) {
```
bryangaleopenfin, line 33: Any way to avoid needing to inject in both the constructor and the class declaration?

bryangaleopenfin (commented):
Looks good. Most of my comments are "I don't really understand this and need it explained to me"

src/client/intents.ts
```diff
@@ -1,3 +1,4 @@
+import { Context, ContactContext } from "./context";
```
MichaelMCoates: spacing in brackets

src/provider/common/Signal.ts
pjbroadbent: Yes, should definitely do that at some point - but not sure yet how best to do that.

A single-file `openfin-services-signal` module would be way too granular, but not sure right now what else it would make sense to group this with. Once the provider is built-out a bit more, we can make a call on how to modularise it.

src/provider/index.ts
pjbroadbent: What's the concern there? Those four types are all returnable from the API, and so shouldn't be in a hidden file.

No objections to re-arranging some of those types from `main.ts` to a more-specific file.

src/provider/index.ts
pjbroadbent: Yeah. This is still just a skeleton at this point - functionality is less than in current develop.

In the interests of unblocking things, will split this up into separate stories rather than trying to re-build all the current feature set into this structure within a single story. If there's no objections to that plan, of course - but makes sense to me.

src/provider/model/AppDirectory.ts
pjbroadbent: Yep, will be webpack-ed in to the provider.

Totally temporary, pretty much all of this file is. Also note how everything is async, even though the data is hard-coded - assuming "worst case" for now, and will decide how to handle the data fetching in a later story.

src/provider/model/Model.ts
MichaelMCoates: Do we think we could run into a situation in which a customer has multiple apps behind a `manifestUrl` that uses different query strings to launch different applications? `startsWith` might return the wrong app in that case?

src/provider/common/Injectables.ts
pjbroadbent: Oops. This is referring to `Injector.ts`, the file got renamed along the way when it became a util rather than using inversify directly. Cuts down on boilerplate a bit, since inversify can do a fair bit more than what we're using it for.

Will update the comment with the right filename.

src/provider/model/AppWindow.ts
pjbroadbent: Oops, yeah - these shouldn't be here.

src/provider/model/AppDirectory.ts
pjbroadbent: Yep - all of this will be built out further down the road. Still lots to do :-)

src/provider/controller/IntentHandler.ts
pjbroadbent: `target` would be a (optional) member of `Intent`. Thought I'd added that, but seems not.

Will add the field to `Intent`, but will save the actual implementation in the provider to a future story.

src/provider/index.ts
pjbroadbent: Sure, can discuss. Bugs me a bit that the entry point here is `index.ts`, but in the client it's `main.ts`. Should pick a consistent pattern.

src/client/internal.ts
pjbroadbent: Using `{payload, response}` would work too, I figured the tuple was a bit more concise - makes the actual API a bit easier to see in a 
`[FindIntentPayload, AppIntent]` => `(payload: FindIntentPayload) => AppIntent` kinda way.

Open to either - for reference, using the latter API definition would look like: 

```
interface API extends APISpecification<APITopic> {
    [APITopic.OPEN]: {payload: OpenPayload, response: void};
    [APITopic.FIND_INTENT]: {payload: FindIntentPayload, response: AppIntent};
    [APITopic.FIND_INTENTS_BY_CONTEXT]: {payload: FindIntentsByContextPayload, response: AppIntent[]};
    [APITopic.BROADCAST]: {payload: BroadcastPayload, response: void};
    [APITopic.RAISE_INTENT]: {payload: RaiseIntentPayload, response: IntentResolution};
}
```

As for `APISpecification`, seems we can add an "extends" (as above) without breaking any of the type checking. Seems like TypeScript doesn't really enforce that `API` has all the right keys, but the "integrity" of `API` still gets enforced when it's used (such as when calling `registerListeners`). Even if the type checking isn't perfect, having the "hint" there is definitely a bonus.

src/demo/components/contacts/ContactsRow.tsx
pjbroadbent: I was wondering if we should get rid of the `Intents` enum. The intents defined in the service are only the "built-in" intents, with the expectation that applications will add many more intent types. Since an enum can't be extended, it's actual usefulness to a user of the service may be limited.

But yeah, should definitely be consistent with intents - just a question of which way that should go, should discuss.

src/provider/common/Injectables.ts
pjbroadbent: TypeScript magic.

We want to be able to define the list of injectables as an enum, but we need the inversify identifiers to be strings. Once transpiled, an `enum` is an object that acts as a bi-directional mapping between string constants and number constants. This type will basically convert the enum from a "`string=>number | number=>string` mapping, to a "`string=>string`" mapping - mapping the `number` properties to `never` (effectively) removes them from the object.

Basically, in "JavaScript world" the `Injectable` enum looks like:
```
{
    APP_DIRECTORY: 0,
    CONTEXT_HANDLER: 1,
    ...
    0: 'APP_DIRECTORY',
    1: 'CONTEXT_HANDLER,
    ...
}
```
whereas `Inject` looks like:
```
{
    APP_DIRECTORY: 'APP_DIRECTORY',
    CONTEXT_HANDLER: 'CONTEXT_HANDLER',
    ...
}
```

src/demo/components/common/IntentButton.tsx
pjbroadbent: Agreed. There was only one place using `* as fdc3` (in `PreferencesStore.ts`), have switched that to importing just the properties that are used.

src/provider/index.ts
```diff
@@ -1,132 +1,114 @@
...
+            [APITopic.FIND_INTENT]: this.findIntent.bind(this),
+            [APITopic.FIND_INTENTS_BY_CONTEXT]: this.findIntentsByContext.bind(this),
+            [APITopic.BROADCAST]: this.broadcast.bind(this),
+            [APITopic.RAISE_INTENT]: this.raiseIntent.bind(this)
```
pjbroadbent: FYI, have also just bumped the TypeScript version to latest. Since these `bind()` calls all returned `any`, it basically made all of the type-safety within `APIHandler` completely pointless. :-)

src/provider/common/Injector.ts
pjbroadbent: True, since `InjectableMap` creates a maps a set of keys to themselves, there's no difference between the two.

`Inject[key]` _maaaaybe_ reflects the intent a little better - but yeah, no difference...

src/provider/controller/AsyncInit.ts
pjbroadbent: Yeah. Though in reality, not likely to ever be needed.

The `await` of these `AsyncInit` classes is now done within the `Injector` util (though this happens after all objects have been created, rather than pre-injection), so _I think_ we shouldn't need to do an `await foo.init()` anywhere in the "main" provider code.

src/provider/controller/IntentHandler.ts
pjbroadbent: Huh, TIL.. :-)

Yeah, I was in two minds about which style to use - but we should just pick one. Will align this with the other classes.

src/provider/model/Environment.ts
pjbroadbent: This is currently set up so that `FinEnvironment` is the only place where `AppWindow` objects are actually instantiated. The intention here is that in a test environment we'll wire up the injector to plug a `TestEnvironment` instance into each `Environment` location, and that `TestEnvironment` will instead create some kind of `TestWindow` class that has the same interface as `AppWindow`, but doesn't use the `fin` API in any way.

To do that cleanly should really mean extracting an interface from `AppWindow`, similar to `FinEnvironment` and `Environment`, but kinda deferring the decision around naming and feasibility of that approach until we start creating some "real" unit tests based on this structure.

To cut down on boilerplate, may have the "test" versions of the environment and window classes _extend_ the real types and selectively override the necessary bits, rather than needing to fully re-implement everything (for instance, `AppWindow` may well have a lot of non-`fin` stuff in it, that we don't want to have to duplicate.

Basically... TBD. Still needs some more thought/discussion.

src/provider/model/Model.ts
pjbroadbent: Hmmm, yeah - good point. Will change `options?: FindOptions` to `prefer?: FindFilter`.

src/provider/model/Model.ts
pjbroadbent: We need some way of mapping a `fin.Application` instance to it's app directory definition, though - and there's not much that the two have in common. If we're able to enforce that an FDC3 `appId` _always_ matches the `uuid` in the app's manifest - then, yes. That's potentially a very sensible thing to do, but need to check that we're able to make that assumption first. It's possibly worth noting that this is the only place we'd need this kind of check - once the window is registered in the model, we'll have a solid link between directory definition and UUID.

@MichaelMCoates Yeah, true - this was a quick first pass. If sticking with this approach, should really parse both URL's and make sure the key parts match - something along those lines.

src/provider/model/Model.ts
pjbroadbent: Could remove the `@inject` from the member and assign in constructor, or make this extend `AsyncInit` and add the listeners post-injection.

Any thoughts/preference?..

src/client/internal.ts
pjbroadbent: Heh, managed to confuse myself a bit here - `API` was actually being defined twice...

Once in `internal.ts`, where it's defined as a `type` and so can't use `APISpecification` (since `type` declarations have no kind of "extends" mechanism), and again in `index.ts` where it's defined as an `interface` (which means it can extend `APISpecification`).

Have removed the duplicate definitions in index.ts - but now, as you point out, can't really reference `APISpecification` here. It's possible if we move the declarations of all those `API*` types from the provider side to the client side, but would prefer to keep those out of the client if possible. Would also make breaking-out `APIHandler` into a separate module (as it's a good candidate for something that should be standardised across services) more complicated.

Will defer this one for now.

src/provider/common/Signal.ts
tomrobertsOF: Fair enough. Wasn't expecting it to be in this PR, but wanted to start the conversation.

tomrobertsOF: It's not even particularly services-specific (or even openfin-specific), so long-term could just split it out as a standalone 'simple-signals' module. Either way, discussion for later

src/provider/controller/IntentHandler.ts
tomrobertsOF: Sure. I'd guessed this was just an oversight, but will definitely need implementing asap as it's in the spec and used but the tests/demo apps

src/provider/index.ts
tomrobertsOF: The hidden/re-exporting suggestion was because it would allow them to appear in the top level docs but be imported into the provider without needing to import a client file with executing code, which is the crux of my issue with this pattern. 

If there is a different arrangement that would produce docs people are happy with while avoiding the need to import main from the provider I'm fine with that.

src/provider/index.ts
tomrobertsOF: Agreed

src/provider/index.ts
tomrobertsOF: I'd say there's no harm in propagating the optional param down to the point where it would be used, even if it just gets ignored with a comment. Targets, while optional, are a pretty significant part of the api spec and I think the skeleton should at the very least acknowledge their existence.

src/client/internal.ts
tomrobertsOF: In my view this comes down to a balance of two competing interests: magic indices make definitions more straightforward but referencing more confusing, whereas object-style makes definitions uglier while making referencing clearer and less error prone. 

I guess the correct solution would depend on whether we expect to do more defining or more referencing over the whole project.

src/demo/components/common/IntentButton.tsx
tomrobertsOF: I tend to like import {...} and direct references when getting only types, but prefer import * and dot-referencing when using functions/methods, but that may just be me rationalising my own poor choices.

src/provider/controller/IntentHandler.ts
tomrobertsOF: FWIW the Inversify docs all have it this way, but that doesn't mean we can't go our own way.

src/provider/index.ts
tomrobertsOF: Yay! Been wanting to do that for a while. Can you also do a quick check for any other places we're binding and make sure the types match up.

src/provider/common/Injectables.ts
bryangaleopenfin: Hmm... I want to run something by you when you're in. Wrapping `keyof typeof Injectable` in brackets seems to change the resulting type (at least as far as VSC is concerned)

src/provider/controller/IntentHandler.ts
bryangaleopenfin: They also recommend against property injection :p - https://github.com/inversify/InversifyJS/blob/master/wiki/property_injection.md

Suggest this is worth a 60 second (and no more!) huddle

src/provider/model/Model.ts
bryangaleopenfin: Let's decide in that 60 second huddle!

src/provider/model/Environment.ts
bryangaleopenfin: Ok, think I've figured out why this is giving me a bad feeling
- We're exposing two different levels of abstraction with this one class - the signals exposing things at the`Identity`/`uuid` level, and the methods exposing things at the `AppWindow` level
- There's nothing about the `wrapApplication` method that means it needs to be in the same class as the signals, they more "just happen" to be grouped together
- We get this odd ping-pong back and forth between this and Model when responding to a `windowCreated` signal

So I think this should be either:
- We change the signals to send `AppWindow`s
- We change `createApplication` to return `Promise<Identity>`, and pull out `wrapApplication` into a factory elsewhere

src/provider/controller/IntentHandler.ts
ottah: +1 for decorators/attributes on their own line.

src/client/internal.ts
pjbroadbent: True. I'd prefer cleaner definitions, particularly when we start making breaking changes that we have to maintain backward compatibility for. When we get to that point, we're going to have multiple API definitions (v1, v2, etc) for each service.

APIHandler is still pretty simple, some extra documentation in there that allows for more readable definitions within the "service" itself would be my preference.

src/provider/APIHandler.ts
```diff
@@ -4,15 +4,56 @@ import {Identity} from 'openfin/_v2/main';
...
+ * Defines the external API of the service. This is a mapping of method identifiers to `APIDefinition` tuples.
+ * 
+ * The keys of this mapping are the string 'topic' values that are used to communicate via the IAB Channel.
+ */
+export type APISpecification<T extends Enum> = {
+    [key in T]: APIDefinition;
 };
 
+/**
+ * Given an entry from an APISpecification, defines the function signature of a callback that is can be registered to
```
tim-dinsdale, line 39: typo: 'that is can be'

src/provider/Signal.ts
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * Signature for an aggregator function.
+ *
+ * On signals that allow callbacks to return a value (i.e. where `R != void`), an aggregator can be used to determine
+ * how the values returned by the signal callbacks are returned back to the code that emitted the signal.
+ *
+ * If no aggregator is provided, the `emit()` function on a signal has a `void` return type.
+ */
+export type Aggregator<R, R2> = (items: R[]) => R2;
+
```
tim-dinsdale: I get what this is doing, but the name is a little misleading, given none of these actually aggregate anything; i.e. there is nothing of the signature T[] => T. I think a better name is sthing like 'PostProcessor'. Thoughts?

pjbroadbent:
Marking as ready for review. Still some changes that need making, but will address in future stories. Aim here was to provide an initial basis for downstream stories, and there's enough currently in place and agreed upon to unblock that work.

MichaelMCoates (approved):

pjbroadbent:
Had some merge issues. PR re-opened as #59.

PR #27, opened 2019-04-01
Add more integration testing for FDC3 service

tomrobertsOF:
Reworks lots of the existing test utils, and should overall make things much easier to integration test. Also added tests for opening apps, resolving programmatically, and context broadcasts and listeners. There are some other areas that still need tests adding, but they will be done as a separate PR later down the line


test/runner.js
```diff
@@ -10,7 +10,7 @@ let port;
 
 // Should be cmd-line args at some point
 const debugMode = true;
-const skipBuild = false;
+const skipBuild = true;
```
tomrobertsOF: Changed this temporarily so I don't have to wait a minute before each test run. Will revert before making the full PR and will be changed to a CLI arg once the service-tooling stuff is in.

tomrobertsOF:
Moved this out of draft as it is blocking other stories and needs to be merged.

test/runner.js
tomrobertsOF: This is now changed back.

src/provider/FDC3.ts
```diff
@@ -295,8 +295,9 @@ export class FDC3 {
                         .then(() => {
                             clearTimeout(timeout);
                             if (context) {
-                                // Pass context to application before resolving
-                                fin.InterApplicationBus.publish('context', {context}).then(resolve).catch((reason: string) => reject(new Error(reason)));
+                                // Pass context to application before resolving (assume that the main window is the one with listeners registered)
+                                fin.InterApplicationBus.send({...app.identity, name: app.identity.uuid}, 'context', context)
```
pjbroadbent, line 299: Is the spread in the right place here? I assumed using the uuid as the name should be a fallback (in case `app.identity` doesn't include `name`), but here it's always overriding `app.identity.name`?

test/demo/open.test.ts
```diff
@@ -0,0 +1,171 @@
...
+            // From the launcher app, call fdc3.open with a valid name
+            await fdc3Remote.open(testManagerIdentity, 'test-app-1');
+
+            // Check that the app is now running
+            await expect(fin.Application.wrapSync({uuid: 'test-app-1'}).isRunning()).resolves.toBe(true);
+        }, 10000);
+
+        test('When passing an unknown app name the service returns an error', async () => {
+            // From the launcher app, call fdc3.open with an unregistered name
+            // TODO: fill in the error message once proider work is done and we know what it will be (SERVICE-392)
```
pjbroadbent: Typo: `proider`. Two more later in the file, too.

As an aside - whilst the error message will come later, it won't be part of SERVICE-392. Not yet written-up the story for `open`.

test/demo/open.test.ts
```diff
@@ -0,0 +1,171 @@
...
+                await expect(fin.Application.wrapSync({uuid: 'test-app-1'}).isRunning()).resolves.toBe(true);
+                await expect(fin.Application.wrapSync({uuid: 'test-app-2'}).isRunning()).resolves.toBe(true);
+            }, 10000);
+        });
+    });
+
+    describe('With context', () => {
+        const validContext: OrganizationContext = {type: 'organization', name: 'OpenFin', id: {default: 'openfin'}};
+        const testAppIdentity1: Identity = {uuid: 'test-app-preregistered-1', name: 'test-app-preregistered-1'};
+        const testAppIdentity2: Identity = {uuid: 'test-app-preregistered-2', name: 'test-app-preregistered-2'};
```
pjbroadbent, line 78: These could do with being top-level and used more widely. Still several instances below of duplicate identity objects.

test/demo/open.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        });
+
+        describe('With an app already running', () => {
+            beforeEach(async () => {
+                // Start by opening a test app
+                await fdc3Remote.open(testManagerIdentity, 'test-app-preregistered-1');
+            });
+
+            test(
+                'When opening the running app it is focused, its context listener is triggered with the correct data, and the promise resolves',
```
pjbroadbent, line 117: Can omit information from the test name that is already covered in the surrounding `describe`s.

As a rule of thumb, the `describe`s should explain the context (given/when), and the `it`s describe the assertion being made (then).

test/demo/open.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+            test('When opening the already running app the app is focused and the promise resolves', async () => {
+                // Focus another window so we do not get false positives from the first open focusing
+                await fin.Window.wrapSync(testManagerIdentity).focus();
+
+                // From the launcher app, call fdc3.open with a valid name
+                await fdc3Remote.open(testManagerIdentity, 'test-app-1');
+
+                // Check that the app is still running
+                await expect(fin.Application.wrapSync({uuid: 'test-app-1'}).isRunning()).resolves.toBe(true);
```
bryangaleopenfin: Why would we be worried about the app not still being running?

test/demo/contextListener.test.ts
```diff
@@ -1,48 +1,135 @@
...
 
-        const fin = await getFinConnection();
+// These tests all use the default channel for context broadcasts and listeners
+// Context-passing in a channelled environment is tested in a seperate file
+describe('Context listeners and broadcasting', () => {
+    // Adapter connection is established in setup. Reassign here for convenience
+    const fin: Fin = global.__FIN__;
 
-        // Check to see that the main window started correctly
+    beforeEach(async () => {
```
bryangaleopenfin, line 17: Why do we want to check for this in some of our tests but not others? And to match the comment, should this check be in an `afterEach` instead/as well? If this is something we want for all tests, could this be automatically set up to run for all tests using `setupFilesAfterEnv`?

test/demo/contextListener.test.ts
```diff
@@ -1,48 +1,135 @@
...
+
+import {OrganizationContext} from '../../src/client/main';
 
+import * as fdc3Remote from './utils/fdc3RemoteExecution';
 
-describe('advanced puppeteer functionality', () => {
-    const testManagerIdentity = {uuid: 'test-app', name: 'test-app'};
-    const testIdentity1 = {uuid: 'test-app-1', name: 'test-app-1'};
-    const testIdentity2 = {uuid: 'test-app-2', name: 'test-app-2'};
+const testManagerIdentity = {
```
bryangaleopenfin, line 7: Worth pulling this out somewhere to save duplication?

test/demo/contextListener.test.ts
```diff
@@ -1,48 +1,135 @@
...
+    describe('Registering and unsubscribing context listeners', () => {
+        const testAppIdentity = {uuid: 'test-app-1', name: 'test-app-1'};
+        beforeEach(async () => {
+            // Open the app to be used in each test
+            await fdc3Remote.open(testManagerIdentity, testAppIdentity.uuid);
+        });
+
+        afterEach(async () => {
+            // Close down the app once done
+            await fin.Application.wrapSync(testAppIdentity).quit(true);
```
bryangaleopenfin, line 35: In other places, we're not passing in `true` to force quit, and catching errors. Why the difference here? And worth pulling out to save duplication?

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -43,25 +43,87 @@ export async function broadcast(executionTarget: Identity, context: Context): Pr
...
+    const id = await ofBrowser.executeOnWindow(executionTarget, function(this:TestWindowContext): number {
+        const listenerID = this.contextListeners.length;
+        this.contextListeners[listenerID] = this.OpenfinFDC3.addContextListener((context) => {
+            this.receivedContexts.push({listenerID, context});
         });
+        return listenerID;
+    });
+
+    return {
+        type: 'context',
```
bryangaleopenfin, line 88: What's `type` here for? Assume there's some plan in mind?

test/demo/utils/jestGlobals.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
bryangaleopenfin: For my understanding, how come this doesn't need to be explicitly imported? Assume there's some magic happening somewhere?

test/demo/contextListener.test.ts
```diff
@@ -1,48 +1,135 @@
...
-        // Creating the apps can take a while, so we need a longer timeout.
-        jest.setTimeout(20000);
+const validContext: OrganizationContext = {type: 'organization', name: 'OpenFin', id: {default: 'openfin'}};
 
-        const fin = await getFinConnection();
+// These tests all use the default channel for context broadcasts and listeners
+// Context-passing in a channelled environment is tested in a seperate file
+describe('Context listeners and broadcasting', () => {
+    // Adapter connection is established in setup. Reassign here for convenience
+    const fin: Fin = global.__FIN__;
```
pjbroadbent: Is it possible to make this a util/import, same as other dependencies? Seems the env script runs early enough for that to happen?..

pjbroadbent (approved):
Some minor points, can be addressed in future PR if necessary.

test/demo/contextListener.test.ts
tomrobertsOF: Done

src/provider/FDC3.ts
tomrobertsOF: app.identity will never have a name

This was a quick fix to make it send to the main window of an app since before it was just broadcasting to everyone (which according to the spec it shouldn't). This can (and should) be revisited as part of any provider changes we do.

test/demo/open.test.ts
tomrobertsOF: At the time I wrote the comment 392 was the catch-all provider story. I'll change it to SERVICE-TBD?

test/demo/contextListener.test.ts
```diff
@@ -1,48 +1,135 @@
...
+            // Open the app to be used in each test
+            await fdc3Remote.open(testManagerIdentity, testAppIdentity.uuid);
+        });
+
+        afterEach(async () => {
+            // Close down the app once done
+            await fin.Application.wrapSync(testAppIdentity).quit(true);
+        });
+
+
```
bryangaleopenfin: Double line break

test/demo/open.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+        describe('With an app already running', () => {
+            beforeEach(async () => {
+                // Start by opening a test app
+                await fdc3Remote.open(testManagerIdentity, 'test-app-1');
+            });
+
+            test('When opening the already running app the app is focused and the promise resolves', async () => {
+                // Focus another window so we do not get false positives from the first open focusing
+                await fin.Window.wrapSync(testManagerIdentity).focus();
```
bryangaleopenfin, line 53: Worth doing - `await expect(fin.System.getFocusedWindow().then(w => w.uuid)).resolves.not.toBe('test-app-1');`, so the test directly observes focus going from one window to another?

test/demo/open.test.ts
```diff
@@ -0,0 +1,171 @@
...
+                    // And that the app is focused
+                    await expect(fin.System.getFocusedWindow().then(w => w.uuid)).resolves.toBe('test-app-preregistered-1');
+
+                    // Retrieve the list of contexts the app received
+                    const preregisteredListener = await fdc3Remote.getRemoteContextListener(testAppIdentity1);
+                    const receivedContexts: Context[] = await preregisteredListener.getReceivedContexts();
+
+                    // Check that the app received the context passed in open and nothing else
+                    expect(receivedContexts.length).toBe(1);
+                    expect(receivedContexts.slice(-1)[0]).toEqual(validContext);
```
bryangaleopenfin: Admittedly noodling here - what's the value of the `slice(-1)`? We've already checked the length is 0, and if that fails, the test won't go any further

bryangaleopenfin (commented):
Looks really good. Test naming is really clear and (as someone who is generally pretty negativeon comments) the comments are really helpful. Most got some questions rather than 'this must change'.

My biggest question is what's the plan for unifying style standards with layouts PR-357? Mainly using `it()` rather than `test()`, `*.inttest.ts` rather than `*.test`, (or changing what we do in layouts) and using the same formatting for junit output.

test/demo/contextListener.test.ts
tomrobertsOF: edit: commented on wrong comment

Both likely artefacts from where I chose to copy/paste each line from. This still needs a cleanup/consistency pass to deal with all of that type of stuff. I'm sure if you look hard enough there's quite a few similar inconsistencies.

test/demo/utils/jestGlobals.ts
tomrobertsOF: [Ambient declaration](https://basarat.gitbooks.io/typescript/docs/types/ambient/intro.html)

tomrobertsOF: Also, I'm changing this per Phil's comment above

test/demo/open.test.ts
tomrobertsOF: Not sure what the question is here?? We do that check once the open call has happened.

test/demo/open.test.ts
tomrobertsOF: Leftover from before I added the length check. Forgot to change it here.

test/demo/contextListener.test.ts
tomrobertsOF: fixed

test/demo/utils/fdc3RemoteExecution.ts
tomrobertsOF: Yea, the utils as they currently stand don't include any stuff for intent listeners. This was in place to allow a more unified "mountEventListener" util which is used by addContextListener and addIntentListener once I got the intent listener stuff in. Since I'm no longer including that in this story I could probably remove it.

test/demo/contextListener.test.ts
tomrobertsOF: As with other comments, this was going to be part of my final cleaning pass before PRing, but  haven't done that yet. Have a couple ideas of ways to do it which avoid hard-coding, but likely won't make it in in time for this PR.

test/demo/contextListener.test.ts
tomrobertsOF: It's there mostly an early catch incase some test unknowingly/accidentally closes the manager window. If we didn't have it and the manager (and so also the service) was closed, we would get lots of failures for a variety of different reasons. With this check we would see clearly exactly when it got closed.

Also, afterEach would be covered by the next test's beforeEach anyway, so don't see the point in doing it twice.

test/demo/open.test.ts
tomrobertsOF: No idea. But worth testing.

test/demo/open.test.ts
tomrobertsOF: done

test/demo/open.test.ts
tomrobertsOF: So for the highlighted case it would be:
```javascript
describe('Opening apps....', () => {
  describe('With context', () => {
    describe('App already running', () => {
      beforeEach() // Open app1
      describe('Opening the already running app', () => {
        beforeEach() // Open app1
          test('The app is focused and its listener triggered')
        })
        describe('Opening a second app', () => {
          beforeEach() // Open app2
          test('Second app opens and listener is triggered')
        })
     })
  })
})
```

We end up with very nested describe blocks with a single test in them. Is that really a good idea?

test/demo/open.test.ts
bryangaleopenfin: I mean a check to confirm that the app *isn't* focused before we open it - `await expect(fin.System.getFocusedWindow().then(w => w.uuid)).resolves.*not*.toBe('test-app-1');`. So we confirm we have the expected state before, do the thing we're testing, then confirm we have the expected state after

tomrobertsOF: Ah, ok. I mean, that would only ever be the case if 
```
await fin.Window.wrapSync(testManagerIdentity).focus();
```
for some reason didn't actually focus the other window, but at that point we're just throwing in extra async calls to the core for the sake of testing runtime code

tomrobertsOF: I think its safe to take as a given that calling focus on a window will focus that window

tomrobertsOF:
@bryangaleOF, re: unifying style standards: I was planning on doing that in one big go once I had everything working, but I guess now it will be bumped into the next PR.

bryangaleopenfin (approved):

PR #28, opened 2019-04-01
[SERVICE-417] Replace IAB usage with Channels API

ottah:
Client & Provider IAB calls changed to use channels instead.


src/provider/FDC3.ts
```diff
@@ -360,27 +360,27 @@ export class FDC3 {
...
     }
 
-    private async sendContext(context: ContextBase): Promise<void> {
-        return fin.InterApplicationBus.publish('context', context);
+    private async sendContext(context: ContextBase): Promise<any> {
+        return this.apiHandler.channel.publish('context', context);
     }
 
-    private async sendIntent(intent: RaiseIntentPayload, targetApp: AppMetadata): Promise<void> {
+    private async sendIntent(intent: RaiseIntentPayload, targetApp: AppMetadata): Promise<any> {
```
bryangaleopenfin: Do we want to change the signature? I feel like a change to our internal messaging system shouldn't change `void` to `any`. Could do this.apiHandler.channel.publish('intent', intent).then(() => {});

bryangaleopenfin: Ditto for `sendContext`

src/provider/FDC3.ts
```diff
@@ -254,7 +254,7 @@ export class FDC3 {
                             clearTimeout(timeout);
                             if (context) {
                                 // Pass context to application before resolving
-                                fin.InterApplicationBus.publish('context', {context}).then(resolve).catch((reason: string) => reject(new Error(reason)));
+                                this.apiHandler.channel.dispatch(app.identity, 'context', {context});
```
bryangaleopenfin: Just want to check we're happy to lose the error handling here

bryangaleopenfin (commented):
Looks good, just a couple of things I wanted to check

src/provider/FDC3.ts
tomrobertsOF: Not just the error handling, but the `.then(resolve)` was also removed. This is what makes sure we have sent the context before resolving the `open` call, and should definitely be there.

src/provider/FDC3.ts
pjbroadbent: +1 - if the return value isn't used, should keep the return type as `void`.

Methods are `async` already, replace the `return`s with `await`s?..

pjbroadbent (approved):

bryangaleopenfin (approved):

PR #29, opened 2019-04-01
[SERVICE-419] Prototype 'Context Channels' mechanism

bryangaleopenfin:
First swing at context channels implementation

Flagging that there'll be some conflicts with @TheoOttahOF's IAB channels work


src/client/contextChannels.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Event that is fired whenever a window changes from one channel to another.
+ *
+ * This includes switching to/from the global channel. The `channel` and
+ * `previousChannel` fields use the same conventions for denoting the global channel as `getChannel`.
+ */
+export function addEventListener(
+    event: 'channel-changed',
+    listener: (event: {identity: Identity; channel: Channel; previousChannel: Channel}) => void, identity?: Identity
```
pjbroadbent: Use `ChannelChangedListener` here?

src/client/contextChannels.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export interface ChannelChangedPayload {
+    identity: Identity,
+    channel: Channel,
+    previousChannel: Channel
+}
+
+
+type ChannelChangedListener = (event: {identity: Identity; channel: Channel; previousChannel: Channel}) => void;
```
pjbroadbent: Use `ChannelChangedPayload` here?

tomrobertsOF (approved):

src/provider/FDC3.ts
```diff
@@ -360,8 +402,8 @@ export class FDC3 {
         });
     }
 
-    private async sendContext(context: ContextBase): Promise<void> {
-        return fin.InterApplicationBus.publish('context', context);
+    private async sendContext(identity: Identity, context: ContextBase): Promise<void> {
+        return fin.InterApplicationBus.send(identity, 'context', context);
```
pjbroadbent, line 406: Are we going to switch this to use an IAB channel? To avoid adding extra work to SERVICE-417?

bryangaleopenfin: Figured leaving this as is for now was going to conflict less with Theo's work than changing to use channels. It's not a new use of IAB, can changing it here would necessitate changing it at the receiver end (right?), which would also conflict

pjbroadbent (approved):

PR #31, opened 2019-04-01
[SERVICE-418] Have service deploy with empty app directory

MichaelMCoates:
Checks our build environment and if the default URL has changed. If we're in production and that URL hasn't changed from the demo, return an empty array. Is this sufficient? Is there something I'm missing here? The provider works fine with this change.


bryangaleopenfin:
Hmm, I read this story as meaning we wanted to have an empty directory JSON we could deploy, though I could be wrong.

In any case, it might be worth changing the name of the current directory JSON to indicate it's the development sample data JSON

src/provider/AppDirectory.ts
```diff
@@ -42,6 +42,10 @@ export class AppDirectory {
     }
 
     private async request(): Promise<Application[]> {
+        // If using the demo app directory in production, return an empty array.
+        if (process.env.NODE_ENV === 'production' && AppDirectory.URL === 'http://localhost:3923/provider/app-directory.json') {
```
tomrobertsOF: Does this work? This file runs in openfin, so we won't have `process.env` right? Unless web pack does something clever with it that I'm not aware of...

tomrobertsOF:
Additional process note: it doesn't seem worth opening draft PRs when the intention is to actually merge the changes. I feel like they should be for more in-progress/half-finished type stuff (the kind of ones where in the past we would have put "DO NOT MERGE" at the top), but open for a broader discussion if people disagree.

src/provider/AppDirectory.ts
MichaelMCoates: It works on my machine? When you specify a `mode` for webpack (`production` or `development`), it drops that into `process.env.NODE_ENV`, so you can check against it. I think it also will pull those lines out of your built code if you use those variables in if blocks like this, but I'd need to get more into the specifics on that. Can someone test on their machine if they have access to `process.env.NODE_ENV`?

MichaelMCoates:
> Additional process note: it doesn't seem worth opening draft PRs when the intention is to actually merge the changes. I feel like they should be for more in-progress/half-finished type stuff (the kind of ones where in the past we would have put "DO NOT MERGE" at the top), but open for a broader discussion if people disagree.

Sounds good! Will do.

src/provider/AppDirectory.ts
pjbroadbent: heh, yeah - seems webpack does some string replacement here.

This is going to be a temporary mechanism, so going to file this one under "it's weird, but it works".

MichaelMCoates:
> Hmm, I read this story as meaning we wanted to have an empty directory JSON we could deploy, though I could be wrong.
> 
> In any case, it might be worth changing the name of the current directory JSON to indicate it's the development sample data JSON

True, I was confused as well. After talking to Phil, I think that this first pass will suffice until we establish true App Directory configuration.

Good point! Changed to `sample-app-directory.json`. Would `dev-app-directory` be better?

pjbroadbent (approved):

src/provider/AppDirectory.ts
tomrobertsOF: Fair enough. If it works then fine by me. This seems like it could be useful in other places, but a discussion for another time. Webpack never ceases to amaze me with it's magic!

tomrobertsOF:
I think I'd prefer `dev-app-directory`, but doesn't really matter since this won't be around too long either way.

tomrobertsOF (approved):

PR #33, opened 2019-04-02
[N/A] Bugfix case where we are moving a window back to the global channel

bryangaleopenfin:


pjbroadbent (approved):

pjbroadbent:
Going to merge with one approval so that we can hand this over to CS

PR #34, opened 2019-04-02
[N/A] Channel model bugfixing

bryangaleopenfin:
Fixes more bugs in Channel Model revealed by integration tests

- A window could be listed as being part of two channels
- When querying windows in channels, we would return whatever had been passed to us previously rather than a sanitized Identity object

### DIFF
--- a/src/provider/ChannelModel.ts
+++ b/src/provider/ChannelModel.ts
@@ -108,7 +108,7 @@ export class ChannelModel {
                     this._channelIdToIdentitiesMap.set(channelId, identities);
                 }
 
-                identities.push(identity);
+                identities.push({name: identity.name, uuid: identity.uuid});
             }
 
             const channel = this._channelIdToChannelMap.get(channelId)!;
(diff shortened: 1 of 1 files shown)


src/provider/ChannelModel.ts
```diff
@@ -90,14 +90,17 @@ export class ChannelModel {
         const previousChannelId = this._identityHashToChannelIdMap.get(identityHash) || GLOBAL_CHANNEL_ID;
```
MichaelMCoates, line 90: If we're concerned about a bad identity passed here, should we have an `identity.name !== undefined` check here?

bryangaleopenfin: For the purposes of the prototype, I'm not concerned about bad identities, and that's probably a thing to check outside of this anyway

bryangaleopenfin: The sanitizing is more just so that the data we send out is 'clean', and doesn't mean anyone's given us a bad identity. It was primarily motivated to make test-writing easier, but it feels like the the 'right' thing to do anyway

tomrobertsOF (approved):

pjbroadbent (approved):

src/provider/ChannelModel.ts
pjbroadbent: Yeah, we'll eventually do the sanitising at the client/APIHandler level, rather than having to put that kind of thing all over the provider.

Doing it here is good for the time being, though.

PR #37, opened 2019-04-04
[N/A] JSdoc linting & .gitignore update

ottah:
- `valid-jsdoc` is deprecated so I've replaced it with `eslint-plugin-jsdoc`
- Warnings for JSdoc only happens when there is documentation that isn't meeting the requirements, but will not warn for missing documentation e.g. React components.
- Added .vscode to the .gitignore as it is a local settings folder and shouldn't be tracked

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,6 +11,9 @@
         "plugin:import/errors",
         "plugin:react/recommended"
     ],
+    "plugins": [
+        "jsdoc"
+    ],
     "globals": {
         "fin": "readonly"
     },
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 .vs
+.vscode
```
pjbroadbent, line 2: I think we decided to version control this, since there's configuration in there to setup the linter within the IDE?

Could be thinking of layouts, but we'd want to do that on both eventually. @ryanpetersonOF?

Might be that we want to ignore the folder, but still include certain files?..

.eslintrc.json
```diff
@@ -60,16 +63,28 @@
...
+        "jsdoc/require-hyphen-before-param-description": 0,
+        "jsdoc/require-param": 1,
+        "jsdoc/require-param-description": 1,
+        "jsdoc/require-param-name": 0,
+        "jsdoc/require-param-type": 0,
+        "jsdoc/require-returns": 0,
+        "jsdoc/require-returns-check": 0,
+        "jsdoc/require-returns-description": 1,
+        "jsdoc/require-returns-type": 0,
+        "jsdoc/valid-types": 1
```
pjbroadbent, line 88: Can you use `error`/`warn`/`off` as above, rather than integers?

.gitignore
rypete: Yeah we had a 'in the real world' discussion about this.  It should be removed from the .gitignore in favo(u)r of eslint configs

tomrobertsOF:
Going to be done with svc-tooling instead

PR #39, opened 2019-04-04
[N/A] API doc updates

pjbroadbent:
Added theming to API docs, renamed modules to produce a more readable sidebar, hidden the 'internal' file from the API docs.


src/client/context.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @module Context
```
bryangaleopenfin: Wondering if this should be plural to match Intents and ContextChannels

bryangaleopenfin (approved):

MichaelMCoates (approved):
There's a couple typos in the docs, but that's for a separate PR. Looks great.

tomrobertsOF (approved):

PR #46, opened 2019-04-05
[N/A] Integration tests around Context Channels

bryangaleopenfin:
3 main parts to this
- Revisions to `addEventListener`, to be closer to what we eventually want it to be (i.e., supporting multiple event types, not just hard-coded to support 'channel-changed'
- Additions to fdc3RemoteExecution to support new context channel API functions
- The tests themselves

Some nastiness to flag:
- Lots of duplication in fdc3RemoteExecution, between code for handle events and code for handling contexts, though I think a lot of that is changing anyway? (@tomrobertsOF)
- `addEventListener` now returns an `EventListener` object, rather than null, which was needed to get `addEventListener` in fdc3RemoteExecution working


res/test/test-app.html
```diff
@@ -38,6 +44,13 @@
                     receivedIntents.push({id: nextIntentListenerID, intent, context});
                 });
             }
+
+            for (const event of events) {
+                const nextEventListenerID = window.eventListeners.length;
+                window.eventListeners[nextEventListenerID] = OpenfinFDC3.addEventListener(event, (payload) => {
```
pjbroadbent: `OpenfinFDC3` has now been renamed to `fdc3`

src/client/main.ts
```diff
@@ -66,6 +69,14 @@ export interface IntentResolution {
     version: string;
 }
 
+export type EventPayload = ChannelChangedPayload;
+export type Event = 'channel-changed';
```
pjbroadbent: Can we rename `Event`, there's already a built-in `Event` type. Makes the `addEventListener` signature below look like it takes an event object.

src/client/main.ts
```diff
@@ -266,3 +287,29 @@ export function addContextListener(handler: (context: Context) => void): Context
...
+        if (index >= 0) {
+            eventListeners.splice(index, 1);
+        }
+
+        eventEmitter.removeListener(event, handler);
+        return index >= 0;
+    }};
+
+    eventListeners.push(listener);
+    eventEmitter.addListener(event, handler);
```
pjbroadbent: How come we have two lists of event listeners? Can we just use `eventEmitter`, and drop `eventListeners`?

Assuming the latter is to help with testing? Not too familiar with that yet - lets follow-up.

test/demo/contextChannels.test.ts
```diff
@@ -0,0 +1,270 @@
+import 'jest';
+import {connect, Fin, Identity, Application} from 'hadouken-js-adapter';
+
+import * as fdc3Remote from './utils/fdc3RemoteExecution';
+
+const testManagerIdentity = {uuid: 'test-app', name: 'test-app'};
+
+const testContext = {type: 'test-context', name: 'contextName1', id: {name: 'contextID1'}};
+
+const startedApps:Application [] = [];
```
pjbroadbent: Is this linted? Whitespace looks a bit off here...

test/demo/contextChannels.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    const app3 = {uuid: 'test-app-3', name: 'test-app-3'};
+    const app4 = {uuid: 'test-app-4', name: 'test-app-4'};
+
+    const appIdentities = [app1, app2, app3, app4];
+
+    // Creating apps takes time, so increase timeout
+    jest.setTimeout(channels.length * 5000);
+
+    const result: Identity[] = await Promise.all(channels.map(async (channel, index) => {
+        const identity = appIdentities[index];
```
pjbroadbent, line 42: Is there always at-most 4 channels? Rather than pre-defining the list of identities, can this be: ?
```
const identity = {uuid: `test-app-${index}`, name: `test-app-${index}`};
```

test/demo/contextChannels.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+        const channelChangingListener = await fdc3Remote.addContextListener(channelChangingWindow);
+
+        // Broadcast our context on the global channel
+        await fdc3Remote.broadcast(testManagerIdentity, testContext);
+
+        // Check our window received our test context
+        const receivedContexts = await channelChangingListener.getReceivedContexts();
+        expect(receivedContexts.length).toBe(1);
+        expect(receivedContexts[0]).toEqual(testContext);
```
pjbroadbent: Would `expect(receivedContexts).toEqual([testContext]);` be equivalent to both of these assertions?

test/demo/contextChannels.test.ts
```diff
@@ -0,0 +1,270 @@
...
+        const redReceivedContexts = await redListener.getReceivedContexts();
+        expect(redReceivedContexts.length).toBe(1);
+        expect(redReceivedContexts[0]).toEqual(testContext);
+
+        const globalReceivedContexts = await globalListener.getReceivedContexts();
+        expect(globalReceivedContexts.length).toBe(0);
+
+        // Check our blue window received no context
+        const blueReceivedContexts = await blueListener.getReceivedContexts();
+        expect(blueReceivedContexts.length).toBe(0);
```
pjbroadbent: `expect(blueReceivedContexts).toHaveLength(0);`?

Check is equivalent, but should give more useful error messages if it fails(?).

test/demo/contextChannels.test.ts
bryangaleopenfin: Yep, linted. Linter allows zero or one spaces here. But removed space anyway

src/client/main.ts
bryangaleopenfin: Changed to EventType

test/demo/contextChannels.test.ts
bryangaleopenfin: Maybe we'll have tests in here in future that use more than 4 windows, but we can set those test apps up when we need to (or we'll have some kind of dynamic directory/manifest thing by then).

I always feel iffy about treating IDs as anything other than totally opaque, and in practice the difference is going to be whether we get an error at `const identity = appIdentities[index]` or whether we get it at `await fdc3Remote.open(testManagerIdentity, identity.uuid)`, which is kinda neither here nor there.

src/client/main.ts
bryangaleopenfin: "Assuming the latter is to help with testing?" - yep, but I think it's changing anyway. Like, there's two competing standards to follow - stay in line with how addContextListener/addIntentListener work, or stay in line with how addEventListener works in layouts, and there's no nice way to satisfy both

bryangaleopenfin:
@pjbroadbent Addressed or at least replied to everything. Can discuss anything outstanding whenever

tomrobertsOF (approved):

tomrobertsOF:
All looks fine to me. Gonna be some conflicts with my testing improvements branch, but I'll deal with them when the time comes.

bryangaleopenfin:
Done with changes for now, ready for re-review (@pjbroadbent)

pjbroadbent (approved):

PR #54, opened 2019-04-11
[SERVICE-414] Part 2: Add integration tests for intentListeners

tomrobertsOF:
Note: There are still quite a few TODOs and TBDs in the tests. Most of these are due to either undefined behaviour or unimplemented features. Despite that I think there is value in getting this into develop so we can test those features as they become better defined.

- Added utils for intentListeners, matching the style of the previous contextListeners. It may make sense to combine the utils for these since there is a fair bit of not-quite-duplicate code, but that will be covered in a future story.
- Added tests for existing behaviour where possible
- Added placeholder tests for potential future features or cases where expected behaviour is not yet defined


res/test/test-app.html
```diff
@@ -33,9 +33,12 @@
             });
     
             for (const intent of intents) {
-                const nextIntentListenerID = window.intentListeners.length;
-                window.intentListeners[nextIntentListenerID] = fdc3.addIntentListener(intent, (context) => {
-                    receivedIntents.push({id: nextIntentListenerID, intent, context});
+                if (window.intentListeners[intent] === undefined) {
```
bryangaleopenfin: I'm feeling confused now why we need to check this. We've just set `intentListeners` to [], I'm not seeing where anything else could have jumped in an modified it. And how would `nextIntentListenerID` be anything other than 0?

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -127,3 +168,31 @@ export async function getRemoteContextListener(executionTarget: Identity, listen
         }
     };
 }
+
+export async function getRemoteIntentListner(executionTarget: Identity, intent: IntentType, listenerID: number = 0): Promise<RemoteIntentListener> {
```
bryangaleopenfin: Spelling of Listener

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -67,14 +68,27 @@ export async function getChannelMembers(executionTarget: Identity, channelId: Ch
...
     unsubscribe: () => Promise<void>;
     getReceivedContexts: () => Promise<Context[]>;
 }
 
+export interface RemoteIntentListener {
+    remoteIdentity: Identity;
+    id: number;
+    intent: IntentType;
+    unsubscribe: () => Promise<void>;
+    getReceivedIntents: () => Promise<Context[]>;
```
bryangaleopenfin: Would it be cleaner to name this getReceivedContexts? const receivedContexts = await listener.getReceivedIntents() in the tests feels odd

bryangaleopenfin (approved):

test/demo/utils/fdc3RemoteExecution.ts
tomrobertsOF: Oh wow, how did I miss that.

test/demo/utils/fdc3RemoteExecution.ts
tomrobertsOF: I went back-and-forth a couple of times on this, but was never really happy with either. If you feel strongly I don't mind changing it.

res/test/test-app.html
tomrobertsOF: I pulled this code out of the remote util, which would need to check if something else had added listeners ahead of it, but I suppose for this initial one the checks are not really needed.

test/demo/utils/fdc3RemoteExecution.ts
bryangaleopenfin: Yeah, can see it's odd either way, but on balance I'm in favour of changing it

res/test/test-app.html
bryangaleopenfin: Feel like they're worth dropping then, if only to stop moments of 'huh, why are these here' :p

tomrobertsOF: Yea, already done

test/demo/utils/fdc3RemoteExecution.ts
tomrobertsOF: Ok, done. Just running tests locally then I'll push it up.

ottah (approved):

test/demo/raiseIntent.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        describe('When the target is running', () => {
+            beforeEach(async () => {
+                await fdc3Remote.open(testManagerIdentity, testAppIdentity.uuid);
+            });
+
+            afterEach(async () => {
+                await fin.Application.wrapSync(testAppIdentity).quit().catch(() => {});
+            });
+
+            test('When calling addIntentListner for the first time, the promise resolves and there are no errors', async () => {
```
MichaelMCoates: `addIntentListener`

MichaelMCoates (approved):

PR #56, opened 2019-04-12
[N/A] Context Channel UI

ottah:
Dropdown menu added to all apps to select the context channel the app is on.


tomrobertsOF (approved):

src/demo/apps/DialerApp.tsx
```diff
@@ -77,6 +80,7 @@ export function DialerApp(props: AppProps): React.ReactElement {
 
     return (
         <div>
+            <ContextChannelSelector float={false} />
```
MichaelMCoates: Do you need to pass `float={false}` here if the default is already false?

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
```diff
@@ -0,0 +1,73 @@
...
+            setCurrentChannelId(channel.id);
+        });
+        getAllChannels().then(channels => {
+            setChannels(channels);
+        });
+    }, []);
+
+    const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+        const {value: id} = event.currentTarget;
+        const channel = channels.find(x => x.id === id);
```
MichaelMCoates: Is there something you can use other than `x` here?

src/demo/apps/DialerApp.tsx
ottah: Cheers didn't notice that. Added defaultProps to the component to address this.

pjbroadbent (approved):

PR #66, opened 2019-04-17
[SERVICE-398] Runtime channels and Jenkinsfile updates

pjbroadbent:
Added the `svc-tool channels` step to the build process for any stable (i.e. `master`) builds. This will create several variants of the provider's `app.json` manifest, each targeting a runtime release channel. e.g: `app.runtime-beta.json` - which is the same as the regular `app.json`, only running on the `beta` runtime.

PR also includes some clean-up of the Jenkinsfile script:
* We now use `agent none` at the top of the file, to prevent the job from occupying an additional executor unnecessarily, this could result in deadlock when there were many jobs running at once.
* There's no longer duplication between the staging and stable builds, and the script is in a format that should allow it to be directly copied between projects.
  * The service name is currently still hard-coded within the script, this will be changed once we get a plugin installed that allows the Jenkinsfile to read JSON files.


MichaelMCoates:
When I was prototyping the top-level input stuff, I needed to put it in a top-level stage, with the rest of the build process living in a `stages` declarative. I think that inputs are scoped to the stage they're in. 

Something along these lines:

```javascript

pipeline {

    agent { label 'linux-slave' }
    stages {
        stage('Set Environment Variables') {
            input {
                message "Would you like to deploy the client to NPM?"
                parameters {
                    choice(name: 'DEPLOY_CLIENT', choices: ['Yes', 'No'], description: '') 
                }
            }
            stages {
                stage('Run Tests') {
                    parallel {
                        stage('Unit Tests') {
                            agent { label 'linux-slave' }
                            steps {
                                echo "DEPLOY_CLIENT in second stage ${DEPLOY_CLIENT}"
                                sh "npm i"
                                sh "npm run test:unit"
                                sh "npm run check"
                            }
                            post {
                                always {
                                    junit "dist/test/results-unit.xml"
                                }
                            }
                        }

```

Jenkinsfile
```diff
@@ -34,98 +49,95 @@ pipeline {
...
+    sh "echo ${GIT_SHORT_SHA} > ./dist/SHA.txt"
+
+    sh "npm run zip"
+    sh "npm install bootprint@2.0.1 bootprint-json-schema@2.0.0-rc.3 --no-save"
+    sh "npm run docs"
+}
+
+def addReleaseChannels() {
+    if (env.BRANCH_NAME == 'master') {
+        sh "npm run channels"
```
MichaelMCoates, line 111: I think we're missing the `channels` declaration in `package.json`. Unless we get it automatically from `svc-tools`?

tomrobertsOF (approved):
This all looks fine in terms of the code. Approved on the assumption that it works as expected when run.

pjbroadbent:
> When I was prototyping the top-level input stuff, I needed to put it in a top-level stage, with the rest of the build process living in a `stages` declarative. I think that inputs are scoped to the stage they're in.

After a bit of googling, I found an alternative where you can wrap the 'input' in a script, and then store the result in a variable. This (combined with the `agent none` change) should prevent the job from using an executor whilst waiting for input and/or an integration test runner.

Jenkinsfile
pjbroadbent: My bad - this went through a few revisions, ended up with the script missing. Have added to `package.json` now.

Jenkinsfile
```diff
@@ -1,15 +1,28 @@
...
+        stage('Input') {
+            when {
+                branch 'master'
+                beforeInput true
+            }
+            steps {
+                script {
+                    env.DEPLOY_CLIENT = input \
+                        message: 'Would you like to deploy the client to NPM?', \
+                        parameters: [choice(name: 'DEPLOY_CLIENT', choices: ['Yes', 'No'], description: '')]
```
tomrobertsOF, line 15: how does this work with the new GitHub web hook stuff? Would we need to go into Jenkins each time and click a button? Could we have some kind of timeout to a default response?

Jenkinsfile
```diff
@@ -22,8 +35,8 @@ pipeline {
                 stage('Integration Tests') {
                     agent { label 'win10-dservices' }
                     steps {
-                        bat "npm i"
-                        bat "npm run test:int"
+                        bat "npm install"
+                        bat "npm run test:int -- --color=false --no-cache --verbose"
```
tomrobertsOF, line 40: I don't think these cli args get passed through to jest, unless someone has changed the runner script since I added it.

Jenkinsfile
pjbroadbent: The build would start as soon as the change is pushed, but will then pause and wait for input. The build will remain paused until someone goes into the Jenkins UI and submits this form. This will _only_ happen when merging to `master` though - other builds will run without manual action required, as happens currently.

Since they're only when we're modifying production, I don't think we need any timeout on these kind of actions. If anything, requiring manual approval on these builds is a bonus.

Jenkinsfile
pjbroadbent: This currently only works in layouts, but these arguments should be getting added to all services once integration tests are added to the tooling (currently WIP).

tomrobertsOF: Ok, may as well leave it then.

Jenkinsfile
tomrobertsOF: Ah ok. If it's only going to prompt for master then that's fine. It's unlikely we'd be doing an unattended production release

tomrobertsOF (approved):

bryangaleopenfin (approved):

PR #67, opened 2019-04-18
[Ignore][SERVICE-453] Fix random test failures around channel-changed events

bryangaleopenfin:
Note this current attempt at a solution does not fix the problem - https://appoji.jira.com/browse/SERVICE-453

### DIFF
--- a/src/provider/APIHandler.ts
+++ b/src/provider/APIHandler.ts
@@ -65,9 +65,9 @@ export class APIHandler<A extends string, P extends {[K in A]: unknown}, R exten
 
         // The 'onConnection' callback fires *just before* the channel is ready.
         // Delaying the firing of our signal slightly, to ensure client is definitely contactable.
-        setImmediate(() => {
+        setTimeout(() => {
             this.onConnection.emit(app);
-        });
+        }, 100);
     }
 
     private onDisconnectionHandler(app:Identity): void {
--- a/src/provider/FDC3.ts
+++ b/src/provider/FDC3.ts
@@ -68,7 +68,10 @@ export class FDC3 {
         const requestedApp: Application | undefined = applications.find((app: Application) => app.name === payload.name);
         return new Promise<void>((resolve: () => void, reject: (reason: Error) => void) => {
             if (requestedApp) {
-                this.openApplication(requestedApp, payload.context).then(resolve, reject);
+                this.openApplication(requestedApp, payload.context).then(() => {
+                    // Delay so that we can be the app has joined the global channel before returning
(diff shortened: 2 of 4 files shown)


src/provider/FDC3.ts
```diff
@@ -68,7 +68,10 @@ export class FDC3 {
         const requestedApp: Application | undefined = applications.find((app: Application) => app.name === payload.name);
         return new Promise<void>((resolve: () => void, reject: (reason: Error) => void) => {
             if (requestedApp) {
-                this.openApplication(requestedApp, payload.context).then(resolve, reject);
+                this.openApplication(requestedApp, payload.context).then(() => {
+                    // Delay so that we can be the app has joined the global channel before returning
```
MichaelMCoates, line 72: `we can be the app`

bryangaleopenfin: Meant to be 'we can be sure the app'! 👍

src/provider/APIHandler.ts
```diff
@@ -65,9 +65,9 @@ export class APIHandler<A extends string, P extends {[K in A]: unknown}, R exten
 
         // The 'onConnection' callback fires *just before* the channel is ready.
         // Delaying the firing of our signal slightly, to ensure client is definitely contactable.
-        setImmediate(() => {
+        setTimeout(() => {
             this.onConnection.emit(app);
-        });
+        }, 100);
```
tomrobertsOF, line 70: Is 100 the lowest number that will work here? I would prefer to have as low of a delay as we can get away with.

test/demo/contextChannels.test.ts
```diff
@@ -2,6 +2,7 @@ import 'jest';
 import {connect, Fin, Identity, Application} from 'hadouken-js-adapter';
 
 import * as fdc3Remote from './utils/fdc3RemoteExecution';
+import {delay} from './utils/delay';
```
tomrobertsOF, line 5: You import but don't use delay.

test/demo/contextListener.test.ts
```diff
@@ -3,6 +3,7 @@ import {OrganizationContext} from '../../src/client/main';
 
 import {fin} from './utils/fin';
 import * as fdc3Remote from './utils/fdc3RemoteExecution';
+import {delay} from './utils/delay';
```
tomrobertsOF, line 6: as above, imported but not used.

test/demo/contextChannels.test.ts
tomrobertsOF: (unless I missed it somewhere)

PR #68, opened 2019-04-23
[SERVICE-424] AppDirectory

ottah:
This PR adds App Directory functionality.

- Offline cache: stores app directory url and apps.
- The app directory is refreshed when the URL is different, or if the directory is an empty array requests are made to see if there is new data.

Cache lifetime needs to be added.


test/tsconfig.json
```diff
@@ -10,6 +10,7 @@
         "es7"
       ],
       "esModuleInterop": false,
+      "experimentalDecorators": true
```
ottah: This flag allows @injectable to work with jest.

src/provider/model/AppDirectory.ts
```diff
@@ -1,42 +1,118 @@
...
+            }
+            this._directory = value;
+        });
+    }
+
+    /**
+     * Update the application directory in memory and storage.
+     * @param applications To place into the directory.
+     */
+    private async updateDirectory(applications: Application[]): Promise<void> {
```
pjbroadbent: For readability/consistency, can you move these down a bit? Helps to have public methods first, then protected/private.

src/provider/model/AppDirectory.ts
```diff
@@ -1,42 +1,118 @@
...
+        // If using the demo app directory in production, set an empty array.
+        if (process.env.NODE_ENV === 'production') {
+            this.updateDirectory([]);
+            return;
+        }
+        const storedURL = await localForage.getItem<string>('url');
+        const currentURL = this._URL;
+        // Only update the directory if the URL has changed
+        // or the directory is empty (the last request may have failed?)
+        if (currentURL !== storedURL || (currentURL === storedURL && this._directory.length === 0)) {
```
pjbroadbent: Check is a bit redundant - `currentURL !== storedURL || this._directory.length === 0`?

src/provider/model/AppDirectory.ts
```diff
@@ -1,42 +1,118 @@
...
+                value = devURL;
+            }
+            this._URL = value;
+        });
+        this._store.getItem<Application[]>(StorageKeys.APPLICATIONS, (error, value) => {
+            if (value === null) {
+                value = [];
+            }
+            this._directory = value;
+        });
```
pjbroadbent: What would the back-end be here? An IndexedDB? Not really seeing what benefits using localForage gives over just using localStorage directly - would be synchronous, and have less boilerplate.

Any reason for the callback style here? I get we can't use async, but could still be promise-based.

src/provider/model/AppDirectory.ts
```diff
@@ -1,42 +1,118 @@
...
+        if (currentURL !== storedURL || (currentURL === storedURL && this._directory.length === 0)) {
+            try {
+                const applications = await this.fetchData(currentURL);
+                await this.updateDirectory(applications);
+                if (currentURL !== storedURL) {
+                    await this.updateURL(currentURL);
                 }
+            } catch (error) {
+                // Dont update
+                return;
```
pjbroadbent: What errors are we expecting here, any reason not to propagate upward? Completely ignoring an error like this is a bit scary-looking... :-)

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,109 @@
...
+            const applications = await store.getItem(StorageKeys.APPLICATIONS);
+            expect(url).toEqual(null);
+            expect(applications).toEqual(null);
+        });
+
+        describe('Storing fetchData', () => {
+            it('Correctly sets applications in the store', async () => {
+                appDirectory['fetchData'] = jest.fn(() => Promise.resolve([]));
+                await appDirectory.getAllApps();
+                const applications = await store.getItem(StorageKeys.APPLICATIONS);
```
pjbroadbent: Does this work?.. `store` at this point would be an object that was created in the previous test run. Seems like it would always be empty (since it's cleared in the afterEach, and then never re-populated)?

The implementation also relies on the `'Correctly sets up the store'` test to run first - you wouldn't be able to use `.only` or `.skip` with these tests.

Every test should be fully independent - would make sense to me to add a `beforeEach` at the same level as the `store.clear()` `afterEach` that creates the directory and extracts the store object. Not sure exactly where `localForage` is saving the data in this case - but if it's `localStorage`, the test could also query the store directly rather than extracting the private field from `appDirectory`.

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,109 @@
...
+    describe('Storage', () => {
+        beforeEach(async () => {
+            appDirectory = new AppDirectory();
+        });
+
+        afterEach(async () => {
+            await store.clear();
+        });
+
+        it('Correctly sets up the store', async () => {
```
pjbroadbent: Can you set up the tests so that the `describe`s and `it`s read as a (fairly concise) sentence? The describes should describe the context, and the it's the expected state/behaviour of the app when it's in that state.

For example, this test has the "full" name: `AppDirectory Unit Tests > Storage > Correctly sets up the store`. Whilst keeping the exact same code (only changing the describe/it strings), this could be more descriptively titled `AppDirectory > When directory is first initialised > Nothing exists within storage`.

Naming in this way also points out some edge cases - this test could fail if I run it immediately after running the demo app, since the storage would persist from the demo app into the test environment. There's also no tests in here that assume there is something in storage when the service starts.

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,109 @@
...
+            it('Correctly sets applications in the store', async () => {
+                appDirectory['fetchData'] = jest.fn(() => Promise.resolve([]));
+                await appDirectory.getAllApps();
+                const applications = await store.getItem(StorageKeys.APPLICATIONS);
+                expect(applications).toEqual([]);
+            });
+
+            it('Correctly stores fetchData in the store and in memory', async () => {
+                const fakeApps = generateApplications(3);
+                appDirectory['fetchData'] = jest.fn(() => Promise.resolve(fakeApps));
```
pjbroadbent: This is a potentially risky way to test the fetch functionality. We're testing that the directory correctly fetches app data, but also replacing the code within the directory that performs that fetch.

This kind of mocking/stubbing should really only be used on things outside of the class being tested - so for example, leave the class completely unmodified, but mock the `fetch` method - since this will change the behaviour of the application without modifying the actual code that's being tested.

Of course, replacing a global such as `fetch` is also potentially risky, as it would leak across tests if not careful. Make sure the "real" `fetch` function is saved first and restored afterward - we could do with some test utils here.

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,109 @@
...
+    });
+
+    describe('Calls to', () => {
+        const fakeApps = generateApplications(3);
+        beforeEach(async () => {
+            appDirectory = new AppDirectory();
+            appDirectory['fetchData'] = jest.fn(() => Promise.resolve(fakeApps));
+        });
+
+        it('getAppByName', async () => {
```
pjbroadbent: As above, can we make the naming clearer here? Consider both the names of the `describe` blocks and the tests themselves.

src/provider/model/AppDirectory.ts
```diff
@@ -2,40 +2,101 @@ import {injectable} from 'inversify';
...
+
+// Demo development app directory url
+const devURL = 'http://localhost:3923/provider/sample-app-directory.json';
+
+
 @injectable()
 export class AppDirectory {
-    private _directory!: Application[];
+    private _directory: Application[] = [];
+    private _URL!: string;
```
bryangaleopenfin: Think 'variables start with a lower case letter' trumps 'URL is an acronym' here

src/provider/model/AppDirectory.ts
```diff
@@ -2,40 +2,101 @@ import {injectable} from 'inversify';
 
 import {Application} from '../../client/directory';
 
+enum StorageKeys {
+    URL = 'url',
+    APPLICATIONS = 'applications'
+}
+
+// Demo development app directory url
+const devURL = 'http://localhost:3923/provider/sample-app-directory.json';
```
bryangaleopenfin: I know we're not entirely consistent, but I think `Url` is more common in our code than `URL` (eg., `manifestUrl`, `getProviderUrl`). Would be good to lean towards consistency. (Though I don't care about this as much as I care about `_url` vs `_URL`)

src/provider/model/AppDirectory.ts
```diff
@@ -2,40 +2,101 @@ import {injectable} from 'inversify';
...
 
-    public async getAppByName(name: string): Promise<Application|null> {
-        await this.fetchData();
+    public constructor() {
+        // Set default values
+        this._URL = localStorage.getItem(StorageKeys.URL) || devURL;
+        const directory = localStorage.getItem(StorageKeys.APPLICATIONS) || [];
+        try {
+            this._directory = (typeof directory === 'string') ? JSON.parse(directory) : directory;
+        } catch (error) {
```
bryangaleopenfin: I would write this:

> const directory = localStorage.getItem(StorageKeys.APPLICATIONS) || [];
> try {
> this._directory = (typeof directory === 'string') ? JSON.parse(directory) : directory;
> } catch (error) {

as:

> const directory = localStorage.getItem(StorageKeys.APPLICATIONS);
> try {
> this._directory = (directory !== null) ? JSON.parse(directory) : [];
> } catch (error) {

Like, we're "ambiguating" which type `directory` is, only to immediately need to disambiguate it just below, which feels odd.

I'd also suggest renaming `directory` to `storedDirectory` maybe? It's natural to assume `_directory` and `directory` are the same type even though they're not

src/provider/model/AppDirectory.ts
```diff
@@ -2,40 +2,101 @@ import {injectable} from 'inversify';
...
+        }
+    }
+
+    /**
+     * Refresh the AppDirectory.
+     */
+    private async refreshDirectory(): Promise<void> {
+        // If using the demo app directory in production, set an empty array.
+        if (process.env.NODE_ENV === 'production') {
+            this.updateDirectory([]);
```
bryangaleopenfin: `await` here? (Or make `updateDirectory` non-`async`)

src/provider/model/AppDirectory.ts
```diff
@@ -2,40 +2,101 @@ import {injectable} from 'inversify';
...
+            return response.json();
+        } else {
+            throw new Error('Error fetching app directory: ' + response.statusText);
+        }
+    }
+
+    /**
+     * Refresh the AppDirectory.
+     */
+    private async refreshDirectory(): Promise<void> {
```
bryangaleopenfin: A rough rule-of-thumb I try to follow is that methods call downwards, so something like `refreshDirectory` would come above `fetchData` or `updateDirectory`, so as you're reading file, you get into progressively more fine-grained detail. Could be worth re-ordering here.

src/provider/model/AppDirectory.ts
```diff
@@ -2,40 +2,101 @@ import {injectable} from 'inversify';
...
+        // If using the demo app directory in production, set an empty array.
+        if (process.env.NODE_ENV === 'production') {
+            this.updateDirectory([]);
+            return;
+        }
+        const storedURL = localStorage.getItem(StorageKeys.URL);
+        const currentURL = this._URL;
+        // Only update the directory if the URL has changed
+        // or the directory is empty (the last request may have failed?)
```
bryangaleopenfin: Could also be the case if this is the first time we're reading the directory, so think worth mentioning that if we're mentioning for failed request case.

And do we want to worry about the corner case of a genuinely empty directory, which would break the"empty arrive === failed request so try" logic?

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,203 @@
...
+import {AppDirectory} from '../../src/provider/model/AppDirectory';
+import {Application} from '../../src/client/directory';
+
+
+enum StorageKeys {
+    URL = 'url',
+    APPLICATIONS = 'applications'
+}
+
+enum Intents {
```
bryangaleopenfin: Can we be consistent on if we're capitalizing enum entries?

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,203 @@
...
+
+type Store = jest.Mocked<Pick<typeof localStorage, 'getItem' | 'setItem'>>;
+
+interface ResponseOptions {
+    status: number;
+}
+
+class Response {
+    private _blob: string;
+    private _options: ResponseOptions;
```
bryangaleopenfin: I'd really like it if we use `readonly` wherever we can, similar to `const`

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,203 @@
...
+        describe('Refreshing the directory', () => {
+            it('Fetches & updates the directory when URLs do not match', async () => {
+                const appDirectory = new AppDirectory();
+                const newApplications = [{...appTemplate, name: 'test-app'}];
+                const spyGetItem = jest.spyOn(global.localStorage, 'getItem');
+                const spyFetch = jest.spyOn(global, 'fetch').mockImplementation(async () => {
+                    return new Response(JSON.stringify(newApplications), {status: 200});
+                });
+                spyGetItem.mockImplementation(() => '__test_url__');
+                appDirectory['_URL'] = '*different_url*';
```
bryangaleopenfin: In real world usage, what would be the right way to change what URL `AppDirectory` uses? Just want to check this is what we want to do here, rather than setting a new URL via localStorage?

bryangaleopenfin (commented):
Looks good. Definitely feels weird that we're handling changing URLs without URLs being configurable yet, but yeah, that's another story.

Got some style feedback (which is super subjective, so feel free to push back on), but the logic looks solid and the tests look good.

test/provider/AppDirectory.unittest.ts
pjbroadbent: Exact details are still TBD, but at the high level will be using the config mechanisms from layouts. A JSON schema will define the available options, and the directory will query the config store for the URL.

pjbroadbent (approved):

test/provider/AppDirectory.unittest.ts
bryangaleopenfin: So are we content that this is a good way to change the URL for the purposes of this unit test?

tomrobertsOF: I think it's probably fine for now. I'd rather have a test which works now but needs updating than wait until we have everything perfect to test with.

bryangaleopenfin: 👍

bryangaleopenfin (approved):

tomrobertsOF (approved):
Nice updates to the tests and fetching stuff. Looks good to me!

PR #70, opened 2019-04-24
[SERVICE-458] Prevent applications from receiving their own broadcasts

tim-dinsdale:
So I know that the name is wrong, and you need to tell me what to put there. The rest is AWESOME

### DIFF
--- a/test/demo/contextListener.test.ts
+++ b/test/demo/contextListener.test.ts
@@ -104,7 +104,7 @@ describe('Context listeners and broadcasting', () => {
                 });
 
                 test('When calling broadcast, only the still-registered listener is triggered', async () => {
-                    await fdc3Remote.broadcast(testAppIdentity, validContext);
+                    await fdc3Remote.broadcast(testManagerIdentity, validContext);
                     const receivedContexts = await Promise.all(listeners.map(listener => listener.getReceivedContexts()));
 
                     // First listener not triggered
@@ -118,7 +118,7 @@ describe('Context listeners and broadcasting', () => {
                 test('A third listener can be registered and triggered as expected', async () => {
                     const newListener = await fdc3Remote.addContextListener(testAppIdentity);
 
-                    await fdc3Remote.broadcast(testAppIdentity, validContext);
+                    await fdc3Remote.broadcast(testManagerIdentity, validContext);
                     const receivedContexts = await newListener.getReceivedContexts();
 
                     expect(receivedContexts.length).toBe(1);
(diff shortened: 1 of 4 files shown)


pjbroadbent (approved):

tomrobertsOF:
Now that the behaviour is no longer TBD, can we fill out implementations for the two tests in `contextLIstener.test.ts` (lines 57 and 94) so we can avoid breaking this in future.

test/demo/contextListener.test.ts
```diff
@@ -91,7 +97,14 @@ describe('Context listeners and broadcasting', () => {
                 }
             });
 
-            test.todo('When calling broadcast from the first app [behavior TBD - do apps receive their own broadcasts?]');
+            test('When calling broadcast from the first app', async () => {
+                await fdc3Remote.broadcast(testAppIdentity, validContext);
+
+                const receivedContexts = await Promise.all(listeners.map(listener => listener.getReceivedContexts()));
+                receivedContexts.forEach((rc)=>console.log(rc));
```
bryangaleopenfin: Is this logging intentional or leftover debugging? And if intentional, would put spaces around '=>' for consistency

bryangaleopenfin (approved):
Approved, but just want to check about the logging

test/demo/contextListener.test.ts
```diff
@@ -54,7 +54,13 @@ describe('Context listeners and broadcasting', () => {
                 expect(receivedContexts[0]).toEqual(validContext);
             });
 
-            test.todo('When broadcast is called from the same app [behavior TBD - do apps receive their own broadcasts?]');
+            test('When broadcast is called from the same app', async () => {
```
tomrobertsOF: Just a quick note on the test titles. It looks like you just removed the placeholder part of the name but didn't replace it with anything. If you look at the others we have something along the lines of `When doing this thing, that stuff will happen`, so for this case I would prefer something like `When broadcast is called from the same app the listener is not triggered`

test/demo/contextListener.test.ts
```diff
@@ -91,7 +97,14 @@ describe('Context listeners and broadcasting', () => {
                 }
             });
 
-            test.todo('When calling broadcast from the first app [behavior TBD - do apps receive their own broadcasts?]');
+            test('When calling broadcast from the first app', async () => {
+                await fdc3Remote.broadcast(testAppIdentity, validContext);
+
+                const receivedContexts = await Promise.all(listeners.map(listener => listener.getReceivedContexts()));
+                receivedContexts.forEach((rc)=>console.log(rc));
+                expect(receivedContexts.length).toBe(2);
```
tomrobertsOF: This length check is covered by the equals check below it, so not really necessary.

rypete (approved):
Looks good

tomrobertsOF (approved):

tomrobertsOF:
Passed on CI, so going to go ahead and merge.

PR #71, opened 2019-04-25
[SERVICE-456] Implement findIntentsByContext

MrP:
Implements https://appoji.jira.com/browse/SERVICE-456

- Adds an implementation of the FDC3 API method findIntentsByContext
- Changes the "Contacts" demo app to use it to display the contact actions.
- As part of that, it introduces a "customConfig.icon" property in the intents in the app directory.  This usage is allowed by the spec, and a perfectly valid example of how to do it, but not intended as the prescriptive way to deal with intent icons.  UI is out of scope for the FDC3 v1 standard.

This touches the same files as https://github.com/HadoukenIO/fdc3-service/pull/68.  
It may make more sense to merge 68 first, then deal with any conflicts here.

### DIFF
--- a/src/demo/apps/ContactsApp.tsx
+++ b/src/demo/apps/ContactsApp.tsx
@@ -54,9 +54,9 @@ export function ContactsApp(): React.ReactElement {
             type: 'fdc3.contact',
             name: '',
             id: {}
-        }).then(appIntents => {
-            console.log('setAppIntents', appIntents);
-            setAppIntents(appIntents);
+        }).then(foundAppIntents => {
+            console.log('setAppIntents', foundAppIntents);
+            setAppIntents(foundAppIntents);
         });
     }, []);
 
--- a/src/demo/apps/ContactsApp.tsx
+++ b/src/demo/apps/ContactsApp.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import * as fdc3 from '../../client/main';
 import {ContactsTable} from '../components/contacts/ContactsTable';
-import {Context, ContactContext} from '../../client/main';
+import {Context, ContactContext, AppIntent} from '../../client/main';
 import '../../../res/demo/css/w3.css';
 import {ContextChannelSelector} from '../components/ContextChannelSelector/ContextChannelSelector';
 
@@ -48,6 +48,18 @@ export function ContactsApp(): React.ReactElement {
         }
     }
 
+    const [appIntents, setAppIntents] = React.useState([] as AppIntent[]);
+    React.useEffect(() => {
+        fdc3.findIntentsByContext({
+            type: 'fdc3.contact',
+            name: '',
+            id: {}
+        }).then(appIntents => {
+            console.log('setAppIntents', appIntents);
+            setAppIntents(appIntents);
+        });
+    }, []);
+
     React.useEffect(() => {
         document.title = 'Contacts';
     }, []);
@@ -73,7 +85,7 @@ export function ContactsApp(): React.ReactElement {
     return (
         <React.Fragment>
             <ContextChannelSelector />
-            <ContactsTable items={contacts} />
+            <ContactsTable items={contacts} appIntents={appIntents} />
         </React.Fragment>
     );
 }
--- a/src/demo/components/contacts/ContactsRow.tsx
+++ b/src/demo/components/contacts/ContactsRow.tsx
@@ -4,31 +4,34 @@ import * as fdc3 from '../../../client/main';
 import {ContactContext} from '../../../client/context';
 import {Contact} from '../../apps/ContactsApp';
 import {IntentButton} from '../common/IntentButton';
-
 import './ContactsRow.css';
-
+import {AppIntent} from '../../../client/main';
 
 interface ContactRowProps {
     item: Contact;
     selected: boolean;
     handleSelect: (item: Contact | null) => void;
+    appIntents: AppIntent[];
 }
 
 export function ContactsRow(props: ContactRowProps): React.ReactElement {
     const {item, selected, handleSelect} = props;
 
-    const handleDial = async (): Promise<void> => {
+    const handleAppIntent = async (appIntent: AppIntent): Promise<void> => {
         if (handleSelect) {
             handleSelect(null);
         }
-        await fdc3.raiseIntent(fdc3.Intents.DIAL_CALL, getContext());
+        await fdc3.raiseIntent(appIntent.intent.name, getContext());
     };
 
-    const handleCall = async (): Promise<void> => {
-        if (handleSelect) {
-            handleSelect(null);
+    const getIntentIcon = (appIntent: AppIntent): string => {
+        if (appIntent && appIntent.apps.length > 0 && appIntent.apps[0].intents) {
+            const intent = appIntent.apps[0].intents.find(intent => intent.name===appIntent.intent.name);
+            if (intent && intent.customConfig) {
+                return intent.customConfig.icon;
+            }
         }
-        await fdc3.raiseIntent(fdc3.Intents.START_CALL, getContext());
+        return 'fa-file';
     };
 
     const handleClick = (event: React.MouseEvent<HTMLElement>) => {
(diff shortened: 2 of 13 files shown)


src/demo/apps/ContactsApp.tsx
```diff
@@ -48,6 +48,18 @@ export function ContactsApp(): React.ReactElement {
         }
     }
 
+    const [appIntents, setAppIntents] = React.useState([] as AppIntent[]);
+    React.useEffect(() => {
+        fdc3.findIntentsByContext({
+            type: 'fdc3.contact',
+            name: '',
+            id: {}
+        }).then(appIntents => {
```
bryangaleopenfin: Would try to avoid the name shadowing here

src/demo/components/contacts/ContactsRow.tsx
```diff
@@ -4,31 +4,34 @@ import * as fdc3 from '../../../client/main';
...
-        await fdc3.raiseIntent(fdc3.Intents.DIAL_CALL, getContext());
+        await fdc3.raiseIntent(appIntent.intent.name, getContext());
     };
 
-    const handleCall = async (): Promise<void> => {
-        if (handleSelect) {
-            handleSelect(null);
+    const getIntentIcon = (appIntent: AppIntent): string => {
+        if (appIntent && appIntent.apps.length > 0 && appIntent.apps[0].intents) {
+            const intent = appIntent.apps[0].intents.find(intent => intent.name===appIntent.intent.name);
```
bryangaleopenfin: Spaces around `===`. (Does that not get picked up by `npm run fix`?)

src/provider/model/AppDirectory.ts
```diff
@@ -20,6 +21,32 @@ export class AppDirectory {
...
+                            apps: [app]
+                        };
+                    }
+                }
+            });
+        });
+        Object.values(appIntentsByName).forEach(appIntent => {
+            appIntent.apps.sort((a, b) => a.appId.localeCompare(b.appId));
+        });
+        return Object.values(appIntentsByName).sort((a, b) => a.intent.name.localeCompare(b.intent.name));
```
bryangaleopenfin: I'd feel safer if this just used plain `sort`.  I wouldn't introduce locale-dependence at this level, and if a consumer really cares about sorting in a particular locale, they can do that at the presentation level

test/demo/findIntentsByContext.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    beforeEach(async () => {
+        // The main launcher app should remain running for the duration of all tests.
+        await expect(fin.Application.wrapSync(testManagerIdentity).isRunning()).resolves.toBe(true);
+    });
+
+    describe('When calling findIntentsByContext with an invalid context', () => {
+        const invalidContext = {
+            twitter: '@testname'
+        };
+        test('the promise rejects', async () => {
```
bryangaleopenfin: Think these should start with a capital letter for consistency with other tests

test/demo/findIntentsByContext.test.ts
```diff
@@ -0,0 +1,64 @@
...
+            // @ts-ignore
```
bryangaleopenfin, line 25: What do we need @ts-ignore here for?

test/demo/findIntentsByContext.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        test('the promise resolves to an array of all compatible AppIntents', async () => {
+            const receivedAppIntents = await fdc3Remote.findIntentsByContext(testManagerIdentity, context);
+            expect(receivedAppIntents).toEqual([
+                {
+                    intent: {
+                        displayName: 'Dial',
+                        name: 'DialCall'
+                    },
+                    apps: [
+                        expect.objectContaining({
```
bryangaleopenfin, line 58: Huh, I did not know we could do this! Cool

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,103 @@
...
+});
+
+describe('Given an App Directory with apps', () => {
+    const appDirectory = new AppDirectory();
+
+    describe('When finding app intents by context with a context implemented by 2 intents in both apps', () => {
+        beforeEach(() => {
+            mockJson.mockResolvedValue(fakeAppDirectory);
+        });
+        it('Should return 2 intents', async () => {
```
bryangaleopenfin, line 63: Please can we make a call on if we're using `it()` or `test()` and stick to one or the other? @tomrobertsOF - thoughts?

test/provider/AppDirectory.unittest.ts
```diff
@@ -0,0 +1,103 @@
...
+                    apps: [fakeApp1]
+                },
+                {
+                    intent: {name: 'testIntent.StartChat', displayName: 'testIntent.StartChat'},
+                    apps: [fakeApp1, fakeApp2]
+                }
+            ] as AppIntent[]);
+        });
+    });
+    describe('When finding app intents by context with a context not implemented by any intent', () => {
```
bryangaleopenfin, line 79: Would put a line break between each of the describe blocks for readability

bryangaleopenfin (commented):
Looks good, just a few minor comments

test/provider/AppDirectory.unittest.ts
MrP: done

test/demo/findIntentsByContext.test.ts
MrP: The test is to check that sending an invalid context throws, which is caught by typescript, but is still possible to receive an invalid context in the service.
I've added a comment there now

test/demo/findIntentsByContext.test.ts
MrP: done

src/demo/components/contacts/ContactsRow.tsx
MrP: done.  `npm run fix` didn't do anything, no.

src/demo/apps/ContactsApp.tsx
MrP: done

test/demo/findIntentsByContext.test.ts
bryangaleopenfin: Ah, makes sense. Would probably favour dropping the comment and @ts-ignore, and using something like `invalidContext as unknown as Context` then, which draws attention to what's happening here

src/provider/model/AppDirectory.ts
MrP: What I wanted to do here was make sure we always return the intents and the apps within each intent in a predictable and deterministic order.  It helps with tests and I think it's nicer on the consumers of the API.  The spec doesn't say anything about order, that's true.

And you're right, at whatever presentation layer this ends up being used, that's where any smarter sorting should be done.

As for `localeCompare`, I think that's just the standard way to compare strings in JS, even if you don't care about the locales.

test/provider/AppDirectory.unittest.ts
MrP: Fair enough.  I'd prefer `test`, myself.  
I copied this from the other unit tests file, MetadataStore.unittest.ts, so I'd say it can be done in a later PR that changes them all.

src/provider/model/AppDirectory.ts
bryangaleopenfin: Ah, sure, I agree that it should be in a predictable and deterministic order, but the result of localeCompare differs depending on the locale of the host machine, right? (And I get its unlikely to make any difference in practice, but it still feels like it would be better not to lean on a method that by design differs depending on the user's locale).

MrP: You're right, it does depend on the language set in the browser.  As it's not a sort that would affect presentation, I've fixed by always setting the locale to `'en'` in `localeCompare`, what do you think?

bryangaleopenfin: 👍

bryangaleopenfin (approved):

tomrobertsOF (approved):

pjbroadbent (approved):

PR #72, opened 2019-04-29
[SERVICE-457] Clear the channel context when the last window is removed from a channel

tim-dinsdale:


test/demo/contextChannels.test.ts
```diff
@@ -246,6 +246,32 @@ describe('When joining a channel', () => {
         expect(payload[0]).toHaveProperty('previousChannel.id', 'blue');
         expect(payload[0]).toHaveProperty('identity', channelChangingWindow);
     });
+
+    it('if everything is unsubscribed, and something rejoins, there is no data left over', async ()=>{
```
bryangaleopenfin: Capitalize for consistency

test/demo/contextChannels.test.ts
```diff
@@ -246,6 +246,32 @@ describe('When joining a channel', () => {
...
+        const [greenWindow1, channelChangingWindow] = await setupWindows('green', 'yellow');
```
bryangaleopenfin: Is there better naming as they both change channel?

test/demo/contextChannels.test.ts
```diff
@@ -246,6 +246,32 @@ describe('When joining a channel', () => {
...
+        const channelChangingWindowListener = await fdc3Remote.addContextListener(channelChangingWindow);
+        await fdc3Remote.broadcast(greenWindow1, testContext);
+
+        let receivedContexts = await channelChangingWindowListener.getReceivedContexts();
+        expect(receivedContexts).toHaveLength(0);
+
+        await fdc3Remote.joinChannel(channelChangingWindow, 'green');
+        receivedContexts = await channelChangingWindowListener.getReceivedContexts();
+        expect(receivedContexts).toHaveLength(1);
```
bryangaleopenfin: Would do this as toEqual([testContext]), to be explicit about what we expect to have received

test/demo/contextChannels.test.ts
```diff
@@ -246,6 +246,32 @@ describe('When joining a channel', () => {
...
+        const members = await fdc3Remote.getChannelMembers(channelChangingWindow, 'green');
+        expect(members).toEqual([]);
+
+        receivedContexts = await channelChangingWindowListener.getReceivedContexts();
+        expect(receivedContexts).toHaveLength(1);
+
+        await fdc3Remote.joinChannel(channelChangingWindow, 'green');
+
+        receivedContexts = await channelChangingWindowListener.getReceivedContexts();
+        expect(receivedContexts).toHaveLength(1);
```
bryangaleopenfin: This caught me out for a moment thinking "surely we should be testing it's zero...". Could be worth a comment like "check no we've received no more contexts"?

bryangaleopenfin (approved):
Approved with comments

test/demo/contextChannels.test.ts
pjbroadbent: From the test description, it seems one channel would suffice for this - is there a reason for using two?

If there's some subtle edge case being covered here, would be good to make that explicit - or even a separate test.

tim-dinsdale: Ah OK. So this test could probably do with some docs. What I am testing is this:

1 - if I join a channel where there is data, I get the listener called 
2 - if I join a channel where there is no data, I get no listener callback.

Which are obviously just two sides of the same coin. 1 is quite hard to test because of the way the setupWindows function joins the channel - I miss the first callback, because it has already happened before I register. So I set up in some dummy channel (yellow) and then join green, and get my callback. Then I clear the green channel by switching everyone away, and when I rejoin I can test 2.

So, Phil, is your question, should I just use yellow rather than global so I have two channels rather than getting global involved?

test/demo/contextChannels.test.ts
tim-dinsdale: Given how confused I was by this when I first wrote it, I think 'it's a fair cop, guv' is the only reasonable response

test/demo/contextChannels.test.ts
tim-dinsdale: I think I have done what you asked

tomrobertsOF (approved):

pjbroadbent (approved):

PR #81, opened 2019-05-07
[SERVICE-460] FDC3 Errors (foundation)

spcfran:
This PR sets up the infrastructure for FDC3 Errors
- FDC3Error class with `code` and `message` fields, and static methods to serialize/deserialize as a plain Error (with just a string) that can be sent over the wire
- Use of serialize/deserialize at both ends of the communication (client / provider)
- Necessary boilerplate to make the whole thing work in tests (as Puppeteer hijacks errors and appends its own stuff), both on test apps and on the fdc3RemoteExecution side
- Implementation of a couple of FDC3Errors for `raiseIntent` and `findIntentsByContext`
- Made a few renamings / type annotation changes while on a side quest (sorry for the offtopic)

### DIFF
--- a/src/client/errors.ts
+++ b/src/client/errors.ts
@@ -1,0 +1,21 @@
+export enum OpenError {
+  AppNotFound = 'AppNotFound',
+  ErrorOnLaunch = 'ErrorOnLaunch',
+  AppTimeout = 'AppTimeout',
+  ResolverUnavailable = 'ResolverUnavailable'
+}
+
+export enum ResolveError {
+  NoAppsFound = 'NoAppsFound',
+  ResolverUnavailable = 'ResolverUnavailable',
+  ResolverTimeout = 'ResolverTimeout',
+  InvalidContext = 'InvalidContext'
+}
+
+export class FDC3Error extends Error{
+  public code: string;
+  public constructor(code: string, message: string) {
+      super(message);
+      this.code = code;
+  }
+}
--- a/src/client/errors.ts
+++ b/src/client/errors.ts
@@ -1,18 +1,18 @@
 export enum OpenError {
-  AppNotFound = 'AppNotFound',
-  ErrorOnLaunch = 'ErrorOnLaunch',
-  AppTimeout = 'AppTimeout',
-  ResolverUnavailable = 'ResolverUnavailable'
+    AppNotFound = 'AppNotFound',
+    ErrorOnLaunch = 'ErrorOnLaunch',
+    AppTimeout = 'AppTimeout',
+    ResolverUnavailable = 'ResolverUnavailable'
 }
 
 export enum ResolveError {
-  NoAppsFound = 'NoAppsFound',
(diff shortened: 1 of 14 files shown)


tomrobertsOF (approved):

src/client/errors.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export class FDC3Error extends Error {
+    /**
+     * If error is FDC3 specific, serialize it as { code, message } so it can be identified as an `FDC3Error` at the client's end.
+     * Otherwise return the error itself
+     * @param error The error
+     */
+    public static serialize(error: Error | FDC3Error): Error {
+        // TODO: is use of `instanceof` OK? or should use `if (error && error.code && error.message)` ?
```
bryangaleopenfin, line 22: Are you able to make a call on this and get rid of the `TODO`? `instanceof` _should_ be fine right? Do we expect this to be called with anything that's come across the wire but hasn't been properly deserialized?

bryangaleopenfin (approved):
Looks good, but would be good to get rid of that TODO before this goes in

src/client/errors.ts
spcfran: Yes! Actually that (and a bunch more TODO's) are addressed in the [second part of the story](https://github.com/spcfran/fdc3-service/pull/1/commits/e31ddb8be86639d8322c6171166add5b55991688)

@pjbroadbent and I agreed `instanceof` is ok

I plan to submit that PR as soon as this one is merged into `feature`, so I guess it's OK to skip it here for now?

test/demo/raiseIntent.test.ts
```diff
@@ -89,10 +90,12 @@ describe('Intent listeners and raising intents', () => {
             });
 
             describe('When the target is *not* registered to accept the raised intent', () => {
-                test('When calling raiseIntent the promise rejects with an error', async () => {
+                test('When calling raiseIntent the promise rejects with an FDC3Error', async () => {
                     const resultPromise = fdc3Remote.raiseIntent(testManagerIdentity, invalidPayload.intent, invalidPayload.context, testAppIdentity.appId);
                     // TODO: decide if this is this the error message we want when sending a targeted intent
                     await expect(resultPromise).rejects.toThrowError(/No applications available to handle this intent/);
+                    await expect(resultPromise).rejects.toHaveProperty('name', 'FDC3Error');
+                    await expect(resultPromise).rejects.toHaveProperty('code', ResolveError.NoAppsFound);
```
pjbroadbent, line 98: For part 2, consider if it's worth making a util for this - going to be a lot of very similar asserts.

pjbroadbent (approved):

test/demo/raiseIntent.test.ts
spcfran: Implemented on [this commit](https://github.com/spcfran/fdc3-service/commit/aeb68a8c83919754c10f9b25d8dcdd347ce61988), will go in v2

PR #82, opened 2019-05-07
findIntentsForContext -> findIntentsByContext

spcfran:
Tiny typo fix.

Also found in FDC3 spec, PR created there as well: https://github.com/FDC3/FDC3/pull/92. Probably best to wait until that one is approved so we are 100% sure that the fix is right


tomrobertsOF (approved):
Good catch.

pjbroadbent:
Note: Edited PR to point at develop, not master.

pjbroadbent (approved):

PR #86, opened 2019-05-16
[SERVICE-480] Broadcasting context doesn't trigger listeners in the sender window, but it does on other windows on the same app

spcfran:


src/provider/controller/ContextHandler.ts
```diff
@@ -39,7 +39,13 @@ export class ContextHandler {
 
         this._channelModel.setContext(channel.id, context);
 
-        return Promise.all(channelMembers.map(identity => this.send(identity, context))).then(() => {});
+        const sourceId = AppWindow.getId(source);
+
+        return Promise.all(channelMembers
```
tomrobertsOF: I know it was like this before, but instead of 
```
return Promise.all(...).then(() => {})
```
could we just do 
```
await Promise.all(...)
```

tomrobertsOF (approved):

src/provider/controller/ContextHandler.ts
spcfran: Sure thing, I also find it clearer that way. Will implement all feedback in one go if that's ok

tomrobertsOF: Yea, no problem. Hardly a blocker, just something I noticed in passing.

test/demo/contextListener.test.ts
```diff
@@ -127,4 +129,53 @@ describe('Context listeners and broadcasting', () => {
...
+            await fdc3Remote.broadcast(testAppMainWindowIdentity, validContext);
+
+            // Window broadcasting the context does NOT receive it
+            const testAppMainWindowReceivedContexts = await testAppMainWindowListener.getReceivedContexts();
+            expect(testAppMainWindowReceivedContexts.length).toBe(0);
+
+            // Window on the same app as the one broadcasting DOES receive context
+            const testAppChildWindowReceivedContexts = await testAppChildWindowListener.getReceivedContexts();
+            expect(testAppChildWindowReceivedContexts.length).toBe(1);
+            expect(testAppChildWindowReceivedContexts[0]).toEqual(validContext);
```
bryangaleopenfin: Would it be simpler to do this as one check - `expect(testAppChildWindowReceivedContexts).toEqual([validContext])`?

bryangaleopenfin (approved):
Approved with comment

test/demo/contextListener.test.ts
spcfran: Very good point, much clearer now. Applied to all tests in the file

PR #87, opened 2019-05-20
[SERVICE-460] FDC3Errors - take 2

spcfran:
Building upon the infrastructure from take 1, this PR:
- implements a number of FDC3Errors, including timeouts
- tests
- Implements raiseIntent with target
- Some assorted refactoring and documenting


Jenkinsfile
```diff
@@ -19,7 +19,7 @@ pipeline {
                     }
```
bryangaleopenfin: Is this file meant to have changed? Phil unified the Jenkins files across our projects a while ago (though may have been done after the architecture branch was create)

test/runner.js
```diff
@@ -55,6 +55,14 @@ async function serve() {
             res.json(testDirectory);
         });
 
+        // Delay response. Useful to test timeouts
+        // Use as http://localhost:3923/fakeDelay/?t=5000
+        app.get('/fakeDelay', (req, res) => {
+            let ms = req.query.t;
+            ms = parseInt(ms) || 0;
+            setTimeout(()=> res.status(200).send({delay: ms}), ms);
```
bryangaleopenfin: Would format as `() =>` to match elsewhere

test/runner.js
```diff
@@ -55,6 +55,14 @@ async function serve() {
             res.json(testDirectory);
         });
 
+        // Delay response. Useful to test timeouts
+        // Use as http://localhost:3923/fakeDelay/?t=5000
+        app.get('/fakeDelay', (req, res) => {
```
bryangaleopenfin, line 60: Might be too much to ask for this PR, but could this be made a general thing for the server on 3923 for any URL? So like, appending ?fakeDelay=X would delay the return of that URL. This is being used to test a manifest taking a long time to load, but right now we're really testing the manifest taking a long time *and* it coming back empty (if I've understood the code correctly)

src/client/internal.ts
```diff
@@ -120,3 +120,21 @@ export interface ContextPayload {
 export interface IntentListenerPayload {
   intent: string;
 }
+
+//
```
bryangaleopenfin: I wouldn't bother having this comment. I feel like "separate before it grows unmanageable" is implicit with all our files :)

src/client/internal.ts
```diff
@@ -120,3 +120,21 @@ export interface ContextPayload {
...
+ * @returns a tuple with the promise and its resolve/reject handlers
+ */
+export function deferredPromise<T = void>(): [Promise<T>, (value?: T) => void, (reason?: any) => void] {
+    let res: (value?: T) => void;
+    let rej: (reason?: any) => void;
+    const p = new Promise<T>((r, rj) => {
+        res = r;
+        rej = rj;
+    });
+    return [p, res! as (value?: T) => void, rej!];
```
bryangaleopenfin: Is this cast needed?

src/demo/apps/LauncherApp.tsx
```diff
@@ -19,11 +19,20 @@ export function LauncherApp(): React.ReactElement {
...
+        try {
+            await fdc3.open(app.appId);
+            console.log(`Opened app ${app.title}`);
+        } catch (e) {
+            // Stringifying an `Error` omits the message!
+            const error: any = {
+                message: e.message,
+                ...e
+            };
+            console.log(e, error);
```
bryangaleopenfin: Better as console.error?

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
```diff
@@ -42,7 +42,14 @@ export function ContextChannelSelector(props: ContextChannelSelectorProps): Reac
                 }
             })
             .catch(error => {
```
bryangaleopenfin: Worth making consistent with the above file? i.e., catch `e` and copy contents to `error`

src/provider/controller/ContextHandler.ts
```diff
@@ -28,11 +28,23 @@ export class ContextHandler {
...
     // TODO: Remove ability to pass an Identity, standardise on AppWindow
+    /**
+     * Send a context to a specific app. Fire and forget
+     * @param app App to send the context to
+     * @param context Context to be sent
+     */
     public async send(app: AppWindow|Identity, context: Context): Promise<void> {
         const identity: Identity = (app as AppWindow).identity || app;
+
+        // Note the call to `dispatch` is not `await`ed, hence the fire and forget behaviour
```
bryangaleopenfin: Would it be better then if we make `send` non-`async` and the return type `void` instead of `Promise<void>`? Feels misleading for it to be async

src/provider/controller/ContextHandler.ts
```diff
@@ -28,11 +28,23 @@ export class ContextHandler {
...
         this._apiHandler.channel.dispatch(identity, APIToClientTopic.CONTEXT, context);
     }
 
+    /**
+     * Broadcast context to all apps in the same channel as the sender, except the sender itself. Fire and forget
```
bryangaleopenfin, line 46: Similar here, if this is fire and forget, better to make it synchronous?

src/provider/controller/IntentHandler.ts
```diff
@@ -41,6 +46,20 @@ export class IntentHandler {
         }
     }
 
+    private async raiseWithTarget(intent: Intent): Promise<IntentResolution> {
```
bryangaleopenfin: Wondering if this should take in target as a second param, to communicate/require that we do definitely have a target (though on the fence if I'd do that myself)

src/provider/index.ts
```diff
@@ -67,14 +68,18 @@ export class Main {
     private async open(payload: OpenPayload): Promise<void> {
         const appInfo: Application|null = await this._directory.getAppByName(payload.name);
 
-        if (appInfo) {
-            const appWindow = await this._model.findOrCreate(appInfo, FindFilter.WITH_CONTEXT_LISTENER);
+        if (!appInfo) {
+            throw new FDC3Error(OpenError.AppNotFound, `No app in directory with name: ${payload.name}`);
+        }
+
+        // This can throw FDC3Error's if app fails to open or times out
```
bryangaleopenfin: 'This can throw an FDC3Error' instead? At least, I don't think the apostrophe should be here

src/provider/index.ts
```diff
@@ -136,8 +142,9 @@ export class Main {
         const appWindow = this._model.getWindow(identity);
         if (appWindow) {
             appWindow.addIntentListener(payload.intent);
-            return Promise.resolve();
+            return;
         } else {
+            // TODO? [SERVICE-429] Should this be an FDC3Error? Or is the AppWindow model an implementation detail and is OK to throw a non-FDC3 error?
```
bryangaleopenfin: Would put 'TODO:' rather than 'TODO?' (Even if we're uncertain if something needs to be done, answering the question and getting rid of the TODO is a definite TODO :p)

src/provider/index.ts
```diff
@@ -146,9 +153,10 @@ export class Main {
         const appWindow = this._model.getWindow(identity);
         if (appWindow) {
             appWindow.removeIntentListener(payload.intent);
-            return Promise.resolve();
+            return;
         } else {
-            throw new Error('App not found in model');
+            // Previously an Error was thrown at this point. Actually, if for some odd reason the window is not in the model,
```
bryangaleopenfin: I wouldn't worry about documenting old behaviour in comments, so would get rid of the first sentence.

src/provider/index.ts
```diff
@@ -146,9 +153,10 @@ export class Main {
...
+            // it's still OK to return successfully, as the caller intention was to remove a listener and the listener is certainly not there.
```
bryangaleopenfin: "caller's"

test/demo/contextChannels.test.ts
```diff
@@ -246,10 +248,55 @@ describe('When joining a channel', () => {
...
+        // Now join green, and we should a callback because of the current state of the green channel.
+        await fdc3Remote.joinChannel(receiveWindow, 'green');
+        receivedContexts = await receiveWindowListener.getReceivedContexts();
+        expect(receivedContexts).toEqual([testContext]);
+
+        // Move all the windows to a different channel, so no one is listening to green. This should cause
+        // the state data to be dropped
+        await fdc3Remote.joinChannel(sendWindow, 'yellow');
+        await fdc3Remote.joinChannel(receiveWindow, 'yellow');
+        const members = await fdc3Remote.getChannelMembers(receiveWindow, 'green');
```
bryangaleopenfin: Would do `await fdc3Remote.getChannelMembers(testManagerIdentity, 'green')` here, as there's no need to do this on a specific window, and the choice of `receiveWindow` over `sendWindow` is arbitrary

test/demo/contextChannels.test.ts
```diff
@@ -246,10 +248,55 @@ describe('When joining a channel', () => {
...
+        expect(members).toEqual([]);
+
+        // Now, subscribe to the green channel again. Because the state data is dropped, we won't get a callback,
+        // and our received contexts will be unchanged (remember, receivedContexts is the life history of all
```
bryangaleopenfin: Phrasing feels odd here maybe "lifetime history" or "receivedContexts is all contexts received over receiveWindow's lifetime"

test/demo/raiseIntent.test.ts
```diff
@@ -76,35 +75,42 @@ describe('Intent listeners and raising intents', () => {
...
 
-            describe('When the target is *not* registered to accept the raised intent', () => {
+            describe('When the target is not in the directory', () => {
                 test('When calling raiseIntent the promise rejects with an FDC3Error', async () => {
-                    const resultPromise = fdc3Remote.raiseIntent(testManagerIdentity, invalidPayload.intent, invalidPayload.context, testAppIdentity.appId);
-                    // TODO: decide if this is this the error message we want when sending a targeted intent
-                    await expect(resultPromise).rejects.toThrowError(/No applications available to handle this intent/);
-                    await expect(resultPromise).rejects.toHaveProperty('name', 'FDC3Error');
-                    await expect(resultPromise).rejects.toHaveProperty('code', ResolveError.NoAppsFound);
+                    // TODO? Does this test make sense here? 'When the target is running' but then 'When the app doesn't exist'? [SERVICE-479 will change this]
```
bryangaleopenfin: Would change to `TODO:`. Or if I'm understanding the test and SERVICE-479 correctly, it would make sense to move this and get rid of the TODO

test/demo/contextChannels.test.ts
```diff
@@ -246,10 +248,55 @@ describe('When joining a channel', () => {
...
+    });
+
+    describe('When the channel to join does not exist', () => {
+        test('An FDC3Error is thrown', async () => {
+            const [window] = await setupWindows(undefined);
+
+            // Join a non-existent channel
+            const joinChannelPromise = fdc3Remote.joinChannel(window, 'non-existent-channel');
+
+            await expect(joinChannelPromise).toThrowFDC3Error(
```
bryangaleopenfin: Cool!

bryangaleopenfin (commented):
Quite a few comments, but mostly minor things, and nothing I feel strongly about if you want to push back on anything. Really like `deferredPromise` and `withTimeout`, they're the elegant version of things I've done before in a bodgy one-use way

Jenkinsfile
spcfran: I think I messed up a little bit while doing some git upkeep 😅 and merged something into my branch that maybe I shouldn't have. But I believe it's only this file. I can revert manually although not sure if that will cause more conflicts when merging all this into develop?

src/client/internal.ts
spcfran: Apparently not! Looks like a leftover from messing around a bit with types here.

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
spcfran: Good catch! Also went from `.then().catch()` to async `try{} catch{}` while at it, for consistency

src/provider/controller/IntentHandler.ts
spcfran: I agree with your point although it feels a bit strange to pass the target twice (in the object + explicitly as string). I've proposed a different solution based on typescript guards, please let me know your thoughts

src/provider/index.ts
spcfran: My intention was to make clear that different errors (with different codes) can be thrown here depending on the 'internal' circumstances - I'm going with removing apostrophe at the moment unless someone else has a strong opinion

src/provider/index.ts
spcfran: Agree all TODOs should be treated with equal importance (and addressed sooner rather than later!!). I tend to use `?` and `!` as I'm using [this extension](https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments) which colours them differently. And actually normal `// TODO:`s are coloured more prominently than `?` so should probably use that from now on :)

test/demo/raiseIntent.test.ts
spcfran: Yes 479 is going to affect these tests quite a bit - Changing to `TODO:` by now

test/runner.js
spcfran: Sounds like a great idea! And useful for other projects I guess. 

Would it make sense for this to be part of the common service tooling?

spcfran:
> Quite a few comments, but mostly minor things, and nothing I feel strongly about if you want to push back on anything. Really like `deferredPromise` and `withTimeout`, they're the elegant version of things I've done before in a bodgy one-use way

Thanks for the super quick response time and the comprehensive feedback!

Jenkinsfile
bryangaleopenfin: Yeah, I think revert manually so there's no change to `Jenkinsfile` in this PR

test/runner.js
bryangaleopenfin: Maybe? Though also, flagging that this will clash with Ryan's PR that eliminates runner.js

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Cool, looks good

bryangaleopenfin (commented):
Looks good. Happy to approve once `Jenkinsfile` is sorted

test/runner.js
spcfran: @ryanpetersonOF FYC - may be useful to have at service-tooling level some middleware that checks for that query string parameter and delays the pipeline (but ends up processing the actual request, unlike my quick and dirty solution here), to test timeout scenarios

bryangaleopenfin (approved):

Jenkinsfile
spcfran: 👌 done

test/runner.js
rypete: We could definitely add this as default middleware to the tooling.  Alternatively, the tooling also lets you import custom middleware so either option will work here.  See: https://github.com/HadoukenIO/fdc3-service/pull/88/files#diff-b6a39a12832df47bd21f2dbcc919998d for working example of this.

tomrobertsOF (approved):

src/demo/apps/LauncherApp.tsx
```diff
@@ -16,14 +16,23 @@ export function LauncherApp(): React.ReactElement {
...
+            console.log(`Opened app ${app.title}`);
+        } catch (e) {
+            // Stringifying an `Error` omits the message!
+            const error: any = {
+                message: e.message,
+                ...e
+            };
+            console.error(e, error);
+            alert(`openApp threw an error!\n${JSON.stringify(error)}`);
+        }
```
tim-dinsdale, line 34: an alert is pretty drastic. If we are going to do that, we need to be a lot clearer - it will need a lot of diagnostics. I think I'd rather keep it in the console.

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
```diff
@@ -30,20 +30,25 @@ export function ContextChannelSelector(props: ContextChannelSelectorProps): Reac
...
+            }
+        } catch (e) {
+            // Stringifying an `Error` omits the message!
+            const error: any = {
+                message: e.message,
+                ...e
+            };
+            console.error(`Unable to join channel ${id}! ${e.message}`);
+            alert(`joinChannel threw an error!\n${JSON.stringify(error)}`);
+        }
```
tim-dinsdale, line 50: as above about alerts.

src/provider/index.ts
```diff
@@ -146,9 +153,10 @@ export class Main {
         const appWindow = this._model.getWindow(identity);
         if (appWindow) {
             appWindow.removeIntentListener(payload.intent);
-            return Promise.resolve();
+            return;
```
tim-dinsdale: Not for this one, but do you actually need this return statement?

src/demo/apps/LauncherApp.tsx
spcfran: it was meant as a quick and dirty way to test the FDC3Errors in the UI :) - and for anyone testing without the console open. 

Happy to remove by now. Maybe we should add something to the backlog to add some nice error handling?

tim-dinsdale: Yes, that's a good plan.

tim-dinsdale (approved):

src/demo/components/common/IntentButton.tsx
```diff
@@ -28,7 +28,8 @@ export function IntentButton(props: IntentButtonProps): React.ReactElement {
             })
             .catch(err => {
                 setButtonState(ButtonStateType.ERROR);
-                alert(`Intent failed with message '${err.message}'`);
+                // Alert commented out as per https://github.com/HadoukenIO/fdc3-service/pull/87/files/d637737b8e74e858d8cc978f1fa9755e3ad6c308#r288216259
+                // alert(`Intent failed with message '${err.message}'`);
```
bryangaleopenfin: Why is this commented out rather than just deleted? The URL doesn't go anywhere anymore

src/client/errors.ts
```diff
@@ -55,3 +65,29 @@ export class FDC3Error extends Error {
...
+ */
+export const Timeouts = {
+    /**
+     * Time for an app to register a listener after opening
+     */
+    ADD_INTENT_LISTENER: 5000,
+    /**
+     * Time for an OpenFin app to start by calling `fin.Application.startFromManifest`
+     */
+    APP_START_FROM_MANIFEST: 5000
```
bryangaleopenfin: We're seeing this take 20+ seconds on Jenkins

src/provider/model/AppWindow.ts
```diff
@@ -1,6 +1,8 @@
 import {Window, Identity} from 'openfin/_v2/main';
 
 import {Application, IntentType} from '../../client/main';
+import {FDC3Error, OpenError, withTimeout, Timeouts} from '../../client/errors';
+import {deferredPromise} from '../../client/internal';
```
bryangaleopenfin: These new imports aren't needed now

src/demo/components/common/IntentButton.tsx
spcfran: the link was meant to be to [this conversation](https://github.com/HadoukenIO/fdc3-service/pull/87#discussion_r288216259). I left it (and also other commented out `alert` calls) as a reminder that something needs to be done at some point / add a story for error handling

bryangaleopenfin:
Cool. Latest changes look good. Should be good to go in one that last test failure is sorted

PR #91, opened 2019-05-24
[SERVICE-478] Complete base context broadcasting functionality (Unit testing)

bryangaleopenfin:
- Separates `AppWindow` into an interface and its `FinAppWindow` implementation. I've put it in `FinEnvironment` so that it doesn't need to be exported. Feels like it's an implementation detail of `FinEnvironment`, so shouldn't be visible elsewhere

- Adds unit tests around `ContextHandler`

- Fixes some circular dependencies, and on-import references to `fin` that were causing problems when running things in Jest

- Still need to do a review to see if this fully completes the story, but think this is worth getting in as a chunk on its own

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -48,24 +48,24 @@ if (typeof fin !== 'undefined') {
 
 export function getServicePromise(): Promise<ChannelClient> {
     if (!channelPromise) {
-        // Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
-        // that includes this, but for now it is easier to put a guard in place.
-        if (fin.Window.me.uuid !== SERVICE_IDENTITY.uuid || fin.Window.me.name !== SERVICE_IDENTITY.name) {
-            channelPromise = typeof fin === 'undefined' ?
-                Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.')) :
-                fin.InterApplicationBus.Channel.connect(SERVICE_CHANNEL, {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
-                    // Register service listeners
-                    channel.register('WARN', (payload: any) => console.warn(payload));  // tslint:disable-line:no-any
-                    channel.register('event', (event: FDC3Event) => {
-                        eventEmitter.emit(event.type, event);
-                    });
-                    // Any unregistered action will simply return false
-                    channel.setDefaultAction(() => false);
-
-                    return channel;
-                });
-        } else {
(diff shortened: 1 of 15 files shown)


src/client/connection.ts
```diff
@@ -66,7 +79,7 @@ if (fin.Window.me.uuid !== SERVICE_IDENTITY.uuid || fin.Window.me.name !== SERVI
  * @param payload Object containing additional arguments
  */
 export async function tryServiceDispatch<T extends APIFromClientTopic>(action: T, payload: APIFromClient[T][0]): Promise<APIFromClient[T][1]> {
-    const channel: ChannelClient = await channelPromise;
+    const channel: ChannelClient = (await channelPromise)!;
```
tomrobertsOF: Would it not be better for this to be `await getServicePromise`? It is probably impossible for it to actually be null, but I'd rather just use the util we have instead of asserting non-null

bryangaleopenfin: Ah, yep, leftover change, fixed

tomrobertsOF: Cool!

src/client/connection.ts
```diff
@@ -40,23 +40,36 @@ export const eventEmitter = new EventEmitter();
...
+if (typeof fin !== 'undefined') {
+    getServicePromise();
+}
+
+export function getServicePromise(): Promise<ChannelClient> {
+    if (!channelPromise) {
+        // Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
+        // that includes this, but for now it is easier to put a guard in place.
+        if (fin.Window.me.uuid !== SERVICE_IDENTITY.uuid || fin.Window.me.name !== SERVICE_IDENTITY.name) {
+            channelPromise = typeof fin === 'undefined' ?
```
spcfran: I spotted this a while ago but now that it's being rewritten could be a good time to fix.

This `undefined` check after accessing `fin.Window....` doesn't make sense. If we want to throw the custom error it should happen right before (and maybe make redundant the extra `if` on line 45?)

tim-dinsdale (commented):
LGTM. I don't like the global getId function, but I appreciate that is hard to do in any other way given it's on a class we don't own; and I am not a fan of injection but that's just personal preference.

tomrobertsOF (approved):

src/client/connection.ts
bryangaleopenfin: Cool, fixed. We can't get rid of the earlier `'undefined'' check, as we do want to do that early getServicePromise() call, but _only_ within the OpenFin environment

tim-dinsdale (approved):

PR #93, opened 2019-05-24
[SERVICE-478] Complete base context broadcasting functionality (Integration tests)

bryangaleopenfin:
This fills in the tests marked as 'todo' in contextListener.test.ts

More detailed tests covering "app windows not receiving their own broadcast" were already present in the "Broadcasting with multiple windows in the same app" `describe` block


tim-dinsdale (approved):
This looks like it's just comment changes now, so approved

tomrobertsOF (approved):
Something weird seems to have happened with previous merges, but I guess these three comments are worth having :D

PR #94, opened 2019-05-28
[SERVICE-428] Finalise context channel API's and lay groundwork

pjbroadbent:


src/client/connection.ts
```diff
@@ -49,7 +48,13 @@ if (fin.Window.me.uuid !== SERVICE_IDENTITY.uuid || fin.Window.me.name !== SERVI
         fin.InterApplicationBus.Channel.connect(SERVICE_CHANNEL, {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
             // Register service listeners
             channel.register('WARN', (payload: any) => console.warn(payload));  // tslint:disable-line:no-any
-            channel.register('event', (event: FDC3Event) => {
+            channel.register('event', async (event: FDC3Event|ChannelEvent) => {
+                // Special-handling for some event types, to convert transport-type event to client-side event.
+                if (event.type === 'channel-changed') {
+                    event.channel = event.channel ? await getChannelById(event.channel.id) : null;
+                    event.previousChannel = event.previousChannel ? await getChannelById(event.previousChannel.id) : null;
```
bryangaleopenfin, line 55: Where this is now async, is there a danger of the client receiving events out of order?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+}
+
+/**
+ * Event fired when a window is removed from a channel. See {@link Channel.addEventListener}.
+ *
+ * Note that this event will typically fire as part of a pair - since windows must always belong to a channel, a window
+ * can only join a channel by leaving it's previous channel. The exceptions to this rule are when the window is created
+ * and destroyed when there will be no previous channel or no current channel, respectively.
+ *
+ * To listen for channel changes across all (or multiple) channel, there is also a top-level {@link ChannelChangedEvent}.
```
bryangaleopenfin, line 58: Add this note to `ChannelWindowAddedEvent` too?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
     /**
-     * Would be generated by the service - what you pass to the API calls below
+     * The window that has just been added to the channel.
      */
-    id: ChannelId;
+    identity: Identity;
 
     /**
-     * Uniquely defines each channel type
+     * The channel that has been modified. Will always be the channel object that {@link Channel.addEventListener} was
```
bryangaleopenfin, line 38: Don't like this phrasing - both channels have been modified. Would make it symmetrical with the below - `The channel that the window belongs to now`

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+     * The channel that the window now belongs to.
+     *
+     * Will be `null` if the window is leaving the channel due to it being closed.
      */
-    name: string;
+    channel: Channel|null;
 
     /**
-     * The colour that should be associated within this channel when displaying this channel in a UI
+     * The channel that has been modified. Will always be the channel object that {@link Channel.addEventListener} was
```
bryangaleopenfin, line 78: Similar comment to above. Both channels have been modified

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
-    return tryServiceDispatch(APIFromClientTopic.GET_ALL_CHANNELS, {});
+export abstract class ChannelBase {
+    /**
+     * Constant that uniquely identifies this channel. Will be generated by the service, and guarenteed to be unique
+     * within the set of channels registered with the service.
+     *
+     * In the case of `desktop` channels (see {@link DesktopChannel}), these id's _should_ persist across sessions. The
+     * channel list is defined by the service, but can be overridden by a desktop owner. If the desktop owner keeps
+     * this list static (which is recommended), then id's will also persist across sessions.
+     */
```
bryangaleopenfin, line 137: IDs rather than id's?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+        this.eventEmitter = new EventEmitter();
+        this.listeners = [];
+    }
+
+    /**
+     * Returns a list of all windows belonging to the specified channel.
+     *
+     * If the window making the call is a member of this channel, it will be included in the results.
+     *
+     * This will only include windows that contain context listeners. If there are no windows on this channel, an empty
```
bryangaleopenfin, line 163: "This will only include windows that contain context listeners." - no longer correct?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+
+    /**
+     * Adds the given window to this channel. If no identity is provided, the window making the call will be the window
+     * added to the channel.
+     *
+     * If the channel has a current context (see {@link getCurrentContext}) then that context will be immediately passed to
+     * the given window upon joining the channel, via its context listener(s).
+     *
+     * Note that all windows will always belong to exactly one channel at all times. If you wish to leave a channel, the
+     * only way to do so is to re-join the default channel. This can be done by calling `channels.defaultChannel.join()`.
```
bryangaleopenfin, line 210: This isn't quite correct - "If you wish to leave a channel, the only way to do so is to re-join the default channel". Change to "If you wish to leave a channel, the only way to do so is to join another channel. A window may rejoin the default channel by calling..."

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+     * Event that is fired whenever a window changes leaves this channel. This includes switching to/from the default
+     * channel.
+     *
+     * The event also includes which channel the window is being added to. The `previousChannel` property within the
+     * event will always be this channel instance.
+     */
+    public addEventListener(eventType: 'window-removed', listener: (event: ChannelWindowRemovedEvent) => void): void;
+
+    public addEventListener<E extends ChannelEvent>(eventType: E['type'], listener: (event: E) => void): void {
+        if (typeof fin === 'undefined') {
```
bryangaleopenfin, line 262: Do we want this check in `addBroadcastListener`/`addContextListener`

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+     *
+     * @param eventType The event being unsubscribed from
+     * @param listener The callback function to remove
+     */
+    public removeEventListener<E extends ChannelEvent>(eventType: E['type'], listener: (event: E) => void): void {
+        if (typeof fin === 'undefined') {
+            throw new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.');
+        }
+
+        this.eventEmitter.addListener(eventType, listener);
```
bryangaleopenfin, line 285: `removeListener`

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+    }
 }
 
 /**
- * Sets the window to the channel with the given identifier
+ * User-facing channels, to display within a colour picker or channel selector component.
```
bryangaleopenfin, line 290: Comment will be in docs, so use American "color"?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+ *
+ * This list of channels should be considered fixed by applications - the service will own the list of user channels,
+ * making the same list of channels available to all applications, and this list will not change over the lifecycle of
+ * the service.
  *
- * Use the special constant 'global' to revert to the global channel.
+ * We do not intend to support creation of 'user' channels at runtime, as this then adds considerable complexity when
+ * implementing a channel selector component (must now support events, 'create channel' UI, reflowing of channel
+ * list, etc).
```
bryangaleopenfin, line 298: Phrasing is oddly casual for the docs, and doesn't need this much detail? Would re-write as

" We do not intend to support creation of 'user' channels at runtime, as this would add considerable complexity when implementing a channel selector component as it would need to support a dynamic channel list"

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+    public constructor() {
+        super(DEFAULT_CHANNEL_ID, 'default');
+    }
 }
 
 /**
- * Returns a list of all windows belonging to the specified channel.
+ * The channel in which all windows will initially be placed.
  *
+ * All windows that have a context listener will always belong to exactly one channel at all times. If they have not
```
bryangaleopenfin, line 338: Drop "that have a context listener"

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+ * explicitly been placed into a channel via a {@link Channel.join} call, they will be in this channel.
+ *
+ * If an app wishes to leave a named channel, it can do so by (re-)joining this channel.
```
bryangaleopenfin, line 341: Use "Desktop Channel" rather than "named channel"?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+
+/**
+ * Fetches a channel object for a given channel identifier. The `channelId` property maps to the {@link Channel.id} field.
+ *
+ * TODO (SERVICE-429): Decide what happens if calling with an invalid channel ID. Will the client know it's
+ * invalid?.. Also, what if desktop channel list hasn't yet been fetched from provider? (may need to be made async).
+ * @param channelId The ID of the channel to fetch.
+ */
+export async function getChannelById(channelId: ChannelId): Promise<Channel|null> {
+    return channelLookup[channelId] || null;
```
bryangaleopenfin, line 353: Intentional that this is never populated?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+function getChannelObject<T extends Channel = Channel>(data: ChannelTransport): T {
+    let channel: Channel = channelLookup[data.id];
+
+    if (!channel) {
+        switch (data.type) {
+            case 'default':
+                channel = defaultChannel;
+                break;
+            case 'desktop':
+                channel = new DesktopChannel(data as DesktopChannelTransport);
```
bryangaleopenfin, line 420: Insert into channelLookup?

src/client/contextChannels.ts
```diff
@@ -2,79 +2,424 @@
...
+ *
+ * @param eventType The event being unsubscribed from
+ * @param listener The callback function to remove
+ */
+export function removeEventListener<E extends ChannelEvent>(eventType: E['type'], handler: (eventPayload: E) => void): void {
+    eventEmitter.removeListener(eventType, handler);
+}
+
+const channelLookup: {[id: string]: Channel} = {
+    [DEFAULT_CHANNEL_ID]: defaultChannel
```
bryangaleopenfin, line 409: Why use this rather than `Map`? (Meant as a genuine, typescript-junior, non-leading question!)

src/provider/ChannelModel.ts
```diff
@@ -1,80 +1,79 @@
 import {Identity} from 'openfin/_v2/main';
```
bryangaleopenfin, line 1: Do we want to be using `Transport` objects at this level? They feel more like things to use as close to the point we're sending stuff across the wire as possible

src/provider/controller/ContextHandler.ts
```diff
@@ -47,32 +51,45 @@ export class ContextHandler {
...
 
-    public async getChannel(payload: GetChannelPayload, source: ProviderIdentity): Promise<Channel> {
-        const identity = payload.identity || source;
+    public async channelBroadcast(payload: ChannelBroadcastPayload, source: ProviderIdentity): Promise<void> {
+        const channel = this._channelModel.getChannelForWindow(source);
+        const channelMembers = this._channelModel.getChannelMembers(channel.id);
+
+        this._channelModel.setContext(channel.id, payload.context);
+
+        return Promise.all(channelMembers.map(identity => this.send(identity, payload.context))).then(() => {});
```
bryangaleopenfin, line 70: Need to filter out requesting window?

src/client/contextChannels.ts
pjbroadbent: This is pointing out that if you have a channel "`c`", and you do `c.addEventListener('window-added', ...)`, the `channel` property of all the events you receive will always be `c` (as in "triple-equal to" `c`, not just an object with the same channel ID).

It's not saying that only `channel` has been modified, just that `channel` will be the channel you're listening to. The same is true of `window-removed`, only with `previousChannel` rather than `channel`.

Perhaps this is unnecessary, or more misleading than helpful. Wording is clunky, but that was the best I could come up with - can discuss.

src/client/contextChannels.ts
pjbroadbent: Yep, will update.

src/client/contextChannels.ts
pjbroadbent: From an earlier draft.. :-) Will update.

src/client/contextChannels.ts
pjbroadbent: It's (indirectly) populated as the result of a `getDesktopChannels` call - but function will not work as expected until after that function is called.

The TODO is to determine how best to handle this.

edit: actually, it's not... Intention was for `getChannelObject` to populate the map, as you commented below.

src/client/contextChannels.ts
pjbroadbent: Pretty arbitrary decision really. Mainly to keep things "lightweight", since we only need a string key. Will also never be removing or filtering, or anything like that. The `Map` type is mostly useful when you want keys that are types other than string.

src/provider/ChannelModel.ts
pjbroadbent: A little bit of a trade-off here. Doing the transport conversion lower-down in the chain may require exposing a bit more of the internals of the model - things that could otherwise be encapsulated.

Need to decide if we want this class being the `APIHandler` endpoint for those calls, or if we want something more at the `index`/`APIHandler`-level to do the conversion.

The implementation story would be a good time to decide this, as I imagine a lot of this will be changing anyway.

src/client/contextChannels.ts
bryangaleopenfin: To be clear, my issue is with the first sentence, not the second

bryangaleopenfin (commented):
Couple more things I want to add!

- Just twigged `window-added` and `window-removed` are dispatched from anywhere, but guess that's why this is a draft. Had wanted to check if we're being intentional about the order these two and `channel-changed` are being fired

- Is there a discussion to be had about if the client's model of what window is where is updated by push or by pull? Like, `getMembers` could be an explicit call to the provider each time, or it could be that the client only calls it once, then updates it's internal model from the events it received. (And is there any danger of event ordering/provider calls getting misordered and giving misleading results).

PR #95, opened 2019-05-28
[N/A] Fix test timeouts on Jenkins

bryangaleopenfin:
Apps will generally take 3-5 seconds to startup on Jenkins (locally they take 1-2 seconds, and the Jest default time out is 5 seconds). Sometimes on Jenkins, apps will take 20+ seconds to start.

This changes the test timeouts to accommodate this, and makes the necessary changes to raiseIntent.test given we can't guarantee a 'quick' app startup time

Aware this will conflict when we pull in develop, given unification of service tests in develop


pjbroadbent (approved):

tim-dinsdale (approved):
The last constant is sad, but as Bryan explained to me, that's in a js script, and the const is defined in a ts script, so

PR #98, opened 2019-05-29
[SERVICE-479] integrate with non directory apps

spcfran:
Ready for review. 

Still not 100% test coverage


src/provider/index.ts
```diff
@@ -139,14 +147,33 @@ export class Main {
...
-        const appWindow = this._model.getWindow(identity);
-        if (appWindow) {
-            appWindow.addIntentListener(payload.intent);
-            return;
-        } else {
-            // TODO: [SERVICE-429] Should this be an FDC3Error? Or is the AppWindow model an implementation detail and is OK to throw a non-FDC3 error?
-            throw new Error('App not found in model');
+        let appWindow = this._model.getWindow(identity);
+
+        // Window is not in model - this should mean that the app is not in the directory, as directory apps are immediately added to model upon window creation
```
pjbroadbent, line 152: Another edge case - app exists in directory and/or is running, but hasn't (yet?) imported the client module.

Very edge-case-y though, not a concern for the time being.

src/provider/MetadataStore.ts
```diff
@@ -1,4 +1,4 @@
-import {Application} from 'openfin/_v2/main';
+import {Application as OpenFinApplication} from 'openfin/_v2/main';
```
bryangaleopenfin, line 1: Cool, makes sense!

src/provider/controller/SelectorHandler.ts
```diff
@@ -71,6 +72,9 @@ export class SelectorHandler extends AsyncInit {
     @inject(Inject.APP_DIRECTORY)
     private _directory!: AppDirectory;
 
+    @inject(Inject.MODEL)
```
bryangaleopenfin, line 75: Suggest move this class to using constructor injection, rather than adding more parameter injection. At the moment it's a bit of a mix, and I'd like it if we could move towards constructor, which I think is what `injectify` recommend

src/provider/index.ts
```diff
@@ -83,11 +83,19 @@ export class Main {
...
+     * @param payload Contains the intent type to find information for
+     */
     private async findIntent(payload: FindIntentPayload): Promise<AppIntent> {
         let apps: Application[];
         if (payload.intent) {
-            apps = await this._directory.getAppsByIntent(payload.intent);
+            apps = await this._model.getApplicationsForIntent(payload.intent);
         } else {
+            // This is a non-FDC3 workaround to get all directory apps by calling `findIntent` with a falsy intent.
+            // Ideally the FDC3 spec would expose an API to access the directory in a more meaningful way
```
bryangaleopenfin, line 98: 👍

src/provider/controller/IntentHandler.ts
```diff
@@ -47,16 +47,35 @@ export class IntentHandler {
...
+        const appWindow = this._model.findWindowByAppName(intent.target);
+
+        if (appWindow) {
+            // Target app is running -> ensure that intent is registered in model
+            if (!appWindow.hasIntentListener(intent.type)) {
+                throw new FDC3Error(
+                    ResolveError.TargetAppDoesNotHandleIntent,
+                    `App '${intent.target}' has not registered listeners for intent '${intent.type}'`
+                );
+            }
```
bryangaleopenfin: Are we worried about the case where an App has multiple windows, and the listeners are on a different window to the one returned by findWindowByAppName?

src/provider/controller/IntentHandler.ts
```diff
@@ -47,16 +47,35 @@ export class IntentHandler {
...
 
-        if (!(appInfo.intents || []).some(appIntent => appIntent.name === intent.type)) {
-            throw new FDC3Error(ResolveError.TargetAppDoesNotHandleIntent, `App '${intent.target}' does not handle intent '${intent.type}'`);
+            appInfo = appWindow.appInfo;
+        } else {
+            // Target app not running -> Try to find in directory
+            appInfo = await this._directory.getAppByName(intent.target);
+            if (!appInfo) {
+                // TODO: Should this message also mention 'nor running' to cover ad-hoc non-directory apps? [SERVICE-479]
```
bryangaleopenfin: Resolve before merging

src/provider/controller/IntentHandler.ts
```diff
@@ -47,16 +47,35 @@ export class IntentHandler {
...
+        let appInfo: Application|null;
 
-        if (!appInfo) {
-            throw new FDC3Error(ResolveError.TargetAppNotInDirectory, `No app in directory with name: ${intent.target}`);
-        }
+        const appWindow = this._model.findWindowByAppName(intent.target);
+
+        if (appWindow) {
+            // Target app is running -> ensure that intent is registered in model
+            if (!appWindow.hasIntentListener(intent.type)) {
```
bryangaleopenfin: Feels odd to do this check here, but in the same method to trust that non-running apps will register listeners. I feel like we should leave this to be caught by `ensureReadyToReceiveIntent`. And what if raiseWithTarget is called during the 5 seconds we allow for the intent listener to be added?

src/provider/model/Model.ts
```diff
@@ -108,31 +131,59 @@ export class Model {
         }
     }
 
+    /**
+     * Gets apps that can handle an intent
+     *
+     * Includes windows that are not in the app directory but have registered a listener for it [SERVICE-479]
```
bryangaleopenfin: Suggest don't worry about ticket references other than for 'TODO's

bryangaleopenfin (commented):
Review of implementation changes - will look at tests separately

src/provider/model/Model.ts
```diff
@@ -108,31 +131,59 @@ export class Model {
...
+     * @param intentType intent type
+     */
+    public async getApplicationsForIntent(intentType: string): Promise<Application[]> {
```
bryangaleopenfin, line 113: Suggest keep public methods above private methods

bryangaleopenfin (commented):
Tests look good. Would maybe add a line of whitespace between tests for readability, and might be too much for this PR, but I wonder if there's any way to reduce code duplication (the tests where we have two listeners and try different combinations of unsubscribing them jump out for me and something we'd want to de-duplicate)

src/provider/controller/IntentHandler.ts
spcfran: Yep @pjbroadbent and I have discussed about the apps-with-multiple-windows thing quite a bit, we decided to leave it for a future story, once we know a bit more about how the resolver deals with individual windows, and whether the concept of "target" may include window information or just app

src/provider/controller/IntentHandler.ts
spcfran: @pjbroadbent thoughts?

src/provider/controller/IntentHandler.ts
spcfran: That's an excellent point. Will delegate that check to `fireIntent` and eventually throw the `Timeout waiting for intent listener to be added` if appropriate.

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Cool cool. @pjbroadbent - just want to confirm this is as captured in Jira as it needs to be before I hit 'resolve conversation'

src/provider/controller/IntentHandler.ts
pjbroadbent: Let's discuss, come up with some wording that makes sense.

src/provider/controller/SelectorHandler.ts
spcfran: ✔ Noted as a nice to have tidy-up task

src/provider/index.ts
spcfran: Mmhh not sure I follow. The app/window calling `addIntentListener` surely has imported the client?

Or do you mean the app/window that is the target of `raiseIntent`? In which case I think the "wait for intent handshake" in `fireIntent` would cover this?

tomrobertsOF (approved):
All fine by me

src/provider/controller/IntentHandler.ts
pjbroadbent: I've added a story to cover the potential upcoming changes that we discussed (SERVICE-505), but wrote it in a "TBD if we do this or not" sense.

May be worth for the time being changing the above check to fetch all windows from the model and search for a matching window here(?).

pjbroadbent: Scratch that. Was thinking the issue was specific to "intents with target", it's a bit more generic than that - and could do with a small story rather than building into this PR.

Have added SERVICE-506 to cover the change.

src/provider/index.ts
pjbroadbent: Sorry, my bad. Was assuming that `identity` could be an app that isn't using the FDC3 client - but since this method only runs if the service receives a message _from_ identity, then that window must have the client imported.

bryangaleopenfin (approved):

pjbroadbent (approved):

PR #106, opened 2019-06-04
[SERVICE-507] Clients can join and query membership of channels using finalized channels API

bryangaleopenfin:
**Notes**

- The client is not optimized to minimize calls to the provider. I like how simple it is right now, and that `getChannelById` and `getDesktopChannels` are symmetrical, but open to other opinions on whether this is worth optimizing

- I think `contextChannels.inttest` covers everything, but the tests definitely aren't what they would be if they'd been written against the new API from the outset. I suspect I'll get a clearer picture of a better organisation of the tests when the `Channel` object gains more functionality in SERVICE-508

- `ChannelModel` is using `ChannelTransport` now, as `Channel` is no longer a flat, data-only type. This feels messy (or at least, we're leaning on `ChannelTransport` to be both the thing for sending channels across the wire, and our stand-in for a dedicated provider-side Channel class), but `ChannelModel` will likely go away in SERVICE-508


pjbroadbent:
@bryangaleOF Having a `common` folder affects the file structure of the code that gets bundled in the NPM module.

For example, `import {foo} from 'openfin-fdc3/contextChannels'` would become `import {foo} from 'openfin-fdc3/client/contextChannels'`. That felt kinda messy when we were discussing it, which is why we've put common stuff in the `src/client` folder up until now.

src/common/validation.ts
```diff
@@ -0,0 +1,34 @@
...
+        const uuidCheck = typeof identity.uuid === 'string';
+        const nameCheck = !identity.name || typeof identity.name === 'string';
+
+        if (!uuidCheck || !nameCheck) {
+            error = true;
+        }
+    }
+
+    if (error) {
+        throw new FDC3Error(IdentityError.InvalidIdentity, `Invalid identity: ${JSON.stringify(identity)}`);
```
pjbroadbent: Do we want to protect against `stringify` throwing, if a non-serialisable object is passed?

src/common/errors.ts
```diff
@@ -18,10 +18,12 @@ export enum ResolveError {
...
 
 export enum ChannelError {
-    // When getChannel / joinChannel on a non existing channel id
-    ChannelDoesNotExist = 'ChannelDoesNotExist',
-    // When trying to create a channel that already exists
-    ChannelAlreadyExists = 'ChannelAlreadyExists'
+    ChannelDoesNotExist = 'ChannelDoesNotExist'
+}
+
+export enum IdentityError {
```
pjbroadbent: Is this getting a bit too granular? The error messages should be detailed, but the codes should be a little more abstracted, only distinguishing between cases that an app is likely to want to handle programmatically.

Might want to think about how this would "scale" - if we'd want to extend this pattern to things like `InvalidIdentity`/`InvalidUuid`/`InvalidChannelId` and `WindowWithIdentityNotFound`/`ChannelWithIdNotFound`, etc. Seems unnecessarily verbose to me, would apps be likely to handle all of those codes?

src/client/contextChannels.ts
```diff
@@ -4,77 +4,216 @@
...
+export type Channel = DesktopChannel|DefaultChannel;
 
 /**
- * Object representing a fully-initialised channel.
+ * Event fired whenever a window changes channel. See {@link addEventListener}.
+ *
+ * This event can be used to track all channel changes, rather than listening only to a specific channel.
  *
- * Whenever the service returns any kind of channel info, it'll always be with this type.
+ * @event 'channel'
```
pjbroadbent: Notice this was in the earlier API refactor PR, but does the extra `'channel'` affect the output? The other event definitions (including in layouts) just use `@event`.

Can we remove if it's not doing anything useful?

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
```diff
@@ -17,52 +17,60 @@ ContextChannelSelector.defaultProps = {
...
     React.useEffect(() => {
-        getChannel().then(channel => {
-            setColor(channel.color);
-            setCurrentChannelId(channel.id);
+        getCurrentChannel().then(channel => {
+            if (channel.type === 'desktop') {
+                setColor(channel.color);
+            } else {
+                // Use white for default channel
+                setColor(0xFFFFFF);
```
pjbroadbent, line 29: Something worth discussing - is removing these properties for the 'default' channel a good idea? I assumed so, but looking at what it means for this sort of UI code (and the `(channel as DesktopChannel).name || 'Default'` below) makes me think consistency across all service-defined channels might be better?..

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
```diff
@@ -17,52 +17,60 @@ ContextChannelSelector.defaultProps = {
...
+                        if (selectedChannel.type === 'desktop') {
+                            setColor(selectedChannel.color);
+                        }
+                    } else {
+                        setColor(0xFFFFFF);
+                    }
+                })
+                .catch((error: Error) => {
+                    console.error(`Unable to join channel ${id}! ${error.message}`);
+                });
```
pjbroadbent, line 57: Should this be async/await rather than chained?

src/provider/APIHandler.ts
```diff
@@ -51,6 +51,12 @@ export type APIImplementation<T extends Enum, S extends APISpecification<T>> = {
     [K in T]: APIAction<S[K]>;
 };
 
+/**
+ * Accept both async and non-async actions. Optionally accept the readonly version of the given type, given
+ * we trust APIHandler not to alter its contents
```
pjbroadbent: I don't think there's any kind of "trust" involved here. `APIHandler` doesn't need to modify the object it gets back from the provider, so it's completely fine for the action to return immutable data.

The data may get mutated once passed to the client, but then that will always be a clone (well, serialised and then de-serialised) of this value.

src/provider/APIHandler.ts
```diff
@@ -86,11 +92,14 @@ export class APIHandler<T extends Enum> {
...
             if (actionHandlerMap.hasOwnProperty(action)) {
-                this._providerChannel.register(action, (payload, source) =>
-                    actionHandlerMap[action](payload, source)
-                        .catch(error => {
-                            throw FDC3Error.serialize(error);
-                        }));
+                this._providerChannel.register(action, async (payload, source) => {
+                    try {
+                        const result = await actionHandlerMap[action](payload, source);
+                        return result;
```
pjbroadbent, line 100: Can this just be `return actionHandlerMap[action](payload, source);`? Since it's an async method, returning a promise will implicitly await.

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -13,14 +13,46 @@
...
 const ofBrowser = new OFPuppeteerBrowser();
 
+const remoteChannels: {[id: string]: RemoteChannel} = {};
+
+export interface RemoteChannel {
+    channel: Channel;
+    id: string;
+
+    getMembers: () => Promise<Identity[]>;
+    join: (identity?: Identity) => Promise<void>
```
pjbroadbent: `;`

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -283,6 +270,80 @@ export async function findIntentsByContext(executionTarget: Identity, context: C
...
+    ).catch(handlePuppeteerError);
+
+    return deserializeChannel(executionTarget, testChannelTransport);
+}
+
+function deserializeChannel(executionTarget: Identity, transport: TestChannelTransport): RemoteChannel {
+    let remoteChannel = remoteChannels[transport.id];
+
+    if (!remoteChannel) {
+        remoteChannel = {
```
pjbroadbent, line 315: Should `RemoteChannel` be a class? Would be a cleaner way of declaring all these functions.

test/demo/utils/jestExtensions.ts
```diff
@@ -9,6 +12,13 @@ declare global {
              * @param message Optionally, assert that a given error message is returned
              */
             toThrowFDC3Error<R>(code: string, message?: string | RegExp): R;
+
+            /**
+             * Used to test that a RemoteChannel represents the expected client-side Channel
+             * @param channel An object with optional `id` and `type` properties that if present we expect the channel to match
+             * @param classType The class we expect the channel to be an instance of
+             */
+            toBeChannel(channel: {id?: ChannelId, type?: string} | string, classType?: Function): R;
```
pjbroadbent: What's the `| string` part for? For matching just the channel ID?..

src/common/errors.ts
bryangaleopenfin: This came from the `@throws` declarations you'd put for `getCurrentChannel`

"would apps be likely to handle all of those codes" - I doubt they'd handle any tbh, but useful for debugging.

For comparison, layouts had distinct errors for 'identity with missing name', 'identity with missing uuid', 'identity with missing name and uuid'. And I don't think distinguishing between "couldn't find the window" and "couldn't make even make sense if what you sent the provider" is getting too granular"

src/common/validation.ts
bryangaleopenfin: Yep, will change

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
bryangaleopenfin: Maybe, but wouldn't regard as essential for this PR

src/provider/APIHandler.ts
bryangaleopenfin: We're stripping off the 'readonly' any passing it to the ChannelProvider, which will regard the type as `any`, so there is a degree of trust there. Makes sense to change `APIHandler ` to `ChannelProvider` though

src/provider/APIHandler.ts
bryangaleopenfin: All integration tests fail if I make that change

test/demo/utils/fdc3RemoteExecution.ts
bryangaleopenfin: 👍

test/demo/utils/jestExtensions.ts
bryangaleopenfin: Changed to Channel ID

test/demo/utils/fdc3RemoteExecution.ts
bryangaleopenfin: Maybe? Suggest revisit in SERVICE-508?

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
bryangaleopenfin: Hmmm, I feel like Default = 'white', and that we're presenting Default as we are, is a UI side decision

src/client/contextChannels.ts
bryangaleopenfin: Done

src/common/validation.ts
bryangaleopenfin: Actually, we're calling this precisely in the cases where identity has been deserialized, so don't feel like it's necessary

src/common/errors.ts
bryangaleopenfin: Modified following discussion. TypeError now used

src/client/utils/async.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Creates a deferred promise and returns it along with handlers to resolve/reject it imperatively
+ * @returns a tuple with the promise and its resolve/reject handlers
+ */
+export function deferredPromise<T = void>(): [Promise<T>, (value?: T) => void, (reason?: any) => void] {
```
pjbroadbent: Seems this is only used in the provider - can we move out of `src/client`?

src/client/utils/validation.ts
```diff
@@ -0,0 +1,32 @@
+import {Identity} from 'openfin/_v2/main';
+
+/**
+ * Validates the provided identity and returns an identity stripped of any extraneous properties
+ */
+export function parseIdentity(identity: Identity): Identity {
```
pjbroadbent, line 10: To keep all our non-external stuff in one place, can we put these in `internal.ts`?

Or if we'd like to keep separate, at least add a `@hidden` comment to the top, to keep them out of the tsdoc.

src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
pjbroadbent: Was referring more to the colour/name being chosen by the application, rather than being prescribed by the service as per the other channels.

As discussed, can defer this for now though. Would require an API change to add those in later, but would be a non-breaking change.

src/common/validation.ts
pjbroadbent: Is that true? Looks like this is also called from the client - before anything is sent over the channel.

This is called by `parseIdentity`, which is called by any client API that expects an `Identity` object. That means this could potentially be called with anything?..

src/common/errors.ts
pjbroadbent: 👍 

The errors in Layouts don't have any kind of error codes, just error messages. IMO, we should still keep those messages as detailed and specific as possible - but the error codes should be more generalised. Current approach looks good.

test/demo/utils/jestExtensions.ts
```diff
@@ -15,10 +15,10 @@ declare global {
 
             /**
              * Used to test that a RemoteChannel represents the expected client-side Channel
-             * @param channel An object with optional `id` and `type` properties that if present we expect the channel to match
+             * @param channel An object with optional `id` and `type` properties that if present we expect the channel to match, or the `ChannlId`
```
pjbroadbent: typo: `ChannlId`

src/common/validation.ts
bryangaleopenfin: Oh, yes, no, no it is not true! Fixed

src/client/utils/async.ts
bryangaleopenfin: Done

bryangaleopenfin:
All comments so far responded to. Ready for re-review

tomrobertsOF (approved):

pjbroadbent (approved):

PR #107, opened 2019-06-05
[SERVICE-462] Intent resolver UI

rypete:
Updates the resolver UI to match the spec provided.  No functionality changes are made apart from the 'always remember' select box, which is missing from the spec.


res/provider/ui/selector.html
```diff
@@ -1,16 +0,0 @@
-<html>
-<head>
-    <title>App Selector</title>
-
-    <!-- Font Awesome -->
-    <script src="https://use.fontawesome.com/ef47d18271.js"></script>
-
-    <link href="https://www.w3schools.com/lib/w3-theme-blue-grey.css" rel="stylesheet">
```
tomrobertsOF, line 8: RIP w3-css

src/provider/common/Injectables.ts
```diff
@@ -11,7 +11,7 @@ enum Injectable {
     ENVIRONMENT,
     INTENT_HANDLER,
     MODEL,
-    SELECTOR,
+    RESOLVER,
```
tomrobertsOF, line 14: Any particular reason for this rename?

src/provider/controller/ResolverHandler.ts
```diff
@@ -11,63 +11,36 @@ import {SERVICE_IDENTITY} from '../../client/internal';
...
 })();
 
 /**
- * When the user is shown an app selector popup, they have the option of telling
- * the service how to handle similar intents in the future.
- *
- * This enum defines the options available to users.
+ * Data passed to app resolver when it is invoked by the provider
  */
-export const enum DefaultAction {
```
tomrobertsOF, line 32: If we're removing the defaults stuff from the UI does it make sense to also cleanup all of the dafaults parts in the provider? (more for @pjbroadbent )

res/provider/ui/resolver/css/resolver.css
```diff
@@ -0,0 +1,146 @@
...
+#header #exit:hover {
+    background-color: #3e3d51;
+    border-radius: 10px;
+}
+
+/*
+    App list
+*/
+
+#app-list {
```
tomrobertsOF: To match what we're doing with other react stuff, shouldn't the css for a component be in a `ComponentName.css` file alongside the component's `.tsx` file and bundled in?

rypete: Yes, you are correct.  However, there is a current problem with the akkurat ttf font and webpack so we had to do it the 'old school' way.

src/provider/common/Injectables.ts
rypete: Its no longer called selector

res/provider/ui/resolver/css/resolver.css
tomrobertsOF: Ok, fair enough. Are we planning on resolving this at any point, and if so can we make a JIRA to track.

src/provider/controller/ResolverHandler.ts
rypete: Can you point to the parts you are referring to?

src/provider/common/Injectables.ts
tomrobertsOF: So that's a no I guess

src/provider/controller/ResolverHandler.ts
tomrobertsOF: Oh, guess that all got removed during the refactor without me noticing. Never mind then.

tomrobertsOF (approved):
Approved assuming the final design has been signed off by whoever should be approving that kind of thing.

res/provider/ui/resolver/css/resolver.css
rypete: Yes but it will need to be a service tooling fix.  We have a JIRA already for it: SERVICE-527

tomrobertsOF: 👍

src/provider/common/Injectables.ts
rypete: Sorry let me be more specific.  The JIRA refers to the UI as the 'resolver'.  It was coded using this in mind.  Speaking to @pjbroadbent it sounds like selector was used previously as there an existing resolver component which is no more.  The selector was not a great name so I think it is a welcomed change.

tomrobertsOF: Fair enough. Thanks!

pjbroadbent: The "selector" name was to avoid confusion with the `resolveIntent` API's, which returned an array of applications and didn't actually use the UI.

Outside of the codebase, the UI has always been referred to as the 'resolver' though, and since the "resolve" API's were renamed, it made sense to also rename the UI implementation to match it's commonly-understood name.

res/provider/ui/resolver/css/resolver.css
pjbroadbent: Can we keep the react-related css in the previous setup, and have this file only include the font stuff?

res/provider/ui/resolver/css/resolver.css
```diff
@@ -0,0 +1,146 @@
+@font-face {
+    font-family: Akkurat;
+    src: url(../assets/Akkurat.ttf);
+}
+
+:root {
+    /*
+        The standard left padding which everything is pushed over by.
+    */
+    --left-padding: 33px;
```
pjbroadbent: Can this padding be on the container, rather than each of the components?

rypete: No, the bars for each app need to be 100% width (of the window), otherwise you will see the borders being cut at the beginning of the padding.

rypete: As discussed, no it cant as it will distort the full width elements

ottah:
[Exit icon](https://raw.githubusercontent.com/HadoukenIO/fdc3-service/b68391382b44ab6664a97bd0856eff8fe3490610/res/provider/ui/resolver/assets/exit.png) is 1120x1120 pixels.

src/provider/view/components/Resolver.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    return (
+        <div id="container">
+            <div id="header">
+                <h1>Select an Application</h1>
+                <div id="exit" onClick={onCancel}>
+                    <img src="assets/exit.png" />
+                </div>
+            </div>
+            <AppList applications={applications} openHandler={onAppOpen}/>
```
ottah: More a styling gripe, but props should be `onVerb` and the function should be `handleVerb`.

`... onOpen={handleOpen} ...`

rypete: Fixed

pjbroadbent:
Looks good, but some possible incoming design tweaks. Will hold this here for a bit whilst conversations progress.

pjbroadbent (approved):

PR #108, opened 2019-06-05
[SERVICE-479] Tests

spcfran:
- A lot of `raiseIntent` tests covering (hopefully) most cases for ad-hoc apps.
  - Including testing of selector - Had to modify the UI to include `id`s that can be easily targeted with puppeteer.
- Some `findIntent` tests for new ad-hoc app cases as well as running directory apps which haven't registered listeners for intents they are supposed to handle)
- Some general refactor/tidy up of tests.


src/client/internal.ts
```diff
@@ -22,6 +22,14 @@ export const SERVICE_IDENTITY = {
     name: 'fdc3-service'
 };
 
+/**
+ * The identity of the selector window, running on the FDC3 service application
+ */
+export const SELECTOR_IDENTITY: Identity = {
+    uuid: SERVICE_IDENTITY.uuid,
+    name: 'fdc3-selector'
```
bryangaleopenfin: Note the name-change to "resolver" in Ryan's PR

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -283,6 +283,17 @@ export async function findIntentsByContext(executionTarget: Identity, context: C
...
+    return ofBrowser.executeOnWindow(executionTarget, async function(this: TestWindowContext, elementSelector: string): Promise<boolean> {
+        const element = this.document.querySelector(elementSelector) as HTMLElement;
+        if (!element) {
+            return false;
+        }
+        element.click();
+        return true;
+    }, elementSelector);
+}
+
```
bryangaleopenfin, line 283: Cool!

test/demo/raiseIntent.inttest.ts
```diff
@@ -371,85 +368,393 @@ only the first listener is triggered exactly once with the correct context, and
...
+        listener.current = await fdc3Remote.addIntentListener(app, intent.type);
+    });
+
+    return listener;
+}
+
+//
+// Utilities
+//
+
```
bryangaleopenfin: Would prefer we follow a convention about how we order file contents, rather than needing these separators. We've generally had types and interfaces at the top of a file. Or there's an argument for pulling these out so they can be used in other tests

test/demo/raiseIntent.inttest.ts
```diff
@@ -52,235 +123,170 @@ describe('Intent listeners and raising intents', () => {
...
                             ResolveError.IntentTimeout,
-                            `Timeout waiting for intent listener to be added. intent = ${validPayload.intent}`
+                            `Timeout waiting for intent listener to be added. intent = ${validIntent.type}`
                         );
                     });
 
                     test('When calling unsubscribe from a second intent listener, then calling raiseIntent from another app, \
-the first listener is triggered exactly once with the correct context', async () => {
-                        const shortLivedListener = await fdc3Remote.addIntentListener(testAppInDirectory, validPayload.intent);
+only the first listener is triggered exactly once with the correct context, and the second is not triggered', async () => {
```
bryangaleopenfin: The "only" is redundant with "and the second is not triggered"

src/provider/view/components/AppCard.tsx
```diff
@@ -23,10 +23,11 @@ export function AppCard(props: AppCardProps): React.ReactElement {
         event.stopPropagation();
         openHandler(app);
     };
+    const className = 'app-card w3-card w3-round-large w3-button' + (selected ? ' selected' : '');
     return (
-        <div className={'app-card w3-card w3-round-large w3-button' + (selected ? ' selected' : '')} onClick={clickHandler} onDoubleClick={doubleClickHandler}>
+        <div id={app.name} className={className} onClick={clickHandler} onDoubleClick={doubleClickHandler}>
```
ottah: Is there any reason for adding an `id` attribute? Also can `app.name` have spaces (invalid css/html selector characters) in it?

src/client/internal.ts
```diff
@@ -22,6 +22,14 @@ export const SERVICE_IDENTITY = {
     name: 'fdc3-service'
 };
 
+/**
+ * The identity of the selector window, running on the FDC3 service application
+ */
+export const SELECTOR_IDENTITY: Identity = {
```
ottah: Rename to `RESOLVER_IDENTITY`.

bryangaleopenfin (commented):
Intermediate review - need to come back and have a proper look through tests

src/provider/view/components/AppCard.tsx
bryangaleopenfin: This is for the integration tests that need to fake clicks on specific buttons. Given that react escapes id strings so that you can't do anything actually breaking, I feel like I wouldn't worry too much about spaces (also, it all seems to work fine even with spaces in ids)

test/demo/raiseIntent.inttest.ts
```diff
@@ -371,85 +368,393 @@ only the first listener is triggered exactly once with the correct context, and
...
+    const appClicked = await fdc3Remote.clickHTMLElement(resolverWindowIdentity, `#${appName}`);
+    if (!appClicked) {
+        throw new Error(`App with name '${appName}' not found in resolver`);
+    }
+    await delay(100);
+    const openClicked = await fdc3Remote.clickHTMLElement(resolverWindowIdentity, '#btn-open');
+    if (!openClicked) {
+        throw new Error('Error clicking open button on resolver. Make sure it has id="btn-open".');
+    }
+    await delay(100);
```
bryangaleopenfin: If we need this delay after every click, could it go inside clickHTMLElement?

test/demo/resolveByIntent.inttest.ts
```diff
@@ -1,53 +1,178 @@
...
+    beforeEach(async () => {
+        await fin.Application.startFromManifest(app.manifestUrl);
+    });
+
+    afterEach(async () => {
+        await fin.Application.wrapSync(app).quit().catch(() => {});
+    });
+}
+
+interface AppIdentity {
```
bryangaleopenfin: Suggest do this as DirectoryAppIdentity, and NonDirectoryAppIdentity, both extending Identity. Then can drop the `?` from appId and we don't have the arbitrary choice of NonDirectoryApp  extedning AppIdentity  rather than the other way around

test/demo/resolveByIntent.inttest.ts
```diff
@@ -1,53 +1,178 @@
...
+function findIntent(intent: string, context: Context | undefined = undefined, sendFromIdentity?: Identity): Promise<DirectoryApp[]> {
+    return fdc3Remote.findIntent(sendFromIdentity || testManagerIdentity, intent, context)
+        .then(appIntent => appIntent.apps);
+}
+
+// TODO: The below helper functions and types are copied from `raiseIntent` tests. Should put everything in a util file
+
+function setupOpenDirectoryApp(app: AppIdentity) {
+    beforeEach(async () => {
+        await fdc3Remote.open(testManagerIdentity, app.uuid);
```
bryangaleopenfin: Our in open.inttest we're doing open(testManagerIdentity, app.name). Suggest do same for consistency

test/demo/resolveByIntent.inttest.ts
```diff
@@ -1,53 +1,178 @@
...
     describe('With context', () => {
         test.todo('TBD what resolving intent with context means as the spec is unclear');
     });
 });
+
+function getDirectoryAppsForIntent(directory: DirectoryApp[], targetIntent: string) {
+    return directory.filter(app => app.intents && app.intents.some(intent => intent.name === targetIntent)).map(app => app.name).sort();
+}
+
+function findIntent(intent: string, context: Context | undefined = undefined, sendFromIdentity?: Identity): Promise<DirectoryApp[]> {
```
bryangaleopenfin: With this and other similar helpers, how much does the final optional parameter actually get used? It worries me that in cases where it does get used, the parameter order is backwards from `fdc3Remote` conventions, hurting readability of the tests.

I'm cautiously tempted to say if you want to make the call from something other than the testManagerIdentity, just use the fd3Remote, without going through this helper.

I also definitely feel iffy about wrapping a helper around the specific call we're actually testing. Would be interested in a second opinion on this

test/demo/resolveByIntent.inttest.ts
```diff
@@ -1,53 +1,178 @@
...
+    return fdc3Remote.findIntent(sendFromIdentity || testManagerIdentity, intent, context)
+        .then(appIntent => appIntent.apps);
+}
+
+// TODO: The below helper functions and types are copied from `raiseIntent` tests. Should put everything in a util file
```
bryangaleopenfin: Can pulling this out be part of this PR?

test/demo/raiseIntent.inttest.ts
spcfran: The delay after each click could be specific to the action in particular e.g. the click to select the app should need less time than the click on the `Open` button, after which we wait for the resolver window to close. I didn't put much thought on the values here, they just seemed to work :)
Perhaps it would still be a good idea to add a small delay on `clickHTMLElement` itself to at least give React some time to propagate UI changes - that way we could omit explicit delays after simple clicks (like the first one in this function), and let consumers add an additional delay if required (like the second one)

test/demo/resolveByIntent.inttest.ts
spcfran: Actually it never gets used :) I agree it's best to remove it and in the rare case a different identity needs to be used, use the `fdc3Remote` form explicitly.

Re. your last concern - Let's see what @pjbroadbent and the rest think. I personally think that such a simple, pure syntactic sugar wrapper doesn't hurt test rigorousness _(is that a word?)_ too much and it makes the code much more readable. Also the signature of the helper is actually closer to the FDC3 API.

test/demo/resolveByIntent.inttest.ts
spcfran: Good point. Also could be good candidates to go out into the utils file

src/provider/view/components/AppCard.tsx
spcfran: Good point here. I'm not sure about the format of `name` (nor `appId`); in the two sample directories we have (for demo app and the tests), all `name`s happen to be valid html IDs, but for example some `appId`s on the test directory (should we decide at some point to go with those) currently aren't valid to use with `querySelector('#id')`.
Perhaps it wouldn't hurt to add some escaping logic on both the resolver and the tests for extra robustness. Or maybe just leave as it is on the resolver, and call `querySelector` with `[id='something']` which is more permissive than `#something`

test/demo/raiseIntent.inttest.ts
spcfran: 👌 moved to separate file

src/client/internal.ts
```diff
@@ -24,6 +24,14 @@ export const SERVICE_IDENTITY = {
     name: 'fdc3-service'
 };
 
+/**
+ * The identity of the resolver window, running on the FDC3 service application
+ */
+export const RESOLVER_IDENTITY: Identity = {
```
pjbroadbent: This is only used in the provider, can we keep it somewhere in `src/provider`? `ResolverHandler.ts`?

test/demo/contextChannels.inttest.ts
```diff
@@ -47,7 +47,7 @@ async function setupWindows(...channels: (ChannelId|undefined)[]): Promise<Ident
         const channel = channels[i];
         const identity = appIdentities[i + offset];
 
-        await fdc3Remote.open(testManagerIdentity, identity.uuid);
+        await fdc3Remote.open(testManagerIdentity, identity.name);
```
pjbroadbent, line 50: Why `name`? `identity` here is a window identity, and `open` expects an application ID. The `identity.uuid` field is the part that identifies the app, with `identity.name` being a window identifier.

spcfran: As per Bryan's suggestion https://github.com/HadoukenIO/fdc3-service/pull/108#discussion_r293745141 . We should have a quick discussion about this and make it consistent everywhere

src/client/internal.ts
spcfran: Makes sense, it's only used in there and integration tests. 

Is it OK to export the const from `ResolverHandler.ts` so it can be imported from the test file, or just make a copy of it in `test\demo\constants.ts`?

test/demo/contextChannels.inttest.ts
bryangaleopenfin: FDC3 `open` goes by `name` from the Directory rather than `appID`. Though not sure what guarantees we have about that matching `name` or `uuid` in the `Identity` object. In either case, we're kinda abusing the `Identity` object right? We're just doing this for convenience and we get away with it because everything in our test apps matches.

Anyway, @pjbroadbent @spcfran, grab me whenever

pjbroadbent: Discussion helped, makes sense now 😃

src/client/internal.ts
pjbroadbent: Copy is fine, so long as it's only in the one place in the tests.

Current setup looks good

bryangaleopenfin:
Closing to continue last tidying in new branch

PR #110, opened 2019-06-14
[SERVICE-508] A client app can interact with a channels without joining

bryangaleopenfin:
This PR:
- Adds ability to broadcast on and listen to a channel without first joining
- Reorganizes context channel tests, splitting into multiple files
- Changes provider logic, so that a non-directory app gets added to the model when it connects, rather than when it adds an intent listener
- Removed ChannelModel and replaces it with ChannelHandler, better following provider conventions
- Pulls out RemoteChannel into a new class
- Tidies exported documentation. Text is not final, but expected types/functions/etc. only should be exposed


res/test/test-app.html
```diff
@@ -8,8 +8,22 @@
     </script>
     <script>
 
-        function errorHandler(error) {
-            throw fdc3.serializeError(error);
```
bryangaleopenfin, line 12: `serializeError` is intentionally not exported, so we can't straightforwardly reuse code here

src/client/contextChannels.ts
```diff
@@ -43,6 +45,10 @@ export interface ChannelChangedEvent {
     previousChannel: Channel|null;
 }
 
+interface ChannelContextListener extends ContextListener {
+    id: ChannelId;
```
pjbroadbent: `channel: Channel`?..

src/client/contextChannels.ts
```diff
@@ -92,13 +114,67 @@ abstract class ChannelBase {
...
+     * Broadcasts the given context on this channel.
+     *
+     * Note that this function can be used without first joining the channel, allowing applications to broadcast on
+     * channels that they aren't a member of.
+     *
+     * This broadcast will be received by all windows that are members of this channel, *except* for the window that
+     * makes the broadcast. This matches the behaviour of the top-level FDC3 `broadcast` function.
+     *
+     * @param context The context to broadcast to all windows on this channel
+     * @throws `TypeError`: If `context` is not a valid Context
```
pjbroadbent: `@link` Context

src/client/contextChannels.ts
```diff
@@ -4,9 +4,11 @@
...
-import {parseIdentity} from './utils/validation';
-import {tryServiceDispatch} from './connection';
-import {APIFromClientTopic, DesktopChannelTransport, ChannelTransport} from './internal';
+import {parseIdentity, parseContext, validateEnvironment, parseChannelId} from './validation';
+import {tryServiceDispatch, getServicePromise} from './connection';
+import {APIFromClientTopic, DesktopChannelTransport, ChannelTransport, APIToClientTopic, ChannelContextPayload} from './internal';
+import {Context} from './context';
+import {ContextListener} from './main';
 
 export type ChannelId = string;
```
pjbroadbent, line 17: Is this alias still a useful addition? The new API has much less focus on channel ID's, given that we now have channel objects.

src/client/internal.ts
```diff
@@ -53,6 +56,7 @@ export enum APIFromClientTopic {
 export enum APIToClientTopic {
   INTENT = 'INTENT',
   CONTEXT = 'CONTEXT',
+  CHANNEL_CONTEXT = 'CHANNEL-CONTEXT'
```
pjbroadbent: Are these names clear? Feels like there should be a verb in there. `CHANNEL_CONTEXT_CHANGED`?..

src/client/validation.ts
```diff
@@ -0,0 +1,113 @@
...
+export function parseContext(context: Context): Context {
+    validateContextIsWellFormed(context);
+
+    return context;
+}
+
+/**
+ * Validates the provided ChannelId
+ */
+export function parseChannelId(channelId: ChannelId): ChannelId {
```
pjbroadbent: Can we collapse the `validate*` methods into these `parse*` methods? These one-liner functions seem a little redundant.

src/client/validation.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (error) {
+        throw new TypeError(`${safeStringify(context, 'Provided Context')} is not a valid Context`);
+    }
+}
+
+/**
+ * Checks that the provided ChannelId is a string
+ */
+function validateChannelIdIsWellFormed(channelId: ChannelId): void {
+    if (channelId === null || typeof channelId !== 'string') {
```
pjbroadbent: `typeof null` is `"object"`, so initial check isn't strictly required.

src/provider/controller/ChannelHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+];
+
+
+@injectable()
+export class ChannelHandler {
+    public readonly onChannelChanged: Signal1<EventTransport<ChannelChangedEvent>> = new Signal1<EventTransport<ChannelChangedEvent>>();
+
+    private readonly _model: Model;
+
+    constructor(@inject(Inject.MODEL) model: Model,) {
```
pjbroadbent: Trailing comma

src/provider/controller/ChannelHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+        this._model = model;
+
+        this._model.onWindowAdded.add(this.onModelWindowAdded, this);
+        this._model.onWindowRemoved.add(this.onModelWindowRemoved, this);
+    }
+
+    public registerChannels(): void {
```
pjbroadbent: Any reason for not making this part of the constructor?

src/provider/controller/ChannelHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+        const defaultChannel = new DefaultContextChannel(DEFAULT_CHANNEL_ID);
+
+        this._model.registerChannel(defaultChannel);
+        for (const channel of DESKTOP_CHANNELS) {
+            this._model.registerChannel(new DesktopContextChannel(channel.id, channel.name, channel.color));
```
pjbroadbent: Rather than the handler registering channels, would it make more sense for the model (or something in `src/provider/model` to own the channel list?

Taking layouts as an example, all the state/data lives within the model, and the controllers exist only to mutate that data.

src/provider/controller/ChannelHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    public getDesktopChannels(): ReadonlyArray<DesktopContextChannel> {
+        return this._model.channels.filter(channel => channel.type === 'desktop') as DesktopContextChannel[];
+    }
+
+    public getWindowsListeningToChannel(channel: ContextChannel): ReadonlyArray<AppWindow> {
```
pjbroadbent: The array returned by this (and the other functions below) is a new object, so would be fine for callers to mutate.

Unless the intention is for the window objects to be read-only? But the encapsulation of the window/channel classes should provide sufficient safety without needing to make them readonly(?)

src/provider/controller/ChannelHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+
+            if (this.isChannelEmpty(previousChannel)) {
+                previousChannel.clearStoredContext();
+            }
+
+            this.onChannelChanged.emit({type: 'channel-changed', identity: appWindow.identity, channel, previousChannel});
+        }
+    }
+
+    public setLastBroadcastOnChannel(channel: ContextChannel, context: Context): void {
```
pjbroadbent, line 59: Could we do this with looser coupling?

To store the last context on each channel, I would've imagined that this handler would listen to a signal on the context handler, rather than have the context handler call-in to this handler.

src/provider/model/ContextChannel.ts
```diff
@@ -1,3 +1,87 @@
...
+    readonly type: string;
+
+    setLastBroadcastContext(context: Context): void;
+    getStoredContext(): Context | null;
+    clearStoredContext(): void;
+
+    serialize(): ChannelTransport;
+}
+
+abstract class ContextChannelBase implements ContextChannel {
```
pjbroadbent, line 15: Is having both an interface and a base class redundant? Can we call the base class `ContextChannel` and remove the interface?

src/provider/model/ContextChannel.ts
```diff
@@ -1,3 +1,87 @@
...
+
+    public setLastBroadcastContext(context: Context) {
+
+    }
+
+    public clearStoredContext(): void {
+
+    }
+
+    public constructor(id: ChannelId) {
```
pjbroadbent: constructor before other methods?

src/provider/model/Model.ts
```diff
@@ -134,6 +133,30 @@ export class Model {
...
+    }
+
+    /**
+     * Registers an appWindow in the model
+     * @param appInfo Application info, either from the app directory, or 'crafted' for a non-registered app
+     * @param identity Window identity
+     * @param isInAppDirectory boolean indicating whether the app is registered in the app directory
+     */
+    private registerWindow(appInfo: Application, identity: Identity, isInAppDirectory: boolean): AppWindow {
+        // TODO: Should apps not in the directory have a channel?
```
pjbroadbent: Yeah - we should allow non-directory apps to belong to a channel. So long as the app imports the client library.

test/demo/contextChannels/broadcastOnChannel.inttest.ts
```diff
@@ -0,0 +1,321 @@
...
+/**
+ * Tests Channel.broadcast(), its interaction with Channel.getCurrentContext(), and Channel.addContextListener
+ */
+
+const testContext = {type: 'test-context', name: 'contextName1', id: {name: 'contextID1'}};
+
+let fin: Fin;
+
+beforeAll(async () => {
+    fin = await connect({address: `ws://localhost:${process.env.OF_PORT}`, uuid: 'TEST-contextChannels-broadcastOnChannel.inttest.ts'});
```
pjbroadbent: Does this test file need to create its own fin connection? Can it use `import {fin} from './utils/fin';` like the others?

test/demo/contextChannels/join.inttest.ts
```diff
@@ -0,0 +1,446 @@
...
+    describe('When a non-directory app has been started', () => {
+        beforeEach(async () => {
+            await fin.Application.startFromManifest(testAppNotInDirectory.manifestUrl);
+        }, appStartupTime);
+
+        afterEach(async () => {
+            await fin.Application.wrapSync(testAppNotInDirectory).quit(true);
+        });
+
+        test('When the channel is the default channel, result contains the non-directory app', async () => {
```
pjbroadbent: Should we also have a `After closing the non-directory app...` test here, as the FDC3-app test above?

test/demo/contextChannels/join.inttest.ts
```diff
@@ -0,0 +1,446 @@
...
+            await notInDirectoryApp.quit(true);
+        }
+
+        const notFdc3App = fin.Application.wrapSync(testAppNotFdc3);
+        if (await notFdc3App.isRunning()) {
+            await notFdc3App.quit(true);
+        }
+    });
+
+    type TestParam = string | Identity | (() => Promise<any>);
```
pjbroadbent: Can this be defined as a tuple?.. To keep the type-safety without unions/casts? Kinda depends how `each` is defined, I guess.

```
type TestParam = [string, Identity, () => Promise<any>];
```

test/demo/contextChannels/join.inttest.ts
```diff
@@ -0,0 +1,446 @@
...
+        const nonExistentWindowIdentity: Identity = {uuid: 'does-not-exist', name: 'does-not-exist'};
+
+        await expect(blueChannel.join(nonExistentWindowIdentity)).
+            toThrowFDC3Error(
+                IdentityError.WindowWithIdentityNotFound,
+                `No connection to FDC3 service found from window with identity: ${JSON.stringify(nonExistentWindowIdentity)}`
+            );
+    });
+
+    describe('When a non-FDC3 has been started', () => {
```
pjbroadbent: Some weird wording here in this block and the one below...
- "When a non-FDC3 _app_ has been started"
- "If _a_ non-FDC3 app ..."
- (etc)

test/provider/ContextHandler.unittest.ts
```diff
@@ -27,29 +31,34 @@ function createMockAppWindowWithName(name: string): AppWindow {
     return mockAppWindow;
 }
 
-function setChannelModelToProvideWindows(...windows: AppWindow[]): void {
+function setGetChannelMembersToReturn(...windows: AppWindow[]): void {
```
pjbroadbent: Is this equivalent to `mockGetChannelMembers.mockReturnValue(windows);`?

src/client/contextChannels.ts
bryangaleopenfin: Are you saying let's have a full-fat channel object here rather than the id? This only ever gets used for checking against the id of incoming objects, and this interface isn't exposed to users anywhere so we can change it later if there's need to

src/client/contextChannels.ts
bryangaleopenfin: Sorted. As an aside - the `{@link Object.method}` links don't work at all. Remove? Dig into? (Silly question probably 😛)

src/client/internal.ts
bryangaleopenfin: Then it becomes inconsistent with `CONTEXT`, and I'd rather keep consistency. Unless we're cool with this being a breaking change, and these become `INTENT_RAISED`, `CONTEXT_SENT`, `CHANNEL_CONTEXT_SENT`?

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: The constructor modifying the model feels surprising. Would rather be have this happen from someone explicitly asking for it to happen

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: Until we this is configurable, there isn't really a right place for this to live. I don't think Model owning the list is less wrong that ChannelHandler owning the list, and pulling it into it's own object - I'm not sure what that gains us. Could potentially just pull out a list the ChannelHandler reads and then puts into the Model. But like, this will all change when we implement this being configurable and this doesn't cause any problems until then.

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: Yeah, this this kinda a leftover from when it did return an array we were still using. Felt safer though? Like, do we always want to guarantee that what we return will be a copy

src/provider/model/ContextChannel.ts
bryangaleopenfin: Fixed

src/provider/model/ContextChannel.ts
bryangaleopenfin: The interface is used in testing - see `mocks.ts`

test/demo/contextChannels/broadcastOnChannel.inttest.ts
bryangaleopenfin: Missed this was a thing. Fixed

test/demo/contextChannels/join.inttest.ts
bryangaleopenfin: Huh, I'm annoyed because I was sure I tried this. Will fix!

src/client/contextChannels.ts
bryangaleopenfin: I'm so torn!

test/provider/ContextHandler.unittest.ts
bryangaleopenfin: Yes.

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: Removed on principle of least 'huh?'

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: `ContextHandler` makes calls to `ChannelHandler` to figure out who to broadcast to. So if we had `ChannelHandler` listen to `ContextHandler`, we'd now have a circular dependency between the two

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: Changed. the trade-off is that Model is now aware of the specific logic of going from the config object to a `DefaultContextChannel`/`DesktopContextChannel`

src/client/internal.ts
bryangaleopenfin: Hmm, best I have for this is `HANDLE_CHANNEL_CONTEXT`. Any narrow verb doesn't work, as this can be sent due to joining a channel, and app explicitly broadcasting a context, or due to an app being opened with a context. And more generic verbs like 'sent' are implicit just from the fact we're setting up an API

bryangaleopenfin:
Ready for re-review

src/provider/controller/ChannelHandler.ts
```diff
@@ -0,0 +1,88 @@
+import {injectable, inject} from 'inversify';
+
+import {Model} from '../model/Model';
+import {Inject} from '../common/Injectables';
+import {DEFAULT_CHANNEL_ID, ChannelId, FDC3Error, ChannelError, ChannelChangedEvent, Context} from '../../client/main';
+import {DesktopContextChannel, DefaultContextChannel, ContextChannel} from '../model/ContextChannel';
```
ottah: `DEFAULT_CHANNEL_ID` & `DefaultContextChannel` are unused.

src/provider/model/Model.ts
```diff
@@ -1,12 +1,16 @@
-import {injectable, inject} from 'inversify';
+import {injectable, inject, id} from 'inversify';
 import {Identity} from 'openfin/_v2/main';
 
 import {Application, AppName, AppId} from '../../client/directory';
 import {Inject} from '../common/Injectables';
-import {Signal0} from '../common/Signal';
+import {Signal0, Signal1} from '../common/Signal';
+import {ChannelId, DEFAULT_CHANNEL_ID, DefaultChannel, DesktopChannel} from '../../client/main';
+import {APIHandler} from '../APIHandler';
```
ottah, line 8: Unused imports

test/demo/contextChannels/join.inttest.ts
```diff
@@ -0,0 +1,440 @@
...
+            testAppInDirectory1,
+            async () => fdc3Remote.open(testManagerIdentity, testAppInDirectory1.name)
+        ], [
+            'a non directory app',
+            testAppNotInDirectory,
+            async () => fin.Application.startFromManifest(testAppNotInDirectory.manifestUrl).then(() => {})
+        ]
+    ];
+
+    describe.each(testParams)('When %s has been starterd', (titleParam: string, appIdentity: Identity, openFunction: () => Promise<any>) => {
```
ottah: Typo

src/client/internal.ts
```diff
@@ -141,6 +151,23 @@ export interface ChannelJoinPayload {
...
+    id: ChannelId,
+}
+
+export interface ChannelAddContextListenerPayload {
+    id: ChannelId;
+}
+
+export interface ChannelRemoveContextListenerPayload {
+    id: ChannelId;
+}
```
ottah, line 169: Is there plans to expand the data within these Channel payloads in the future? The majority of the interfaces in this file are all the same, could they not be merged into 1? `ChannelPayload` maybe?

src/client/main.ts
```diff
@@ -277,13 +277,60 @@ export function addContextListener(handler: (context: Context) => void): Context
...
+
     eventEmitter.removeListener(eventType, handler);
 }
 
 function hasIntentListener(intent: string): boolean {
     return intentListeners.some(intentListener => intentListener.intent === intent);
 }
+
+if (typeof fin !== 'undefined') {
+    getServicePromise().then(channelClient => {
```
ottah, line 296: Can `validateEnvironment()` be used here aswell? Depends if we want to throw an error before any calls are made by a client.

src/provider/model/Model.ts
```diff
@@ -186,13 +210,66 @@ export class Model {
...
+        if (!appWindow) {
+            let appInfo: Application;
+
+            // Attempt to copy appInfo from another appWindow in the model from the same app
+            const appWindowFromSameApp = this.findWindowByAppId(identity.uuid);
+            if (appWindowFromSameApp) {
+                appInfo = appWindowFromSameApp.appInfo;
+            } else {
+                // There are no appWindows in the model with the same app uuid - Produce minimal appInfo from window information
+                const application = fin.Application.wrapSync(identity);
```
ottah: `application` is unused.

ottah (commented):
Good work Bryan.
Still going through, but here are few minor comments.

src/client/main.ts
bryangaleopenfin: getServicePromise will do its own environment checking, so shouldn't need `validateEnvironment` here

src/client/internal.ts
bryangaleopenfin: Yeah.... I kinda like it as it is though can't explain it well! It's a tidy convention, and the there's no fundamental reason why the types are the same, they "just happen" to have the same contents, so I wouldn't want to tie them together

src/client/main.ts
pjbroadbent: We deliberately don't throw just from the import - we used to, until we added some extra guards.

This was a client request - throwing on import can make unit testing difficult. We want to immediately connect to the provider when running in OpenFin, but when running in a non-OF environment, we want to only throw if/when you call an API (and not before).

ottah: Ah, that makes sense.

pjbroadbent (approved):

ottah (approved):

PR #111, opened 2019-06-21
[NA] non-directory apps in demo launcher

spcfran:
- Create duplicate manifests for each of the apps in the launcher, only with a different UUID so that they don't appear in the app directory.
- Add some buttons to the launcher for creating these non-directory apps.
- Create a fake news app that exists only as a non-directory app, with a launcher button


src/demo/apps/LauncherApp.tsx
```diff
@@ -34,10 +34,48 @@ export function LauncherApp(): React.ReactElement {
...
-            {applications.map((app, index) => <AppCard key={app.appId + index} app={app} handleClick={openApp} />)}
+            {applications.map((app, index) => <AppCard key={app.appId + index} app={app} handleClick={openApp} isDirectoryApp />)}
+            <hr/>
+            <h2>Non-directory apps</h2>
+            {NON_DIRECTORY_APPS.map((app, index) => <AppCard key={app.appId + index} app={app} handleClick={launchApp} />)}
         </div>
     );
 }
+
+const NON_DIRECTORY_APPS: Application[] = ([
```
bryangaleopenfin, line 73: There's probably not a good solution here, but I feel like hard-coding a 'nice' title and description would be better than presenting raw IDs in the demo Launcher

src/demo/apps/LauncherApp.tsx
```diff
@@ -34,10 +34,48 @@ export function LauncherApp(): React.ReactElement {
...
+] as Array<[string, string]>).map(([id, icon]) => ({
+    appId: `${id}-nodir`,
+    name: `${id}-nodir`,
+    manifestType: 'openfin',
+    manifest: `http://localhost:3923/demo/configs/non-directory/app-${id}-nodir.json`,
+    icons: [
+        {icon: `http://localhost:3923/demo/img/app-icons/${icon}.svg`}
+    ],
+    title: `${id}`,
+    description: `Sample ${id} (no dir)`
```
bryangaleopenfin: Suggest either drop the '(no dir)' entirely, since these are already presented in the 'Non-directory apps' section, or replace with '(non-directory)' since this isa front-end, user facing thing

tomrobertsOF (approved):
All fine by me!

src/demo/apps/LauncherApp.tsx
```diff
@@ -34,10 +34,48 @@ export function LauncherApp(): React.ReactElement {
...
+    };
+
     return (
         <div>
             <h1>Launcher</h1>
-            {applications.map((app, index) => <AppCard key={app.appId + index} app={app} handleClick={openApp} />)}
+            {applications.map((app, index) => <AppCard key={app.appId + index} app={app} handleClick={openApp} isDirectoryApp />)}
+            <hr/>
+            <h2>Non-directory apps</h2>
+            {NON_DIRECTORY_APPS.map((app, index) => <AppCard key={app.appId + index} app={app} handleClick={launchApp} />)}
```
bryangaleopenfin: Suggest make isDirectoryApp explicit for both, it's not clear which would be the sensible default

src/demo/components/launcher/AppCard.tsx
```diff
@@ -6,11 +6,12 @@ import {Application} from '../../../client/directory';
 
 interface AppCardProps {
     app: Application;
+    isDirectoryApp?: boolean;
```
bryangaleopenfin: Suggest drop the `?`, as it's not obvious what `undefined` would be in this case

src/demo/components/news/NewsCard.tsx
```diff
@@ -0,0 +1,24 @@
...
+    headline: string;
+}
+
+export function NewsCard({item}: {item: NewsItem}) {
+    const {date, headline} = item;
+
+    return (
+        <li className="news-item w3-card">
+            <div className="news-item__date">
+                <span className="news-item__date__date">{date.toDateString()}</span>
```
bryangaleopenfin, line 16: Is the `__` a convention? Not sure what this gets us

src/demo/components/news/NewsData.ts
```diff
@@ -0,0 +1,28 @@
...
+        headline: getRandomHeadline(symbol)
+    };
+}
+
+function getRandomDate() {
+    return new Date(new Date().valueOf() - Math.random() * 1e9);
+}
+
+function getRandomHeadline(symbol: string) {
+    return `This is a fake news article for ${symbol}`;
```
bryangaleopenfin: Suggest go with something like 'Important news about ${symbol}'. We're not explicitly flagging our data as fake elsewhere, and "fake news" is a super politically charged term. Would also remame `getFakeNews` to `getNewsItem` or similar

src/demo/index.tsx
```diff
@@ -47,7 +49,7 @@ function SelectApp(props: SelectAppProps): React.ReactElement {
     const {uuid} = props;
     let selectedApp: JSX.Element;
 
-    switch (uuid) {
+    switch (uuid.replace('-nodir', '')) {
```
bryangaleopenfin: Treating IDs as anything other than totally opaque always makes me nervous! 😛

Suggest do the pre-processing on the UUID in one place - so move this up to the same place as we remove color from the UUID, and see if there's a sensible name instead of `uuid` on `SelectAppProps` since what we're dealing with is no longer really a UUID

src/provider/index.ts
```diff
@@ -162,14 +163,7 @@ export class Main {
             } else {
                 // There are no appWindows in the model with the same app uuid - Produce minimal appInfo from window information
                 const application = fin.Application.wrapSync(identity);
```
bryangaleopenfin: Consider moving this line inside `getApplicationInfo`? Just so our interaction with the low-level OpenFin API is self contained

res/demo/configs/non-directory/app-blotter-nodir.json
```diff
@@ -0,0 +1,20 @@
+{
+    "devtools_port": 9090,
+    "licenseKey": "64605fac-add3-48a0-8710-64b38e96a2dd",
+    "startup_app": {
+        "name": "FDC3 Blotter POC [non-directory]",
```
bryangaleopenfin: Suggest use title case a non-square brakets since this is ostensibly user-facing, so "FDC3 Blotter POC (Non-Directory)"

bryangaleopenfin (commented):
Looks good, just some minor feedback

src/demo/components/news/NewsCard.tsx
spcfran: It's based on [BEM](https://en.bem.info/methodology/quick-start/#nesting-1) although not strictly implemented. Didn't put much thought into it, happy to change to align better with the rest of the application

bryangaleopenfin: Ah k, nah, sounds reasonable, happy for this to stay as is

src/demo/index.tsx
spcfran: What about removing the hack altogether and explicitly referencing every uuid? There aren't that many after all:
```tsx
    switch (uuid) {
        case 'fdc3-launcher':
        case 'fdc3-launcher-nodir':
            selectedApp = <LauncherApp />;
            break;
```

bryangaleopenfin: Your call on that. We're already doing string manipulation to strip out the colour anyway, so could go either way

bryangaleopenfin (approved):

bryangaleopenfin:
Looks good, but removing my own approval for strictness, as I made the last few changes

pjbroadbent (approved):

pjbroadbent:
Merging without build due to CI issues, builds locally.

PR #112, opened 2019-06-27
[SERVICE-479] Allow non-directory applications to also use intents (Tests)

bryangaleopenfin:
This is adopted from Fran's PR with some tidy-up from me

This adds:
- Tests around non-directory apps using intents
- A common teardown method for use in all tests
- A constants file with our commonly used app IDs. (Other tests have not been updated to use these, and we would likely have merge hell if they were. It may be worth a further test tidy PR once both this and 508 are in)
- Helper functions in each test file to wrap fdc3Remote


src/provider/view/components/AppCard.tsx
```diff
@@ -17,10 +17,11 @@ export function AppCard(props: AppCardProps): React.ReactElement {
         event.stopPropagation();
         openHandler(app);
     };
+
     return (
-        <li className="app-card" onClick={clickHandler}>
+        <li id={app.name} className="app-card" onClick={clickHandler}>
```
ottah: Will `app.name` always be a valid CSS selector string? (not contain invalid characters e.g. spaces)

test/demo/findIntentsByContext.inttest.ts
```diff
@@ -1,25 +1,30 @@
...
-    beforeEach(async () => {
-        // The main launcher app should remain running for the duration of all tests.
-        await expect(fin.Application.wrapSync(testManagerIdentity).isRunning()).resolves.toBe(true);
-    });
+/**
+ * A context not accepted by any intent in any directory app
+ */
+const unknownContext = {
+    type: 'test.ContextTypeUnknown',
+    name: 'Test Name'
```
TheoOttahOF: Perhaps these could be moved into `constants.ts`?
A similar object to `invalidContext` is being used in `open.inttest.ts`.

src/provider/view/components/AppCard.tsx
bryangaleopenfin: Possibly not, but I think that's ok?

The ID is here for allowing tests to press the button for given test apps, so we're controlling the app IDs in this case. Where we're not controlling app IDs, given this is going through react rather than going directly into HTML, even if we try doing something nasty in the ID, like trying to close the quote marks surrounding the id and inject some HTML, everything still just works (including getElementById).

What it would break is if we were trying to style buttons by app name, but I wouldn't worry about that while we're not doing this.

If we definitely want to escape this, I think the most sensible thing to be would be to use `encodeURI`, but even then we might not have a valid ID, if it starts with a digit.

bryangaleopenfin: Ok, so, having spent more time digging into this rabbit hole than I probably should do, turns out this _is_ a rabbit hole 😛

The robust thing to do would be to
- Leave AppCard.tsx as is
- Import this - https://github.com/mathiasbynens/cssesc
- Use rewrite our query selector in `raiseIntent` as:
    `#${cssesc(appName, {isIdentifier: true})}`

and this then work no matter what's in the appName string - spaces, new lines, emoji...

But, I feel like this isn't work adding an extra dependency to solve

tomrobertsOF: This feels like one of those things that will come back as a bug in 3-6 months time once we've all forgotten about it. Is there not some better unique identifier we could use that would avoid needing the app name, or could we instead just come up with some deterministic mapping from app name to a selector-safe string, like a very simple hash or something.

Also, do we know what happens in the UI if an invalid string is passed in as an app name? I imagine it could cause weird things to the DOM depending on what it is, but maybe react can escape it properly to avoid that.

bryangaleopenfin: Having spent some time trying to break this, either react is escaping whatever is put in here, or behind the scenes, it's just doing `element.id = appName`, so escaping is a non-issue here

tomrobertsOF: Ok, that's fair. If it's only going to be a problem for the query selector in the test util then I'm less concerned as we have complete control over that env.

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -270,6 +270,17 @@ export async function findIntentsByContext(executionTarget: Identity, context: C
     }, context).catch(handlePuppeteerError);
 }
 
+export async function clickHTMLElement(executionTarget: Identity, elementSelector: string): Promise<boolean> {
```
tomrobertsOF, line 209: Puppeteer has some fairly good builtins for DOM interaction which we're using to do things similar to this in notifications testing, but probably not worth the effort of changing everything now that this is written.

test/demo/utils/common.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export interface NonDirectoryAppIdentity extends AppIdentity {
+    manifestUrl: string;
+}
+
+/**
+ * Quits an OpenFin app (or multiple in parallel) using `force=true` and swallowing errors
+ * @param apps Identities of the apps to quit
+ */
+export async function quitApp(...apps: Identity[]) {
```
tomrobertsOF: Annoying and nitpicky, but it would make more sense to me if this was called `quitApps` since it takes multiple.

test/demo/utils/common.ts
```diff
@@ -0,0 +1,110 @@
...
+    await Promise.all(apps.map(app => fin.Application.wrapSync(app).quit(true).catch(() => {})));
+}
+
+export async function waitForAppToBeRunning(app: Identity): Promise<void> {
```
tomrobertsOF, line 34: It may make sense to include some kind of timeout mechanism here to avoid waiting indefinitely if the app creation fails for some reason.

test/demo/utils/common.ts
```diff
@@ -0,0 +1,110 @@
...
+    while (!await fin.Application.wrapSync(app).isRunning()) {
+        await delay(100);
+    }
+
+    // Additional delay to ensure app window is ready for puppeteer connection
+    await delay(500);
```
tomrobertsOF, line 50: I wouldn't expect this delay to be necessary, but if we see problems without it then I guess its fine to leave it.

test/demo/constants.ts
```diff
@@ -1 +1,81 @@
+import {SERVICE_IDENTITY} from '../../src/client/internal';
+
+import {AppIdentity, NonDirectoryAppIdentity, DirectoryAppIdentity} from './utils/common';
+
 export const appStartupTime = 30 * 1000;
+
+export const testManagerIdentity: AppIdentity = {uuid: 'test-app', name: 'test-app'};
+
+export const resolverWindowIdentity: AppIdentity = {
```
tomrobertsOF: Could we not pull this from the provider somewhere? Wouldn't expect it to ever change, but seems dangerous defining it in two separate places.

test/demo/utils/common.ts
```diff
@@ -0,0 +1,110 @@
...
+import {fin} from './fin';
+import * as fdc3Remote from './fdc3RemoteExecution';
+import {delay} from './delay';
+
+export type Boxed<T> = {value: T}
+
+/**
+ * Identity with a mandatory `name`
+ */
+export interface AppIdentity extends Identity {
```
tomrobertsOF: This name may cause some confusion as strictly speaking an app in OF terms is only a uuid whereas a uuid/name pair is an OF window identifier. This is really more of a problem of consolidating OF nomenclature with FDC3, but maybe we could add some clarifier to this type to make it clear what it's an identifier for.

tomrobertsOF (commented):
Few comments around tests and utils

test/demo/utils/common.ts
bryangaleopenfin: Yeah, seeing problems without this. Before, the timeout in the loop was 500, which meant that *most* of the time, the app was given long enough

tomrobertsOF: Hmmm, fair enough. Its a one-off so not a huge problem.

test/demo/utils/common.ts
bryangaleopenfin: Yeah, this was a big conversation with Phil while you were away 😛. Basically the way we declare the constants holding data about our test apps was kinda odd, because sometimes name was being treated as an app directory name, and sometimes name was being treated as an Identity name, and putting a type to this data highlighted that,

I've renamed the type to reflect that it's kinda fuzzy what this type really _is_ and added a comment about the use of `name`

tomrobertsOF: Not the first time it's been talked about and certainly won't be the last. As long as it's mostly clear from context what it means then I think we should be fine.

bryangaleopenfin:
@TheoOttahOF, @tomrobertsOF I've resolved or replied to your feedback. Ready for another look

tomrobertsOF (approved):
Thanks! All good.

src/provider/view/components/AppCard.tsx
ottah: Yeah I guess it isn't that important at the moment.
This is how I would probably do it:
`<li className="app-card" data-appName={app.name}>...`

Using an escape/encoding method so it can be used in the selector.
The selector would then be:
`.app-card [data-appName="test-app-1"]`

ottah (approved):
Good work! 👌

src/provider/view/components/AppCard.tsx
bryangaleopenfin: Oooooooh. Actually yeah, that's really nice. I'll leave the escaping, but I'll make the change so it's using `data-appName`

tomrobertsOF: From memory that's how we do it in notifications and it works well.

bryangaleopenfin:
@TheoOttahOF, @tomrobertsOF Marking for re-review due to non-trivial merge from develop and a bit of post-merge tidy-up. Would appreciate a quick once-over of commits from today

src/client/contextChannels.ts
```diff
@@ -216,3 +315,21 @@ export function getChannelObject<T extends Channel = Channel>(channelTransport:
 
     return channel as T;
 }
+
+function hasChannelContextListener(id: ChannelId) {
```
ottah: Return type.

ottah (approved):

src/client/contextChannels.ts
bryangaleopenfin: :+1:

tomrobertsOF (approved):

PR #116, opened 2019-07-01
[SERVICE-509] Clients can listen for window-added and window-removed events on any channel

bryangaleopenfin:
This PR adds
- Ability to add `window-added` and `window-removed` listeners to channel objects
- `EventRouter` on the client side, to correctly dispatch events event from the channel object or to listeners added to the top level
- Changes to `EventHandler` on the provider side to add extra `target` information, that is sent across the IAB channel for use by `EventRouter`


src/client/contextChannels.ts
```diff
@@ -21,11 +26,80 @@ export type ChannelId = string;
  */
 export type Channel = DesktopChannel | DefaultChannel;
 
+export type FDC3ChannelEvent = ChannelWindowAddedEvent | ChannelWindowRemovedEvent;
+export type FDC3ChannelEventType = 'window-added' | 'window-removed';
```
pjbroadbent: Should be able to avoid duplication of event types here:

`export type FDC3ChannelEventType = FDC3ChannelEvent['type'];`

src/client/contextChannels.ts
```diff
@@ -316,10 +444,62 @@ export function getChannelObject<T extends Channel = Channel>(channelTransport:
     return channel as T;
 }
 
+/**
+ * @hidden
+ */
+export function dispatchChannelEvents(event: ChannelChangedEvent): void {
```
pjbroadbent: Not seeing where this is referenced?..

src/client/contextChannels.ts
```diff
@@ -183,6 +259,56 @@ export abstract class ChannelBase {
...
+     *
+     * Has no effect if `eventType` isn't a valid event, or `listener` isn't a callback registered against `eventType`.
+     *
+     * @param eventType The event being unsubscribed from
+     * @param listener The callback function to remove
+     */
+    public async removeEventListener<E extends FDC3ChannelEvent>(eventType: E['type'], listener: (event: E) => void): Promise<void> {
+        validateEnvironment();
+
+        channelEventEmitters[this.id].removeListener(eventType, listener);
```
pjbroadbent, line 306: The emitter is essentially a private member, but it's always accessed through a global. Is that weird?

Good encapsulation, though. :-)

src/client/contextChannels.ts
```diff
@@ -316,10 +444,62 @@ export function getChannelObject<T extends Channel = Channel>(channelTransport:
...
+
+        channelEmitter.emit('window-added', windowAddedEvent);
+    }
+}
+
 function hasChannelContextListener(id: ChannelId) {
     return channelContextListeners.some(listener => listener.channel.id === id);
 }
 
+function onWindowAdded(eventTransport: EventTransport<FDC3Event>): FDC3ChannelEvent {
```
pjbroadbent: Can we name these as "deserializers"? They sound like event handlers, but are really part of some "toolchain-level translation stuff". Or define them inline, like the emitter provider.

Is there a way to type the argument as `EventTransport<ChannelWindowAddedEvent>`, rather than needing a cast? The type string is the same, would've thought there's some TypeScript magic that would make it pick up the right type.

src/client/main.ts
```diff
@@ -82,11 +83,14 @@ export interface IntentListener {
 /**
  * Defines all events that are fired by the service
  */
-export type FDC3Event = ChannelChangedEvent;
-export type FDC3EventType = 'channel-changed';
+export type FDC3Event = FDC3MainEvent | FDC3ChannelEvent;
+export type FDC3EventType = FDC3MainEventType | FDC3ChannelEventType;
+
+export type FDC3MainEvent = ChannelChangedEvent;
+export type FDC3MainEventType = 'channel-changed'
```
pjbroadbent: Ditto for removing duplicated constants, also missing `;`.

src/provider/controller/ChannelHandler.ts
```diff
@@ -2,21 +2,22 @@ import {injectable, inject} from 'inversify';
...
 import {DesktopContextChannel, ContextChannel} from '../model/ContextChannel';
 import {AppWindow} from '../model/AppWindow';
-import {Signal1} from '../common/Signal';
-import {EventTransport} from '../../client/internal';
+import {Signal3} from '../common/Signal';
 
 @injectable()
 export class ChannelHandler {
-    public readonly onChannelChanged: Signal1<EventTransport<ChannelChangedEvent>> = new Signal1<EventTransport<ChannelChangedEvent>>();
+    public readonly onChannelChanged: Signal3<AppWindow, ContextChannel | null, ContextChannel | null>;
```
pjbroadbent, line 17: Can you add a comment to explain the meaning of the arguments? Not obvious from the method signature.

There's a "soft" convention for this in some of the other services: https://github.com/HadoukenIO/layouts-service/blob/develop/src/provider/model/DesktopWindow.ts#L277-L282

src/provider/controller/ChannelHandler.ts
```diff
@@ -26,20 +27,28 @@ export class ChannelHandler {
...
         return this._model.windows.filter(window => window.channel === channel);
     }
 
-    public getChannelContext(channel: ContextChannel): Context | null {
-        return channel.getStoredContext();
+    public getWindowsListeningForContextsOnChannel(channel: ContextChannel): AppWindow[] {
+        return this._model.windows.filter(window => window.hasContextListener(channel));
+    }
+
+    public getWindowsListeningForEventsOnChannel(channel: ContextChannel, eventType: FDC3ChannelEventType): AppWindow[] {
```
pjbroadbent, line 55: Is this a bit over-verbose? `getWindowsListeningToChannel`?

Does this make a distinction between if you're listening to a channel because you're in that channel, vs listening directly to a channel object?

test/demo/utils/RemoteChannel.ts
```diff
@@ -77,4 +84,43 @@ export class RemoteChannel {
...
+
+                const handler = (payload: FDC3ChannelEvent) => {
+                    this.receivedChannelEvents.push({listenerID, payload});
+                };
+
+                const unsubscribe = () => {
+                    channel.removeEventListener(eventType as any, handler as any);
+                };
+
+                channel.addEventListener(eventType as any, handler as any);
```
pjbroadbent, line 108: What's with the `any`'s here?

Do we not lint this file? These should need linter exceptions to pass `npm run check`..?

test/demo/utils/jestExtensions.ts
```diff
@@ -24,9 +24,13 @@ declare global {
 }
 
 expect.extend({
-    async toThrowFDC3Error<T = any>(received: Promise<T>, code: string, message?: string | RegExp) {
+    async toThrowFDC3Error<T = any>(received: Promise<T> | (() => any), code: string, message?: string | RegExp) {
```
pjbroadbent: `() => T`?..

test/provider/ChannelHandler.unittest.ts
```diff
@@ -0,0 +1,260 @@
...
+    it('If Model returns a channel, ChannelHandler returns the channel', () => {
+        const testChannel = new DesktopContextChannel('test', 'test', 0);
+
+        (mockModel.getChannel as jest.Mock<ContextChannel | null, [string]>).mockReturnValue(testChannel);
+
+        expect(channelHandler.getChannelById('test')).toEqual(testChannel);
+    });
+
+    it('If Model returns null, ChannelHandler throws an exception', () => {
+        (mockModel.getChannel as jest.Mock<ContextChannel | null, [string]>).mockReturnValue(null);
```
pjbroadbent: A lot of this casting seems unnecessary, there's Jest types that I think can clean this up (will need a cast when you create the mock, but then not when using it).

Can we discuss when you have the time?

src/provider/controller/ContextHandler.ts
```diff
@@ -40,8 +40,8 @@ export class ContextHandler {
      * @param context Context to send
      * @param source Window sending the context. It won't receive the broadcast
      */
-    public broadcast(context: Context, source: AppWindow): void {
-        this.broadcastOnChannel(context, source, source.channel);
+    public async broadcast(context: Context, source: AppWindow): Promise<void> {
+        return this.broadcastOnChannel(context, source, source.channel);
```
tomrobertsOF, line 44: I could probably check myself, but where does this await chain end? Is it once the broadcast reaches the provider, when the provider has sent to all clients, when the clients have all completed their callback, or some other timing? the spirit of broadcast is that it's fire-and-forget, so want to be sure we're only awaiting what is strictly necessary to catch problems.

src/provider/index.ts
```diff
@@ -3,7 +3,7 @@ import {inject, injectable} from 'inversify';
 import {Identity} from 'openfin/_v2/main';
 import {ProviderIdentity} from 'openfin/_v2/api/interappbus/channel/channel';
 
-import {RaiseIntentPayload, APIFromClientTopic, OpenPayload, FindIntentPayload, FindIntentsByContextPayload, BroadcastPayload, APIFromClient, IntentListenerPayload, GetDesktopChannelsPayload, GetCurrentChannelPayload, ChannelGetMembersPayload, ChannelJoinPayload, ChannelTransport, DesktopChannelTransport, GetChannelByIdPayload, EventTransport, ChannelBroadcastPayload, ChannelGetCurrentContextPayload, ChannelAddContextListenerPayload, ChannelRemoveContextListenerPayload} from '../client/internal';
+import {RaiseIntentPayload, APIFromClientTopic, OpenPayload, FindIntentPayload, FindIntentsByContextPayload, BroadcastPayload, APIFromClient, IntentListenerPayload, GetDesktopChannelsPayload, GetCurrentChannelPayload, ChannelGetMembersPayload, ChannelJoinPayload, ChannelTransport, DesktopChannelTransport, GetChannelByIdPayload, EventTransport, ChannelBroadcastPayload, ChannelGetCurrentContextPayload, ChannelAddContextListenerPayload, ChannelRemoveContextListenerPayload, ChannelAddEventListenerPayload, ChannelRemoveEventListenerPayload} from '../client/internal';
```
tomrobertsOF: This is a lot of imports from one file. It was originally done as one because there was a very limited set of types that needed exporting. Probably a question for @pjbroadbent too, but do you think it would make sense to start splitting internal by functionality rather than having it all in one. Not sure where we would make the splits, but maybe worth discussion?

tomrobertsOF (commented):
Couple of comments

test/provider/ChannelHandler.unittest.ts
bryangaleopenfin: Fixed. Notes:
- jest.spyOn doesn't seem to work as expected (throws an exception), maybe because the object I'm applying it to is already a mock, so using Object.definePropery
- Is there a way to make `Writable` (defined int `types.ts`) apply only to a single property?

test/demo/utils/jestExtensions.ts
bryangaleopenfin: :+1:

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: As discussed.

This is for windows listening on the channel object itself

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: 👍

src/client/contextChannels.ts
bryangaleopenfin: Removed

src/client/main.ts
bryangaleopenfin: :+1:

src/client/contextChannels.ts
bryangaleopenfin: :+1:

src/client/contextChannels.ts
bryangaleopenfin: 👍 

Naming changed, but trying typescript magic gives this:

foo


    public registerDeserializer<ChannelWindowAddedEvent>(eventType: 'window-added', handler: (event: EventTransport<ChannelWindowAddedEvent>) => ChannelWindowAddedEvent): void;
    public registerDeserializer<ChannelWindowAddedEvent>(eventType: 'window-removed', handler: (event: EventTransport<ChannelWindowRemovedEvent>) => ChannelWindowAddedEvent): void;
    public registerDeserializer<ChannelChangedEvent>(eventType: 'window-removed', handler: (event: EventTransport<ChannelChangedEvent>) => ChannelWindowAddedEvent): void;

    public registerDeserializer<E extends FDC3Event>(eventType: E['type'], handler: (event: EventTransport<E>) => E): void {
        this._deserializers[eventType] = (event: EventTransport<FDC3Event>) => handler(event as EventTransport<FDC3Event>);
    }

test/demo/utils/RemoteChannel.ts
bryangaleopenfin: As discussed

src/client/contextChannels.ts
bryangaleopenfin: `channelEventEmitters` gets used globally here:

        eventHandler.registerEmitterProvider('channel', (channelId: ChannelId) => {
            return channelEventEmitters[channelId];
        });

so if it lived on the Channel, it would need to be public

As a secondary "ugh I don't want to change this", having the emitter live on the ConxtextChannel caused problems when the Channels are sent across the wire as part of FDC3RemoteExecution

src/provider/index.ts
bryangaleopenfin: Hmm, the first split I'd probably make would to pull out the API and Payload definitions (which themselves I'd be reluctant to split further), which wouldn't reduce the size of this import.

tomrobertsOF: Yea, as I said it's hard to see where there could be a logical separation. Lets have a quick chat offline and if we can't think of anything better then lets just leave it for now.

src/provider/controller/ContextHandler.ts
bryangaleopenfin: Fine as discussed

src/provider/index.ts
bryangaleopenfin: Discussed - will leave as is

bryangaleopenfin:
@pjbroadbent @tomrobertsOF Ready for re-review. Note that the final commit, prompted by a discussion with Tom, changes Channel.broadcast to be fire-and-forget, to better match top-level `broadcast` behaviour

tomrobertsOF (approved):

src/client/contextChannels.ts
pjbroadbent: Needs a nasty cast in the `registerDeserializer` class, but managed to get something that works:

EventRouter.ts:
```
export type EventDeserializer<E extends FDC3Event> = (event: EventTransport<E>) => E;

export class EventRouter {
    // Don't need overrides here, we only do that on addEventListener for docs purposes
    public registerDeserializer<E extends FDC3Event>(eventType: E['type'], handler: EventDeserializer<E>): void {
        this._deserializers[eventType] = handler as unknown as EventDeserializer<FDC3Event>;
    }
}
```

contextChannels.ts:
```
function deserializeWindowAddedEvent(eventTransport: EventTransport<ChannelWindowAddedEvent>): ChannelWindowAddedEvent {
    const identity = eventTransport.identity;
    const channel = getChannelObject(eventTransport.channel!);
    const previousChannel = eventTransport.previousChannel ? getChannelObject(eventTransport.previousChannel) : null;

    return {type: 'window-added', identity, channel, previousChannel};
}

eventHandler.registerDeserializer('window-added', deserializeWindowAddedEvent);
```

test/provider/ChannelHandler.unittest.ts
pjbroadbent: This should do it:
`export type PartiallyWritable<T, P extends keyof T> = Writable<Pick<T, P>> & Omit<T, P>;`

Use like:
`let mockModel: jest.Mocked<PartiallyWritable<Model, 'onWindowAdded' | 'onWindowRemoved'>>;`

It'll take a certain subset of properties of `Writable<T>`, and combine that with the opposite subset of properties of `T`.

bryangaleopenfin: 👍

src/provider/controller/ChannelHandler.ts
bryangaleopenfin: Fiiine

src/client/contextChannels.ts
bryangaleopenfin: :+1:

bryangaleopenfin:
@pjbroadbent Ready for another look

pjbroadbent (approved):

PR #117, opened 2019-07-05
[SERVICE-506][SERVICE-504] Consider listeners of all application windows when selecting intent targets

bryangaleopenfin:
This PR:
- Changes `raiseIntent` to target all open windows of an app, rather than an arbitrary choice of one window
- Removes the `FindFilter` system from `Model`, and instead methods on `Model` that returned a single window will now return all relevant windows
- Tidies some naming in `AppWindow`
- Changes `ensureReadyToRevieveIntent` to `isReadyToReceiveIntent`, which returns a boolean rather than throwing an exception, since one window in an app not adding an intent listener does not necessarily indicate a failure
- Changes timeout logic to be "5 seconds from when the window was created" rather than "5 seconds from when the call was made" (this resolves SERVICE-504)
- Splits up and removes duplication from `raiseIntent` tests (suggest it will be easier to see changes looking at individual commits, rather than trying to find the "real" changes among my copying and pasting one file into two)

This story runs up against some technical debt, which I've logged in SERVICE-544


tomrobertsOF:
One minor housekeeping comment. If this also addresses SERV-504 can we add it to the PR title in the style that we did for e.g. [this layouts pr](https://github.com/HadoukenIO/layouts-service/pull/342): `[SERVICE-XXX][SERVICE-YYY] Title ...`

src/provider/model/Model.ts
```diff
@@ -82,32 +72,24 @@ export class Model {
...
-        const matchingWindow = this.findWindowByAppId(appInfo.appId, {prefer});
+    public async findOrCreate(appInfo: Application): Promise<AppWindow[]> {
+        const matchingWindows = this.findWindowsByAppId(appInfo.appId);
 
-        if (matchingWindow) {
-            await matchingWindow.focus();
-            return matchingWindow;
+        if (matchingWindows.length > 0) {
+            await Promise.all(matchingWindows.map(window => window.focus()));
+            return matchingWindows;
```
pjbroadbent, line 82: Focusing multiple windows?

Would `bringToFront` be more applicable?.. Though suppose we would still want to focus one of the windows.

src/provider/model/FinEnvironment.ts
```diff
@@ -181,26 +194,35 @@ class FinAppWindow implements AppWindow {
...
-            slot.remove();
-            throw new FDC3Error(ResolveError.IntentTimeout, `Timeout waiting for intent listener to be added. intent = ${intent}`);
+            const [didTimeout] = await withTimeout(Timeouts.ADD_INTENT_LISTENER - age, waitForIntentListenerAddedPromise);
+
+            if (didTimeout) {
+                slot.remove();
+                return false;
+            } else {
+                return true;
+            }
```
pjbroadbent: Despite knowing what was going on here, took me a while to figure out how this actually worked.

A version without utils, is this any easier to follow?..
```
// App may be starting - Give it some time to initialize and call `addIntentListener()`, otherwise timeout
await new Promise((resolve) => {
    // Wait for intent to be added
    const slot = this._onIntentListenerAdded.add(intentAdded => {
        if (intentAdded === intent) {
            slot.remove();
            resolve(true);
        }
    });

    // Return false if no listener added within timeout
    setTimeout(() => {
        slot.remove();
        resolve(false);
    }, Timeouts.ADD_INTENT_LISTENER);
});
```

src/provider/model/FinEnvironment.ts
```diff
@@ -181,26 +194,35 @@ class FinAppWindow implements AppWindow {
...
-                slot.remove();
-                resolve();
-            }
-        });
+        const age = this._creationTime === undefined ? undefined : Date.now() - this._creationTime;
 
-        const [didTimeout] = await withTimeout(Timeouts.ADD_INTENT_LISTENER, waitForIntentListenerAddedPromise);
+        if (age === undefined || age >= Timeouts.ADD_INTENT_LISTENER) {
+            // App has been running for a while
+            return this.hasIntentListener(intent);
```
pjbroadbent: We know that this would be false at this point. Could `return false;`, but intent may not be as clear..?

src/provider/controller/IntentHandler.ts
```diff
@@ -110,13 +109,27 @@ export class IntentHandler {
...
 
     private async fireIntent(intent: Intent, appInfo: Application): Promise<IntentResolution> {
-        const appWindow = await this._model.findOrCreate(appInfo, FindFilter.WITH_INTENT_LISTENER);
-        await appWindow.ensureReadyToReceiveIntent(intent.type);
-        const data = await this._apiHandler.channel.dispatch(appWindow.identity, APIToClientTopic.INTENT, {context: intent.context, intent: intent.type});
+        const appWindows = await this._model.findOrCreate(appInfo);
+
+        const dispatchResults = await Promise.all(appWindows.map(async (window: AppWindow): Promise<[boolean, any]> => {
+            if (await window.isReadyToReceiveIntent(intent.type)) {
+                const data = await this._apiHandler.channel.dispatch(window.identity, APIToClientTopic.INTENT, {context: intent.context, intent: intent.type});
```
pjbroadbent: Do we have a timeout on how long an app can take within its intent handler? Probably something we should add a story for.

src/provider/controller/IntentHandler.ts
```diff
@@ -110,13 +109,27 @@ export class IntentHandler {
...
+            throw new FDC3Error(ResolveError.IntentTimeout, `Timeout waiting for intent listener to be added. intent = ${intent.type}`);
+        }
+
+        const dispatchResultData = dispatchResults.filter((dispatchResult) => dispatchResult[1] !== undefined).map(dispatchResult => dispatchResult[1]);
+
         const result: IntentResolution = {
             source: appInfo.name,
             version: '1.0.0',
-            data
+            data: dispatchResultData.length === 0 ? undefined : dispatchResultData.length === 1 ? dispatchResultData[0] : dispatchResultData
```
pjbroadbent: Not sure about the logic here, feel like we should be more prescriptive in how we handle multiple intent listeners. Telling apps we may give you a value, or we may give you an array of values might make this hard for apps to handle.

As in, always return an array - or return the first non-falsey (or non-undefined) result, something like that..?

src/provider/controller/IntentHandler.ts
```diff
@@ -110,13 +109,27 @@ export class IntentHandler {
...
+            if (await window.isReadyToReceiveIntent(intent.type)) {
+                const data = await this._apiHandler.channel.dispatch(window.identity, APIToClientTopic.INTENT, {context: intent.context, intent: intent.type});
+                return [true, data];
+            } else {
+                return [false, undefined];
+            }
+        }));
+
+        if (dispatchResults.every(dispatchResult => dispatchResult[0] === false)) {
+            throw new FDC3Error(ResolveError.IntentTimeout, `Timeout waiting for intent listener to be added. intent = ${intent.type}`);
```
pjbroadbent: Message formatting is inconsistent here. Other error messages with params all (seem to) use either `blah blah: ${value}` or `blah '${value}' blah` style.

src/provider/index.ts
```diff
@@ -103,10 +103,12 @@ export class Main {
         }
 
         // This can throw FDC3Errors if app fails to open or times out
-        const appWindow = await this._model.findOrCreate(appInfo, FindFilter.WITH_CONTEXT_LISTENER);
+        const appWindows = await this._model.findOrCreate(appInfo);
 
         if (payload.context) {
-            return this._contextHandler.send(appWindow, parseContext(payload.context));
+            return Promise.all(appWindows.map(window => {
+                return this._contextHandler.send(window, parseContext(payload.context!));
```
pjbroadbent, line 115: Would this hang (or reject?) if one of the app's windows didn't have a context listener? Is that deferred to another story?

src/provider/model/FinEnvironment.ts
```diff
@@ -33,6 +33,8 @@ export class FinEnvironment extends AsyncInit implements Environment {
      */
     public readonly windowClosed: Signal1<Identity> = new Signal1();
 
+    private readonly windowCreationTimes: {[id: string]: number} = {};
```
pjbroadbent: These creation times have a fairly weird way of getting into the window object. This map ought to be redundant (since model has a map already, and windows know their own creation time) but isn't.

Worth a quick chat about how best to wire this up.

src/provider/model/FinEnvironment.ts
```diff
@@ -48,16 +50,23 @@ export class FinEnvironment extends AsyncInit implements Environment {
     }
 
     public wrapApplication(appInfo: Application, identity: Identity, channel: ContextChannel): AppWindow {
-        return new FinAppWindow(parseIdentity(identity), appInfo, channel);
+        identity = parseIdentity(identity);
+        const creationTime = this.windowCreationTimes.hasOwnProperty(getId(identity)) ? this.windowCreationTimes[getId(identity)] : undefined;
```
pjbroadbent: This is equivalent to just `const creationTime = this.windowCreationTimes[getId(identity)];`

src/provider/model/Model.ts
```diff
@@ -236,6 +176,33 @@ export class Model {
...
 
+    /**
+     * Registers an appWindow in the model
+     * @param appInfo Application info, either from the app directory, or 'crafted' for a non-registered app
+     * @param identity Window identity
+     * @param isInAppDirectory boolean indicating whether the app is registered in the app directory
+     */
+    private registerWindow(appInfo: Application, identity: Identity, isInAppDirectory: boolean): AppWindow {
+        const appWindow = this._environment.wrapApplication(appInfo, identity, this._channelsById[DEFAULT_CHANNEL_ID]);
+        appWindow.channel = this._channelsById[DEFAULT_CHANNEL_ID];
```
pjbroadbent: Seems this assignment happens within `wrapApplication` already?

Also, when wrapping an application is `channel` ever anything other than `DEFAULT_CHANNEL_ID`? Didn't see any instances, but the interface makes searching a little fiddly.

test/demo/utils/jestExtensions.ts
```diff
@@ -11,25 +19,31 @@ declare global {
...
              */
-            toThrowFDC3Error<R>(code: string, message?: string | RegExp): R;
+            toThrowFDC3Error(code: string, message?: string | RegExp): R;
 
             /**
              * Used to test that a RemoteChannel represents the expected client-side Channel
              * @param channel An object with optional `id` and `type` properties that if present we expect the channel to match, or the `ChannelId`
              * @param classType The class we expect the channel to be an instance of
              */
             toBeChannel(channel: {id?: ChannelId, type?: string} | ChannelId, classType?: Function): R;
```
pjbroadbent, line 29: Not sure I understand the generics here, is `R` always `CustomMatcherResult`, then?

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Nah, I agree `return false`. Changed

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Fixed

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Written like the, VSC thinks the type of creationTime is number. I can do, `const creationTime: number | undefined`, but the fact that I'd need to feels like a language-nudge to check that the property actually exists. I could do a Map<number> instead....

src/provider/index.ts
```diff
@@ -103,10 +103,12 @@ export class Main {
...
+            })).then(() => {});
```
tomrobertsOF: Rather than `return Promise.all(...).then(() => {})`, can we just await the `Promise.all` without the extra `then`?

src/provider/model/Model.ts
bryangaleopenfin: Changed

test/demo/utils/jestExtensions.ts
bryangaleopenfin: Not sure I do either, but built-in matchers will always return the type that gets passed into the the `expect`, and this continues that. I removed the `<R>` in this line as it seemed redundant

src/provider/controller/IntentHandler.ts
bryangaleopenfin: My logic here is that it's up to the app what data gets returned, and if we're always putting it in an array, the app loses the option of returning anything other than an array.

(Also, with the current client API, there's no way for the intent-raiser to get at this data, so it's kinda moot)

src/provider/index.ts
bryangaleopenfin: Possibly if the window didn't even import the FDC3 client, but then no-more-or-less than when this was only firing at a single window

bryangaleopenfin: And it's in the general tech-debt story to add a handshake, which would clear this up anyway

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Added todo

src/provider/model/FinEnvironment.ts
bryangaleopenfin: "windows know their own creation time" - like, from the OpenFin API?

It's done this way so it's the actual creation time, rather than (creation + however long it takes for `registerWindow` to get called), and unless there is an OpenFin API, it's only FinEnvironment that knows that, as it's the only place we can sensibly listen for the `window-created` event

I was very tempted, to get rid of wrapApplication, and have windowCreated dispatch an `AppWindow`, as I /think/ this would remove the need for the map, but thought that might be controversial 😛

src/provider/model/Model.ts
bryangaleopenfin: That's a point, I feel like I don't properly understand the difference...

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Actually, changing to Map

src/provider/model/Model.ts
bryangaleopenfin: Ah, that's not an openfin call, it just goes to `setAsForeground`

bryangaleopenfin: Changed so apps are focused in creation order, so this at least will deterministic

Ideal world - ZIndexer would be an importable model, and we'd be able to to bring all windows of this app to the front, while preserving their relative ordering, but this feels like the next best thing

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Hmm, we use "do this thing but with a timeout" in a bunch of places, and it's worth that being unified. Pushed a change, any better? 😬

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Data will not be returned. Left as future story

src/provider/index.ts
bryangaleopenfin: Added to tech-debt story to consider this specific case

src/provider/controller/IntentHandler.ts
bryangaleopenfin: ✅

src/provider/index.ts
bryangaleopenfin: ✅

test/demo/utils/jestExtensions.ts
bryangaleopenfin: ✅

src/provider/model/Model.ts
bryangaleopenfin: ✅

src/provider/model/FinEnvironment.ts
bryangaleopenfin: ✅

src/provider/model/FinEnvironment.ts
bryangaleopenfin: ✅

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Tidied - tried a version where `withTimeout` rejects, rather than returning a tuple, but this made `FinEnvironment.createApplication` messy (due to having to do some woo to not catch a possible `startFromManifest` error and mistakenly throw a timeout error), so stuck with `withTimeout` as is, but with some tidying to `isReadyToReceiveIntent`

✅

bryangaleopenfin:
Responded to all feedback and ready for re-review @pjbroadbent 

Controversial: We now depend on `wrapWindow` only being called once for each window, otherwise we don't get to delete from our `pendingWindows` map. This means `createApplication` returns `Promise<void>` rather than `Promise<AppWindow>`

pjbroadbent:
@bryangaleOF Regarding `wrapApplication` (I'm assuming you mean wrapApplication, rather than wrapWindow?) only being called once, that _should_ be the case anyway - weird things are going to happen if we have multiple window objects for the same window. Could be worth adding a comment to that function, to mention that caveat, though.

src/provider/index.ts
pjbroadbent: Can we tag this with the ticket number?

src/provider/model/FinEnvironment.ts
```diff
@@ -15,6 +15,23 @@ import {AppWindow} from './AppWindow';
 import {ContextChannel} from './ContextChannel';
 import {getId} from './Model';
 
+interface PendingWindow {
+    creationTime: number | undefined;
+    count: number;
```
pjbroadbent: `index`?..

src/provider/model/FinEnvironment.ts
```diff
@@ -43,73 +63,84 @@ export class FinEnvironment extends AsyncInit implements Environment {
...
-    [key: string]: boolean;
-}
-
-interface ChannelEventMap {
-    [channelId: string]: {[eventId: string]: boolean};
-}
-
 class FinAppWindow implements AppWindow {
+    public readonly appWindowNumber: number;
+    public channel: ContextChannel;
```
pjbroadbent, line 169: Can we check that we're being consistent here in terms of what's `public readonly`, vs what's `private` with a getter? Same for `public` vars vs private var plus getter+setter.

src/provider/index.ts
bryangaleopenfin: Added todo, but felt like this was the wrong place for it, so is in `main`

src/provider/model/FinEnvironment.ts
bryangaleopenfin: ✅

src/provider/model/FinEnvironment.ts
bryangaleopenfin: We're generally inconsistent, but I've changed to "private with getter" for consistency within this class at least

We have no setters anywhere in FDC3, so have left the public not-readonly as is

bryangaleopenfin:
Responded to all feedback and ready for re-review @pjbroadbent

pjbroadbent (approved):

tomrobertsOF (approved):

src/provider/index.ts
bryangaleopenfin: 👍

tomrobertsOF:
@bryangaleOF All looks good to me. Happy to merge once the conflicts are resolved.

PR #120, opened 2019-07-09
[SERVICE-542] Refresh FDC3 API docs

tim-dinsdale:
Documentation


src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
...
+/**
+ * Built-in context to define a financial instrument.
+ */
+export interface InstrumentContext extends Context {
+    /**
+     * The context type is always 'fdc3.instrument'.
+     */
+    type: 'fdc3.instrument';
+    /**
+     * Can contain some of a ticker, an [ISIN](https://www.isin.org/isin/), a [CUSIP](https://www.cusip.com/cusip/index.htm),
```
tomrobertsOF: "Can contain some of a ticker, ...." seems weirdly phrased, but maybe I just don't understand finance terms enough.

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
     id?: {[key: string]: string|undefined;};
```
tomrobertsOF: Not really about docs, but is ID really optional? It already allows basically anything as properties, but I would expect that some kind of identifier would be required?

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
...
     id: {[key: string]: string}&{email?: string; twitter?: string; phone?: string};
 }
-export interface SecurityContext extends Context {
-    type: 'fdc3.security';
-    id: {[key: string]: string}&{default: string};
+
+/**
+ * Built-in context to define a financial instrument.
+ */
+export interface InstrumentContext extends Context {
```
tomrobertsOF, line 65: Because this inherits from `Context` the generated docs have `name: undefined | string` in this interfaces page too. Does an instrument context take name in that form or do we want to explicitly say it's undefined in this case?

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
...
+     */
     type: 'fdc3.contact';
+    /**
+     * Name of the contact.
+     */
     name: string;
+    /**
+     * The contact data. Currently has optional fields for email address, Twitter handle, and phone number.
+     */
     id: {[key: string]: string}&{email?: string; twitter?: string; phone?: string};
```
tomrobertsOF, line 59: Again, not 100% about the docs, but this weird type intersection syntax generates docs with the definition `id: object & object` which look a bit weird. Is there a better way of handling this syntactically that would make the docs look a bit better?

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
...
+     */
+    id: {[key: string]: string}&{ticker?: string; ISIN?: string; CUSIP?: string; SEDOL?: string;
+        RIC?: string; BBG?: string; PERMID?: string; FIGI?: string;};
 }
+
+/**
+ * Built-in context to define an organization.
+ */
 export interface OrganizationContext extends Context {
+    /**
```
tomrobertsOF, line 93: As with InstrumentContext, leaving `name` out makes it inherit the comment from the parent. Want to be sure that's what we want.

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
+ * This type alias exists to disambiguate raw string app identity from the [[AppId]].
+ */
 export type AppName = string;
 
 /**
- * An application in the app directory
+ * An application in the app directory.
  */
 export interface Application {
+    /**
```
tomrobertsOF, line 33: These properties come out ordered alphabetically in the docs. Not sure if there is some way in typedoc to preserve the ordering of interfaces or if that's even something we want, but thought I'd raise it for discussion.

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
+     */
     icons?: Icon[];
+    /**
+     * Additional config. Currently unused by the OpenFin implementation.
+     */
     customConfig?: NameValuePair[];
-
-    intents?: Intent[];
+    /**
+     * The set of intents associated with this AppDir entry.
```
tomrobertsOF: There is a bit of inconsistency in the use of `AppDir entry` vs `app directory entry`. Can we decide which we want and use it consistently throughout.

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
+     * Images that can be displayed as part of the app directory entry. Use these for screenshots, previews or similar. These are not the
+     * application icons: use 'icons' for that.
+     */
     images?: AppImage[];
+    /**
+     * Contact email address.
+     */
     contactEmail?: string;
+    /**
+     * Support email address.
```
tomrobertsOF: This comment seems a bit unnecessary as it doesn't really give any more info than the property name. If we're going to comment this kind of thing can we add a bit more depth.

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
 }
 
-interface NameValuePair {
+/**
+ * The application allows extra configuration to be passed in, and uses an array of NameValuePairs,
+ * where the key is the name.
+ */
+export interface NameValuePair {
+    /**
+     * Key for the extra configuration.
```
tomrobertsOF: I think we'd be fine without these property comments since they're described in the interface comment anyway, but should probably discuss as a group whether we need a comment on every single thing.

src/client/errors.ts
```diff
@@ -2,32 +2,89 @@
...
+    /**
+     * Indicates that a window with a particular OF Identity cannot be found.
+     */
     WindowWithIdentityNotFound = 'WindowWithIdentityNotFound'
 }
 
+/**
+ * Class used to hold errors returned by the FDC3 provider.
+ */
 export class FDC3Error extends Error {
```
tomrobertsOF, line 90: We get some extra undocumented properties in the docs from this inheritance. Is there a way we can comment them?

src/client/main.ts
```diff
@@ -56,23 +69,50 @@ export interface AppIntent {
...
+/**
+ * Listener for context broadcasts. Generated by [[addContextListener]].
+ */
 export interface ContextListener {
+    /**
+     * The callback for when this listener receives a context broadcast.
+     */
     handler: (context: Context) => void;
     /**
      * Unsubscribe the listener object.
```
tomrobertsOF: This could maybe do with some elaboration about what exactly it means to unsubscribe

src/client/main.ts
```diff
@@ -81,9 +121,12 @@ export interface IntentListener {
 }
 
 /**
- * Defines all events that are fired by the service
+ * Defines all events that are fired by the service.
  */
 export type FDC3Event = FDC3MainEvent | FDC3ChannelEvent;
+/**
+ * Type to define the type of events fired by the service.
```
tomrobertsOF: "**Type** to define the **type**...." seems a bit awkward. Maybe "Alias for every valid eventType string." or something similar

src/provider/intents.ts
```diff
@@ -0,0 +1,32 @@
...
+ * General-purpose intent type
+ */
+export interface Intent<T=string, C extends Context=Context> {
+    type: T;
+    context: C;
+    target?: string;
+}
+
+// Built-in intents
+export type DialCallIntent = Intent<typeof Intents.DIAL_CALL, ContactContext>;
```
tomrobertsOF: Don't think you need the `typeof` here or the next line.

src/provider/model/FinEnvironment.ts
```diff
@@ -4,17 +4,19 @@ import {Identity, Window} from 'openfin/_v2/main';
...
 import {Timeouts} from '../constants';
+import {IntentType} from '../intents';
 import {parseIdentity} from '../../client/validation';
 
 import {Environment} from './Environment';
 import {AppWindow} from './AppWindow';
 import {ContextChannel} from './ContextChannel';
 import {getId} from './Model';
 
+
```
tomrobertsOF: Double line break

tomrobertsOF (commented):
Some comments to start things off.

src/provider/intents.ts
bryangaleopenfin: Did you suggest getting rid of these? Makes even more sense with these being exclusively in the provider

src/client/main.ts
```diff
@@ -280,12 +339,23 @@ export function addContextListener(handler: (context: Context) => void): Context
  */
 export function addEventListener(eventType: 'channel-changed', handler: (event: ChannelChangedEvent) => void): void;
 
+/**
+ * Generic event listener. This is not for intent or context subscription: use [[addIntentListener]] and [[addContextListener]], respectively.
```
bryangaleopenfin: For consistency, would put things in terms of what this does, rather than what it is
"Subscribes to an event type"

src/client/main.ts
```diff
@@ -280,12 +339,23 @@ export function addContextListener(handler: (context: Context) => void): Context
...
+ * @param identity The OF window identity. Currently unused.
+ */
 export function addEventListener(eventType: FDC3MainEventType, handler: (event: FDC3MainEvent) => void, identity?: Identity): void {
     validateEnvironment();
 
     eventEmitter.addListener(eventType, handler);
 }
 
+/**
+ * Remove any events you are subscribed to.
```
bryangaleopenfin: Makes this sound like a "remove all"
"Remove subscription from an event type"?

src/client/main.ts
```diff
@@ -280,12 +339,23 @@ export function addContextListener(handler: (context: Context) => void): Context
...
+ * @param eventType The type of the event to remove.
+ * @param handler The listener you had previously used to subscribe with.
```
bryangaleopenfin: For consistency with addIntentListener, handler rather than listener

src/client/main.ts
```diff
@@ -280,12 +339,23 @@ export function addContextListener(handler: (context: Context) => void): Context
  */
 export function addEventListener(eventType: 'channel-changed', handler: (event: ChannelChangedEvent) => void): void;
 
+/**
+ * Generic event listener. This is not for intent or context subscription: use [[addIntentListener]] and [[addContextListener]], respectively.
+ * @param eventType The event type.
+ * @param handler The callback when the event is fired.
```
bryangaleopenfin: For consistency with `addIntentListener`, "The handler to call when the event is fired."

src/client/main.ts
```diff
@@ -252,6 +308,9 @@ export function addIntentListener(intent: string, handler: (context: Context) =>
 
 /**
  * Adds a listener for incoming context broadcast from the Desktop Agent.
+ *
+ * To unsubscribe, use the returned [[ContextListener]].
+ * @param handler The callback function to call when we receive a broadcast context.
```
bryangaleopenfin: For consistency with addIntentListener, "The handler to call when we receive a broadcast context."

src/client/main.ts
```diff
@@ -210,15 +259,22 @@ export function broadcast(context: Context): void {
...
-export async function raiseIntent(intent: string, context: Context, target?: string): Promise<IntentResolution> {
+export async function raiseIntent(intent: string, context: Context, target?: AppName): Promise<IntentResolution> {
     return tryServiceDispatch(APIFromClientTopic.RAISE_INTENT, {intent, context: parseContext(context), target});
 }
 
 /**
  * Adds a listener for incoming Intents from the Agent.
+ *
+ * To unsubscribe, use the returned [[IntentListener]].
+ * @param intent The name of the intent to listen to.
```
bryangaleopenfin: "listen for"?

src/client/main.ts
```diff
@@ -252,6 +308,9 @@ export function addIntentListener(intent: string, handler: (context: Context) =>
 
 /**
  * Adds a listener for incoming context broadcast from the Desktop Agent.
```
bryangaleopenfin: Worth mentioning channels here? And standardize on "Desktop Agent" or "desktop agent"

src/client/main.ts
```diff
@@ -187,18 +234,20 @@ export async function findIntent(intent: string, context?: Context): Promise<App
...
  * ```
+ * @param context The context to send to find the intents that support it.
  */
 export async function findIntentsByContext(context: Context): Promise<AppIntent[]> {
     return tryServiceDispatch(APIFromClientTopic.FIND_INTENTS_BY_CONTEXT, {context: parseContext(context)});
 }
 
 /**
- * Publishes context to other apps on the desktop.
+ * Publishes context to other apps on the desktop. Any apps using [[addContextListener]] will receive this.
```
bryangaleopenfin, line 247: Deliberate choice to not mention channels? Second sentence feels like a simplification to the point of being wrong

src/client/main.ts
```diff
@@ -187,18 +234,20 @@ export async function findIntent(intent: string, context?: Context): Promise<App
  * // raise the intent, passing the given context, targeting the app
  * await agent.raiseIntent(startChat.intent.name, context, selectedApp.name);
  * ```
+ * @param context The context to send to find the intents that support it.
```
bryangaleopenfin: The 'to send' feels implicit in the fact this is a function call. "The context returned intents must support"?

src/client/main.ts
```diff
@@ -150,6 +195,8 @@ export async function open(name: string, context?: Context): Promise<void> {
  * // raise the intent against a particular app
  * await agent.raiseIntent(appIntent.intent.name, context, appIntent.apps[0].name);
  * ```
+ * @param intent The intent name to find.
+ * @param context An optional context to send to find the intent.
```
bryangaleopenfin, line 193: Huh, this actually gets completely ignored by the provider right now

src/client/main.ts
```diff
@@ -123,6 +166,8 @@ const contextListeners: ContextListener[] = [];
  *     //with context
  *     agent.open('myApp', context);
  * ```
+ * @param name The app name to launch.
+ * @param context A context to pass to the app post-'launch.
```
bryangaleopenfin: Stray '?

src/client/main.ts
bryangaleopenfin: Just "Event types fired by the service"?

src/client/main.ts
```diff
@@ -56,23 +69,50 @@ export interface AppIntent {
...
+ * Listener type alias. One of ContextListener or IntentListener.
+ */
 export type Listener = ContextListener | IntentListener;
 
+/**
+ * Listener for context broadcasts. Generated by [[addContextListener]].
+ */
 export interface ContextListener {
+    /**
+     * The callback for when this listener receives a context broadcast.
```
bryangaleopenfin: Ditto below about handler/callback/listener choice for word

src/client/main.ts
```diff
@@ -30,18 +30,31 @@ export * from './intents';
 export * from './errors';
 
 /**
- * Intent descriptor
+ * Contains the data you need to describe an intent.
```
bryangaleopenfin: "Describes an intent"? the rest feels redundant give that it is an interface

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
     id?: {[key: string]: string|undefined;};
 
     /**
+     * @hidden
      * A context object is open for extension with any custom properties/metadata.
```
bryangaleopenfin: Would reverse, so you start by describing the thing you're specifically commenting. "Any custom properties or metadata. This is used to allow for extension of the context object"

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
...
+    /**
+     * The context type is always 'fdc3.contact'.
+     */
     type: 'fdc3.contact';
+    /**
+     * Name of the contact.
+     */
     name: string;
+    /**
+     * The contact data. Currently has optional fields for email address, Twitter handle, and phone number.
```
bryangaleopenfin: Is `currently` needed? Might we remove these?

src/client/context.ts
bryangaleopenfin: "The instrument data. Can contain..." (Or change the comment for `id` on ContactContext to just "Can containemail address, Twitter handle, and phone number.")

src/client/context.ts
```diff
@@ -35,22 +36,60 @@ export interface Context {
...
+    /**
+     * The context type is always fdc3.organization.
+     */
     type: 'fdc3.organization';
-    id: {[key: string]: string}&{default: string};
+    /**
+     * Can contain either or both an [LEI](https://www.gleif.org/en/about-lei/introducing-the-legal-entity-identifier-lei)
+     * and a [PERMID](https://permid.org/).
+     */
+    id: {[key: string]: string}&{LEI?: string; PERMID?: string};
```
bryangaleopenfin, line 106: Similar feedback as for InstrumentContext id. Would make consistent with other FooContext id comments

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
 
 /**
- * Type definitions that is are wherever the API expects the ID or of an application to be passed. We have both, since
- * internally appId is used to identify app, but externally we use name to repsect specification
+ * Type alias to indicate when an Application Identifier should be passed. Application Identifiers
+ * are described [here](https://fdc3.finos.org/docs/1.0/appd-discovery#application-identifier).
  *
- * It is not possible for TypeScript to verify that only valid application names/appIds are passed
- * to the FDC3 API. This type definition is more a hint to the callee, over a plain "string" argument.
+ * This type alias exists to disambiguate raw string app identity from the [[AppName]].
```
bryangaleopenfin: I don't feel like this is true? Like, if we had 3 or 0 other AppFoo strings the sentence about this one still holds and we'd still want this

Though if we keep this, should it be: "This type alias exists to disambiguate *the* raw string app identity from the [[AppName]]."?

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
  */
 export type AppId = string;
+/**
+ * Type alias to indicate when an Application Name should be passed. This will be a human-readable
+ * name that will be displayed on screen.
+ *
+ * This type alias exists to disambiguate raw string app identity from the [[AppId]].
```
bryangaleopenfin: Same comment as for `AppId`. Though if we keep this line, should it be: "This type alias exists to disambiguate *the displayable app* name from the [[AppId]]."?

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
+     * The manifest type. Always 'openfin'.
+     */
     manifestType: string;
-
+    /**
+     * The version of the app. Please use [semantic versioning](https://en.wikipedia.org/wiki/Software_versioning).
+     */
     version?: string;
+    /**
+     * The human-readable title of the app, typically used by the launcher UI. If not provided, the name is used.
```
bryangaleopenfin: Still don't get the difference between name and title, though I assume this is comment ambiguity reflecting spec ambiguity?

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
+     */
     title?: string;
+    /**
+     * Tooltip used by any UIs that display app information. A short explanatory text string.
```
bryangaleopenfin: Would flip so it starts with the description of the thing "A short explanatory text string. For use in tooltips in any UIs that display app information."

src/client/directory.ts
bryangaleopenfin: Disagree that it's lose or expand. Sometimes there isn't that much to say, but for the docs we want *something*

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
+export interface Icon {
+    /**
+     * A URL that points to an icon.
+     */
     icon: string;
 }
 
-interface NameValuePair {
+/**
+ * The application allows extra configuration to be passed in, and uses an array of NameValuePairs,
```
bryangaleopenfin: Would try to rephrase for consistency, even if it sounds redundant. "A name value pair that something something"

src/client/directory.ts
bryangaleopenfin: Is the NVP fundamentally for configuation, or does that just happen to be it's use right now? I'm too wondering if we want to drop comments in this one, since meaning is so specific to where it's being used

src/client/directory.ts
```diff
@@ -10,60 +10,150 @@
...
     name: string;
+    /**
+     * Value for the extra configuration.
+     */
     value: string;
 }
-
-interface Intent {
+/**
+ * This is a representation of an [FDC3 Intent](https://fdc3.finos.org/docs/1.0/intents-intro) supported by the app in the app directory.
```
bryangaleopenfin: Would drop "This is"

src/client/errors.ts
```diff
@@ -2,32 +2,89 @@
  * @module Errors
  */
 
+/**
+  * Error codes returned by actions involving creating applications.
+  */
 export enum OpenError {
+    /**
+     * Indicates that an application is unable to be found in the AppDir.
```
bryangaleopenfin: Feel like "cannot be" scans better than "is unable to be"

src/client/errors.ts
```diff
@@ -2,32 +2,89 @@
...
+     */
     ErrorOnLaunch = 'ErrorOnLaunch',
+    /**
+     * Indicates that an application was not created in a timely manner, and the provider timed out.
+     */
     AppTimeout = 'AppTimeout',
+    /**
+     * Currently unused.
+     */
     ResolverUnavailable = 'ResolverUnavailable',
```
bryangaleopenfin, line 24: I think get rid of, or indicate what we expect this to be used for

src/client/errors.ts
```diff
@@ -2,32 +2,89 @@
...
 }
 
+/**
+ * Error codes returned by attempts firing intents to applications.
```
bryangaleopenfin: "attempts at" instead? "Error codes returned when attempting to fire an intents at an applications."?

src/client/errors.ts
```diff
@@ -2,32 +2,89 @@
...
     TargetAppDoesNotHandleIntent = 'TargetAppDoesNotHandleIntent',
+    /**
+     * Indicates that the provider has started an app to receive an intent, and that it has timed out waiting that app.
+     */
     IntentTimeout = 'IntentTimeout',
+    /**
+     * Currently unused.
+     */
     ResolverUnavailable = 'ResolverUnavailable',
+    /**
```
bryangaleopenfin, line 53: Same comment as for other ResolverUnavailable

src/client/errors.ts
```diff
@@ -2,32 +2,89 @@
...
+     */
     ResolverClosedOrCancelled = 'ResolverClosedOrCancelled',
 }
 
+/**
+ * Error codes returned by the context channel system.
+ */
 export enum ChannelError {
+    /**
+     * Indicates that [[getChannelById]] has failed because no such channel exists to broadcast on.
```
bryangaleopenfin: Applies to other comments as well - do we want to specific particular calls in error descriptions? We'll use this error code if we ever add any other call that takes a string channel ID

bryangaleopenfin (commented):
Lots of comments, but mostly wanting to squish inconsistencies in vocabulary

src/client/context.ts
tim-dinsdale: Fair, trying to clean this up.

src/client/context.ts
tim-dinsdale: So the way the contexts work today is that the base is optional but the derived is required. My typescript is not really strong enough to say whether this is a good thing or not

src/client/context.ts
tim-dinsdale: I think this is correct, because we require the well-understood id fields to be filled in. The name isn't that much use.

src/client/context.ts
tim-dinsdale: I have no idea about this one, sorry

src/client/context.ts
tim-dinsdale: Again, I think this is correct. Better to have the LEI or whatever than just free text name strings, so we should be relying on them.

src/client/directory.ts
tim-dinsdale: Yeah fair. Looks like you [can't change ordering anyway.](https://github.com/TypeStrong/typedoc/issues/112)

src/client/directory.ts
tim-dinsdale: :+1:

src/client/directory.ts
tim-dinsdale: I reckon I can be a bit less lazy though

src/client/directory.ts
tim-dinsdale: I'd actually rather lose these too. I put them to in to catch 'em all. I'll delete them

src/client/errors.ts
tim-dinsdale: No we can't, but I will put a note in saying what the base class is. That should be enough.

src/provider/intents.ts
tim-dinsdale: Maybe but this is just a cut and paste from elsewhere, so I am not sure I want to change it.

src/provider/model/FinEnvironment.ts
tim-dinsdale: dang it

src/provider/intents.ts
tim-dinsdale: Phil and I talked. The problem is that, like contexts, we provide a couple of starter ones to get you going (here)[https://fdc3.finos.org/docs/1.0/intents-spec], so we should probably have them in our API; so I left them here, and moved everything else into the provider.

src/client/main.ts
tim-dinsdale: :+1:

src/client/main.ts
tim-dinsdale: :+1:

src/client/main.ts
tim-dinsdale: :+1: I improved it somewhat.

src/client/main.ts
tim-dinsdale: :+1:

src/client/main.ts
tim-dinsdale: 👍

src/client/context.ts
tomrobertsOF: What I'm saying is, can we make the doc comment for name reflect the meaning of name for this specific derived type instead of just having it re-print the base type's comment, or if name is not a valid property for an InstrumentContext can we make it clear that it should be undefined for the derived type.

src/client/context.ts
tomrobertsOF: Again, more about the wording of the doc comment on the derived type than the existence of the name property itself.

src/client/directory.ts
tomrobertsOF: Ok, as long as we're happy with the look.

src/client/directory.ts
tomrobertsOF: @bryangaleOF, I would say a property called "supportEmail" is pretty self-descriptive and adding a comment saying "Support Email" doesn't really add anything of any value. If we're going to have doc comments on every property they should at least form a full sentence, even it if doesn't say much more in actuality.

src/client/errors.ts
tomrobertsOF: 👍

src/client/main.ts
tim-dinsdale: Yes. This is the FDC3 1.0 standard, which doesn't include them.

src/client/main.ts
tim-dinsdale: 👍

src/client/main.ts
tim-dinsdale: it's a glottal stop! Good catch, fixing

src/client/context.ts
pjbroadbent: Should actually be possible to merge them both (?..)

```
id: {
    email?: string;
    twitter?: string;
    phone?: string;
    [key: string]: string;
};
```
----
For some reason I can't reply to this. So instead I editted it.

Property 'email' of type 'string | undefined' is not assignable to string index type 'string'

is what you get for that

src/client/contextChannels.ts
```diff
@@ -335,7 +335,7 @@ export class DesktopChannel extends ChannelBase {
     public readonly color: number;
 
     /**
-     * @hidden
+     * @hideconstructor
```
pjbroadbent: Seems typedoc doesn't support this tag - I still see the constructor in the output.

src/client/main.ts
```diff
@@ -187,18 +234,20 @@ export async function findIntent(intent: string, context?: Context): Promise<App
  * // raise the intent, passing the given context, targeting the app
```
pjbroadbent: Just above the fold, there's a code sample that can be better-formatted.

typedoc uses backticks to create code blocks:

> \```ts
> [{
> // JSON goes here
> }]
> \```

src/client/main.ts
```diff
@@ -280,12 +339,23 @@ export function addContextListener(handler: (context: Context) => void): Context
  */
 export function addEventListener(eventType: 'channel-changed', handler: (event: ChannelChangedEvent) => void): void;
 
+/**
+ * Generic event listener. This is not for intent or context subscription: use [[addIntentListener]] and [[addContextListener]], respectively.
+ * @param eventType The event type.
+ * @param handler The callback when the event is fired.
+ * @param identity The OF window identity. Currently unused.
```
pjbroadbent: Are we going to allow listening to events from other windows? If not (or if not any time soon), then IMO we should remove this param.

src/provider/intents.ts
```diff
@@ -0,0 +1,32 @@
+import {Context, ContactContext} from '../client/context';
+import {Intents} from '../client/intents';
+
+/**
+ * TypeScript definitions for FDC3-defined intents.
+ *
+ * These structures are defined by the Intents FDC3 working group. The definitions here are based on current
+ * proposals and are not final. There may be minor differences between the current spec and the definitions here,
+ * in order to support this demo.
+ */
```
pjbroadbent: Can we remove this? It's a leftover from the initial prototype - specs are now finalised.

There's also a similar comment at the top of `main.ts`, and possibly elsewhere too.

src/provider/intents.ts
pjbroadbent: Given this is now in the provider, these sample intents aren't of any use to users of the service, though.

The provider itself only passes-through intents, so doesn't need any real understanding of the different intent types.

I still feel it'd be good for consistency to keep these "built in" intent definitions within the client (assuming we fill-in the rest of the standard intent set), but agree that realistically they're not likely to be of any use...

--------

Edit because I can't reply. I disagree. We shouldn't ship types that have no use to anyone using our API. It does occur to me that I could make the intent strings more strongly typed and provide some constants. Would you like that?

src/client/main.ts
```diff
@@ -56,23 +69,50 @@ export interface AppIntent {
...
+    /**
+     * For future use. Right now always the string '1.0.0'.
+     */
     version: string;
 }
 
+/**
+ * Listener type alias. One of ContextListener or IntentListener.
+ */
 export type Listener = ContextListener | IntentListener;
```
pjbroadbent: There's also a `ChannelContextListener` in `contextChannels.ts`, but it's not exported.

It does get returned when you add a listener to a channel, but we type it as `ContextListener`. Not sure if we're deliberately hiding the channel listener type? @bryangaleOF any thoughts?

src/client/main.ts
```diff
@@ -280,12 +339,23 @@ export function addContextListener(handler: (context: Context) => void): Context
  */
```
pjbroadbent, line 369: In the fold, the event docs refer to `getChannel` - this is now called `getChannelById`. We should also make it a link. Also, not sure that fully makes sense any more?.. That was written when channels were identified by strings, rather than being full objects.

Should we explicitly mention that this captures events from all channels, to differentiate it from the events on `ChannelBase`?

src/client/main.ts
pjbroadbent: Most of these comments mention errors thrown by the function, but they're all typed as `Error` rather than `FDC3Error`. Should likely also link to the error type.

In contextChannels.ts and the other services, we use `@throws` tags to document the error cases, can we use the same here?

src/client/contextChannels.ts
```diff
@@ -335,7 +335,7 @@ export class DesktopChannel extends ChannelBase {
     public readonly color: number;
```
pjbroadbent: There's some broken `@link` tags elsewhere in this file - I think `Channel` needs to be `BaseChannel`.

src/client/context.ts
tim-dinsdale: You get 

`Property 'email' of type 'string | undefined' is not assignable to string index type 'string'.`

src/client/context.ts
tim-dinsdale: Aah! Yes that makes sense, fixing

src/client/context.ts
tim-dinsdale: Got it, fixing

src/client/main.ts
tim-dinsdale: :+1:

src/client/main.ts
tim-dinsdale: Trying to avoid channels - we'll need to go back and do them properly. Agreed on Desktop Agent vs desktop agent. I resolved to lower case, because otherwise I would need to capitalize Intent and Context and I kind of hate that. It looks like a Fighting Fantasy book. You see the Goblin. Will you Test Your Luck, or fight it?

src/client/main.ts
tim-dinsdale: 👍

src/client/context.ts
tim-dinsdale: 👍

src/client/context.ts
tim-dinsdale: 👍

src/client/directory.ts
tim-dinsdale: So I feel like explaining how typescript works should not be a feature of our docs. So I removed that. Happy to return it if that;s the consensus.

src/client/directory.ts
tim-dinsdale: So turns out I was completely wrong about this. I will update.

src/client/errors.ts
tim-dinsdale: Yeah that's fair

tim-dinsdale: ...although it's quite polite to be able to link the error to the call? Hmm I can;t decide

src/client/directory.ts
bryangaleopenfin: Ah, yep, totally agree on "let's not explain Typescript in our docs" 🙂

src/client/directory.ts
tim-dinsdale: done

src/client/directory.ts
tim-dinsdale: Yeah this is fair. I cut this down somewhat

src/client/errors.ts
tim-dinsdale: I presume this was written for future use; so I am not sure. Maybe I will resolve merge clashes and see if it's still there. Either way I am not sure I want to do that in this doc PR

src/client/errors.ts
tim-dinsdale: I cleaned this up

src/client/errors.ts
tim-dinsdale: ...although it's quite polite to be able to link the error to the call? Hmm I can;t decide

src/client/contextChannels.ts
tim-dinsdale: I don;t think you can hide constructors. At least this takes the warning away. Want me to remove completley?

src/client/main.ts
tim-dinsdale: I can actually only find one that needs an @throws; and I put that in

src/client/main.ts
tim-dinsdale: I think I will remove 'currently unused' params in another pr

tomrobertsOF (approved):
Approved pending conflict fixing

src/client/main.ts
bryangaleopenfin: Added SERVICE-578 for this

src/client/errors.ts
bryangaleopenfin: Will leave for you to decide and resolve 😛

src/client/directory.ts
tim-dinsdale: I had another go! And added another AppName. Oh I wish TS had strong type aliases

src/client/errors.ts
tim-dinsdale: OK I actually think I like 

1. having the functions that you can then jump back to and
2. making you lot keep the comments up to date 

:D

src/client/contextChannels.ts
tim-dinsdale: fixed

src/client/contextChannels.ts
tim-dinsdale: Done

src/client/main.ts
```diff
@@ -81,12 +121,20 @@ export interface IntentListener {
...
 export type FDC3EventType = FDC3MainEventType | FDC3ChannelEventType;
-
+/**
+ * Events that are dispatched from the top-level event listener, via [[addEventListener]].
+ */
 export type FDC3MainEvent = ChannelChangedEvent;
+/**
+ * Event types fired from the top-level event listener.
+ */
 export type FDC3MainEventType = FDC3MainEvent['type'];
```
pjbroadbent: Can we hide these (`FDC3Event`/`FDC3EventType`/`FDC3MainEvent`/`FDC3MainEventType`) - they don't appear in the rest of the docs*, and aren't useful to clients - they're mostly there to make our internal types easier.

*: Except for `removeEventListener`, but that should have overrides defined for each event - same as `addEventListener`. With those overrides, the lowest-level "catch all" definition doesn't appear in the docs.

src/client/contextChannels.ts
```diff
@@ -213,7 +213,7 @@ export abstract class ChannelBase {
      * channels that they aren't a member of.
      *
      * This broadcast will be received by all windows that are members of this channel, *except* for the window that
-     * makes the broadcast. This matches the behaviour of the top-level FDC3 `broadcast` function.
+     * makes the broadcast. This matches the behavior of the top-level FDC3 `broadcast` function.
      *
      * @param context The context to broadcast to all windows on this channel
      * @throws `TypeError`: If `context` is not a valid {@link Context}
```
pjbroadbent, line 260: This is maybe more "types" than "docs", but should `addContextListener` below return a `ChannelContextListener` rather than `ContextListener`?

Seems we may be deliberately ignoring that `ChannelContextListener` exists?.. It's non-exported, and doesn't appear in the `Listener` alias in `main.ts`? @bryangaleOF, is this intentional?..

src/client/main.ts
```diff
@@ -123,15 +171,17 @@ const contextListeners: ContextListener[] = [];
...
  */
-export async function open(name: string, context?: Context): Promise<void> {
+export async function open(name: AppName, context?: Context): Promise<void> {
     return tryServiceDispatch(APIFromClientTopic.OPEN, {name, context: context && parseContext(context)});
 }
 
 /**
  * Find out more information about a particular intent by passing its name, and optionally its context.
  *
- * findIntent is effectively granting programmatic access to the Desktop Agent's resolver.
```
pjbroadbent, line 115: Can we use use backtick's around the `findIntent` here?

And in the code example just below, we should use `"fdc3.StartChat"` over `"StartChat"`. The latter doesn't match the FDC3 conventions for an intent name.

src/client/main.ts
```diff
@@ -178,6 +242,7 @@ export async function findIntent(intent: string, context?: Context): Promise<App
  * //     apps: [{ name: "Skype" }, { name: "Symphony" }, { name: "Slack" }]
  * // }];
```
pjbroadbent: This example is here twice - can we remove this non-formatted version?

There's also a "nested" code sample within `findIntent` - can we use the same structure there?

Not sure if this helps or just makes things harder to find, but might make more sense to have the code samples here on the docs for each function, and the JSON example as part of the `AppIntent` docs? (With a `See [[AppIntent]] for a sample response` comment?)

src/client/main.ts
```diff
@@ -123,15 +171,17 @@ const contextListeners: ContextListener[] = [];
  *     //with context
```
pjbroadbent: In the description above, can we link to the referenced types?

Slight caveat on the description - this isn't quite the same as opening then broadcasting, as that would send the context to all apps rather than just the app named by the `name` argument.

src/client/main.ts
```diff
@@ -210,15 +277,23 @@ export function broadcast(context: Context): void {
  * //raise an intent to start a chat with a given contact
  * const intentR = await agent.findIntent("StartChat", context);
  * //use the IntentResolution object to target the same chat app with a new context
- * agent.raiseIntent("StartChat", newContext, intentR.source);
+ * agent.raiseIntent("StartChat", newContext, intentR.apps[0].name);
  * ```
+ * @param intent The intent name to raise.
+ * @param context The context that will be sent with this intent.
+ * @param target An optional [[AppName]] to send the intent to.
+ * @throws [[FDC3Error]]`
```
pjbroadbent: There's an extra backtick here.

Can we mention the cases that cause the API to throw? Or at least, link to the relevant enum, since the reasons can be found there.

src/client/main.ts
```diff
@@ -210,15 +277,23 @@ export function broadcast(context: Context): void {
  * //raise an intent to start a chat with a given contact
```
pjbroadbent: Can we update this comment to match the example below - we're not raising an intent here.

Probably also where the `source` below came from.

src/client/directory.ts
```diff
@@ -1,69 +1,157 @@
...
+     * An application manifest, used to launch the app. This should be a URL that points to an OpenFin json manifest.
+     */
     manifest: string;
+    /**
+     * The manifest type. Always 'openfin'.
+     */
     manifestType: string;
-
+    /**
+     * The version of the app. Please use [semantic versioning](https://en.wikipedia.org/wiki/Software_versioning).
```
pjbroadbent: https://semver.org ?

src/client/main.ts
```diff
@@ -56,23 +69,50 @@ export interface AppIntent {
  * ```
```
pjbroadbent, line 70: In the description above, does this description make sense? We don't use the terms "Chain" or "Client-Service" anywhere else in the docs, and there's also no real distinction within the service between intents that do and don't return data.

src/client/contextChannels.ts
bryangaleopenfin: It definitely looks intentional!

I think this was the API we decided on, but `ChannelContextListener` was needed internally

src/client/main.ts
tim-dinsdale: :+1:

src/client/main.ts
tim-dinsdale: Done and done

src/client/main.ts
tim-dinsdale: Whoops and done. I like in-line examples, it's easier to see the whole story when you are reading the one functions.

src/client/main.ts
tim-dinsdale: Done and I think done

src/client/main.ts
tim-dinsdale: I... have no idea where this came from and agree

src/client/directory.ts
tim-dinsdale: Works for me!

src/client/main.ts
tim-dinsdale: Can't figure this one out, sorry

src/client/main.ts
tim-dinsdale: Done.

src/client/main.ts
bryangaleopenfin: Flagging that this is still inconsistent -
The comment on the event specific addEventListener, the generic addEventListener, and the generic removeListener, all take very different approaches. Suggest at least start each with a one sentence of what it actually does, so "Subscribes to a particular event type", "Subscribes to the channel changed event", and "Unsubscribes from a particular event type"

src/client/main.ts
pjbroadbent: IMO, it should be any function that can throw an FDC3 error - which I think is all of them?..

The way we document will depend on the final structure of the error codes though, can look at as part of SERVICE-561.

src/client/context.ts
pjbroadbent: Been playing around with this some more. Not possible to combine optional properties with `[key: string]: string`, and cannot list individual `id` fields within the generated TypeScript docs without creating a separate type, which seems a bit bloat-y.

IMO, rely on listing the individual fields within the `id` comment (as currently), and on IDE intellisense.

src/provider/intents.ts
pjbroadbent: IMO, we provide definitions for the "default" contexts, so makes sense to do it for intents too.

But it's true that these types are less useful, since `raiseIntent` (etc) takes each individual field rather than an intent object.

There's no point in defining these within the provider, as the provider has no need to know the shape of specific intents or contexts. They should either be in the client (if we want to make them part of the "official" API), or in the demo app (if we don't).

src/client/main.ts
pjbroadbent: @bryangaleOF Have updated comments

src/client/main.ts
pjbroadbent: I've removed this arg. Couldn't see any other "currently unused" params.

src/client/main.ts
pjbroadbent: Made the above change

src/client/main.ts
pjbroadbent: Updated. The comments within the code snippet didn't match-up with the actual code.

Now raises an intent using the resolver, then raises a targeted intent.

src/provider/intents.ts
bryangaleopenfin: It sounds like you're saying you'll take these out but these are still in

src/client/main.ts
bryangaleopenfin: This doesn't seem to have changed, or is at least still inconsistent. Discuss when you're in?

src/client/main.ts
bryangaleopenfin: @pjbroadbent Still inconsistent whether we're calling it a "callback" or a "handler" in two very similar methods

src/client/contextChannels.ts
bryangaleopenfin: @pjbroadbent Happy to resolve this? I think we had this conversation and `ChannelContextListener ` is now exported

And not docs, but is it an issue that this method returns a promise, where other add*Listeners, also with handshakes, return synchronously?

bryangaleopenfin:
@pjbroadbent  Taken another look, still outstanding comments

src/client/main.ts
bryangaleopenfin: @pjbroadbent  Still a couple of uses of callback in `main.ts` and `contextChannel.ts`

src/client/contextChannels.ts
pjbroadbent: https://appoji.jira.com/browse/SERVICE-791 covers the latter change about making the listeners async.

package.json
```diff
@@ -45,6 +45,7 @@
     "hadouken-js-adapter": "^1.44.1",
     "inversify": "^5.0.1",
     "inversify-inject-decorators": "^3.1.0",
+    "jest": "^24.9.0",
```
bryangaleopenfin, line 48: How come we're doing this?

src/provider/intents.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+/**
+ * Type definition that is used wherever the API expects the name of an intent to be passed.
+ *
+ * Since applications are free to define their own intents, it is not possible for TypeScript to verify that only valid
+ * intent names are passed to the FDC3 API. This type definition is more a hint to the callee, over a plain "string"
+ * argument.
+ */
+export type IntentType = (keyof typeof Intents)|string;
```
bryangaleopenfin, line 20: Huh, do we still not have a linting rule for spacing around ` | `?

src/client/context.ts
```diff
@@ -1,29 +1,26 @@
-/**
- * @module Contexts
- */
-
 /**
  * TypeScript definitions for envelope and context objects.
```
bryangaleopenfin: Envelope?

src/client/context.ts
```diff
@@ -1,29 +1,26 @@
...
  *
- * These structures are defined by the Contexts FDC3 working group. he definitions here are based on current
- * proposals and are not final. There may be minor differences between the current spec and the definitions here,
- * in order to support this demo.
+ * These structures are defined by the Contexts FDC3 working group. This contains the Context interface for you to create your own
+ * contexts, as well as a set of standard contexts agreed by the WG.
```
bryangaleopenfin: Wouldn't abbreviate working group to WG, (or would capitalize "working group" earlier)

src/client/context.ts
```diff
@@ -1,29 +1,26 @@
...
+ * @module Contexts
  */
 
 /**
- * General-purpose context type
- *
- * A context object is open for extension with any custom properties/metadata.
+ * General-purpose context type, as defined by [FDC3](https://fdc3.finos.org/docs/1.0/context-intro).
+ * A context object is a well-understood datum that is streamable between FDC3 participants. As a result
```
bryangaleopenfin, line 12: Is "streamable" the right term here?

src/client/context.ts
```diff
@@ -35,27 +32,76 @@ export interface Context {
     id?: {[key: string]: string|undefined};
```
bryangaleopenfin, line 32: What's the line breaking rule here? Comment could be one line and still shorter than other comment lines in this file

src/client/errors.ts
```diff
@@ -2,33 +2,112 @@
  * @module Errors
  */
 
+/**
+ * Error codes returned by actions involving creating applications.
```
bryangaleopenfin, line 6: "Created", "started", or "launched"? Inconsistent with block following

src/client/errors.ts
```diff
@@ -2,33 +2,112 @@
...
+     * Indicates that, in the case when a 'target' argument is passed to [[raiseIntent]], no
+     * such app either exists in the application directory or is currently running.
+     */
     TargetAppNotAvailable = 'TargetAppNotAvailable',
+    /**
+     * Indicates that, in the case when a 'target' argument is passed to [[raiseIntent]], the app is not able to handle this intent.
+     */
     TargetAppDoesNotHandleIntent = 'TargetAppDoesNotHandleIntent',
+    /**
+     * Indicates that the provider has started an app to receive an intent, and that it has timed out waiting that app.
```
bryangaleopenfin: "waiting for that app to handle the intent"

src/client/errors.ts
```diff
@@ -2,33 +2,112 @@
...
+     */
     ResolverClosedOrCancelled = 'ResolverClosedOrCancelled',
 }
 
+/**
+ * Error codes returned by the context channel system.
+ */
 export enum ChannelError {
+    /**
+     * Indicates that [[getChannelById]] has failed because no such channel exists to broadcast on.
```
bryangaleopenfin: Would drop "to broadcast on", and possibly replace with "with the given ID"

src/client/errors.ts
```diff
@@ -2,33 +2,112 @@
...
+ * [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) type.
+ *
+ * Note that not all errors raised by the service will be of type `FDC3Error`. Standard JavaScript error types such as
+ * `TypeError` and `Error` can also be thrown by the API.
+ */
 export class FDC3Error extends Error {
+    /**
+     * A string from one of [[OpenError]], [[ResolveError]], [[ChannelError]] or [[IdentityError]].
+     *
+     * Future versions of the service may add additional error codes, applications should allow for the possibility of
```
bryangaleopenfin: Change ", " to ". ", or " so" so it's definitely either one sentence or two.

src/client/errors.ts
```diff
@@ -2,33 +2,112 @@
...
+    /**
+     * Always `'FDC3Error'`.
+     */
+    public name: string;
+
+    /**
+     * Description of the error that occurred.
+     *
+     * These messages are not intended to be user-friendly, we do not advise displaying them to end users. If
+     * error-specific user messaging is required, use [[code]] to determine which message should be displayed.
```
bryangaleopenfin: Think "what message" rather than "which message", as we're not giving a list of options for messaging.

src/client/directory.ts
```diff
@@ -1,60 +1,169 @@
...
+ * using [[Application.customConfig]].
+ *
+ * This type alias exists to disambiguate the raw string app identity from the [[AppName]].
  */
 export type AppId = string;
+/**
+ * App Name is the machine-readable name of the app, but it may well be sufficiently
+ * human-readable that it can be used in user interfaces. If it's not, please use the title.
+ *
+ * This type alias exists to disambiguate raw string app name from the [[AppId]].
```
bryangaleopenfin: "disambiguate the raw string app name"

bryangaleopenfin: (Also don't agree with the explanation of the type alias, but think that's a discussion we had about 10 years ago and I lost)

src/client/directory.ts
```diff
@@ -1,60 +1,169 @@
...
+     */
     appId: AppId;
+    /**
+     * The machine-readable app name, used to identify the application in various API calls to the application directory.
+     * This may well be human-readable, too. If it's not, you can provide a title, and that will be used everywhere
+     * a name needs to be rendered to a user.
+     */
     name: AppName;
+    /**
+     * An application manifest, used to launch the app. This should be a URL that points to an OpenFin json manifest.
```
bryangaleopenfin: Capitalize JSON

src/client/directory.ts
```diff
@@ -1,60 +1,169 @@
...
+    /**
+     * The version of the app. Please use [semantic versioning](https://semver.org/).
+     */
     version?: string;
+    /**
+     * The human-readable title of the app, typically used by the launcher UI. If not provided, [[name]] is used.
+     */
     title?: string;
+    /**
+     * A short explanatory text string. For use in tooltips used by any UIs that display app information.
```
bryangaleopenfin: "shown" rather than "used"?

src/client/directory.ts
```diff
@@ -1,60 +1,169 @@
...
+     */
     tooltip?: string;
+    /**
+     * Longer description of the app.
+     */
     description?: string;
+    /**
+     * Images that can be displayed as part of the app directory entry. Use these for screenshots, previews or similar. These are not the
+     * application icons: use [[icons]] for that.
```
bryangaleopenfin: Think this is " - " rather than ": "

src/client/directory.ts
```diff
@@ -1,60 +1,169 @@
...
+    /**
+     * Icons used in the app directory display. A launcher may be able to use various sizes.
+     */
     icons?: Icon[];
+    /**
+     * Additional config.
+     *
+     * The OpenFin FDC3 service supports the following configuration values:
+     * * `appUuid`: Informs the service that the application launched by this [[manifest]] will have this UUID. By
+     * default, the service will expect the UUID of the application to match the [[appId]], this configuration value
```
bryangaleopenfin: Think that last comma should be a full stop

src/client/contextChannels.ts
```diff
@@ -1,7 +1,21 @@
 /**
+ * Context channels allow end-user filtering of context broadcasts. Each window is assigned to a particular
+ * context channel, and any [[broadcast|broadcasts]] by any window in that channel will only be recevied by the other
+ * windows in that channel. The assignment of windows to channels would typically be managed by the user, through
+ * either a channel selector widget built into the window itself, or through a separate channel manager application.
+ *
+ * All windows will initially be placed in the [[defaultChannel|default channel]], and will remain so unless they
```
bryangaleopenfin: Think either get rid of "placed", or change "remain so" to "remain there"

src/client/contextChannels.ts
```diff
@@ -16,7 +30,7 @@ import {ContextListener} from './main';
 import {Transport} from './EventRouter';
 
 /**
- * Type used to identify specific Channels. Though simply an alias of `string`, use of this type indicates usage of the string
+ * Type used to identify specific Channels. Though simply an alias of `string`, use of this type indicates use of the string
  * as a channel identifier, and that the user should avoid assuming any internal structure and instead treat as a fully opaque object
  */
 export type ChannelId = string;
```
bryangaleopenfin, line 36: Change the two "e.g"s below to "e.g."?

bryangaleopenfin: And "within this channel" to "with this channel"

src/client/contextChannels.ts
```diff
@@ -47,12 +61,14 @@ export interface DisplayMetadata {
 export type Channel = DefaultChannel | SystemChannel | AppChannel;
 
 /**
- * Event fired when a window is added to a channel. See {@link Channel.addEventListener}.
+ * Event fired when a window is added to a channel. See {@link ChannelBase.addEventListener}.
  *
  * Note that this event will typically fire as part of a pair - since windows must always belong to a channel, a window
  * can only join a channel by leaving it's previous channel. The exceptions to this rule are when the window is created
```
bryangaleopenfin: "it's" -> "its"

src/client/contextChannels.ts
```diff
@@ -78,13 +94,13 @@ export interface ChannelWindowAddedEvent {
...
 /**
- * Event fired when a window is removed from a channel. See {@link Channel.addEventListener}.
+ * Event fired when a window is removed from a channel. See {@link ChannelBase.addEventListener}.
  *
  * Note that this event will typically fire as part of a pair - since windows must always belong to a channel, a window
  * can only join a channel by leaving it's previous channel. The exceptions to this rule are when the window is created
  * and destroyed when there will be no previous channel or no current channel, respectively.
  *
- * To listen for channel changes across all (or multiple) channel, there is also a top-level {@link ChannelChangedEvent}.
+ * To listen for channel changes across all (or multiple) channels, there is also a top-level {@link ChannelChangedEvent}.
```
bryangaleopenfin, line 103: Below, change "added to" to "removed from"

src/client/contextChannels.ts
```diff
@@ -158,8 +174,8 @@ export interface ChannelContextListener extends ContextListener {
 /**
  * Class representing a context channel. All interactions with a context channel happen through the methods on here.
  *
- * When users wish to generically handle both {@link SystemChannel}s and the {@link DefaultChannel}, generally the
- * {@link Channel} type should be used instead of {@link ChannelBase}.
+ * When users wish to generically handle both {@link SystemChannel}s, {@link AppChannel}s and the
+ * {@link DefaultChannel}, generally the {@link Channel} type should be used instead of {@link ChannelBase}.
```
bryangaleopenfin, line 178: Below, change "reset back into it's initial context-less state" to "reset back into *its* initial context-less state"

src/client/contextChannels.ts
```diff
@@ -286,49 +302,58 @@ export abstract class ChannelBase {
...
      *
      * The event also includes which channel the window is being added to. The `previousChannel` property within the
      * event will always be this channel instance.
      */
-    public async addEventListener(eventType: 'window-removed', listener: (event: ChannelWindowRemovedEvent) => void): Promise<void>;
+    public async addEventListener(eventType: 'window-removed', handler: (event: ChannelWindowRemovedEvent) => void): Promise<void>;
 
-    public async addEventListener<E extends ChannelEvents>(eventType: E['type'], listener: (event: E) => void): Promise<void> {
+    /**
+     * Subscribes to a particular event. This is not for subscribing to context updates on this channel, use [[addContextListener]] to receive broadcasts.
```
bryangaleopenfin: Split into two sentences - "Instead, use [[addContextListener]] to receive broadcasts."

src/client/contextChannels.ts
```diff
@@ -359,6 +388,8 @@ export class DefaultChannel extends ChannelBase {
  * This list of channels should be considered fixed by applications - the service will own the list of user channels,
  * making the same list of channels available to all applications, and this list will not change over the lifecycle of
  * the service.
+ *
+ * To fetch the list of available channels, use [[getSystemChannels]].
  */
 export class SystemChannel extends ChannelBase {
     public readonly type!: 'system';
```
bryangaleopenfin, line 395: Full stop at the end of the line below

src/client/main.ts
```diff
@@ -80,57 +134,63 @@ const intentListeners: IntentListener[] = [];
...
+export async function open(name: AppName, context?: Context): Promise<void> {
     return tryServiceDispatch(APIFromClientTopic.OPEN, {name, context: context && parseContext(context)});
 }
 
 /**
  * Find out more information about a particular intent by passing its name, and optionally its context.
  *
- * findIntent is effectively granting programmatic access to the Desktop Agent's resolver.
+ * `findIntent` is effectively granting programmatic access to the desktop agent's resolver.
  * A promise resolving to the intent, its metadata and metadata about the apps that registered it is returned.
```
bryangaleopenfin: "the apps registered to handle it", ditto for `findIntentsByContext`

src/client/main.ts
```diff
@@ -2,11 +2,14 @@
  * @module Index
```
bryangaleopenfin, line 2: Suggest a pass over the comments in example code in this file, varies between `//Comment`, `// comment`, and `//comment`

bryangaleopenfin (commented):
Sorry, spotted a bunch more things. Think the only ones I really care about are "it's" v. "its". Can live with the more subjective stuff being ignored

package.json
pjbroadbent: Fixes a build error in the tests that appeared after a merge. Happens due to various dependencies including two different versions of jest, meaning jest is installed somewhere in the dependency tree that the build/test tooling can't find.

Not entirely sure why it only just started now, but it's something we've seen before intermittently.

src/client/context.ts
pjbroadbent: serializable?..

src/client/context.ts
pjbroadbent: 120 chars is my usual cut-off point. Have put this onto one line.

src/client/errors.ts
pjbroadbent: Needs some clean-up, lets address as part of SERVICE-561

src/provider/intents.ts
pjbroadbent: 🤷‍♂ 

This matches what's currently in develop: https://github.com/HadoukenIO/fdc3-service/blob/develop/src/client/intents.ts#L52

src/client/errors.ts
```diff
@@ -64,7 +65,7 @@ export enum ResolveError {
  */
 export enum ChannelError {
     /**
-     * Indicates that [[getChannelById]] has failed because no such channel exists to broadcast on.
+     * Indicates that [[getChannelById]] has failed because no such channel exists to with the given ID.
```
bryangaleopenfin: "to with" -> "with"

src/client/context.ts
bryangaleopenfin: shareable?

bryangaleopenfin: Meh

bryangaleopenfin (approved):

PR #121, opened 2019-07-11
[N/A] .Net window support

bryangaleopenfin:
Adds untested support for .Net windows, and removes unwanted `fin` reference from `Model`

### DIFF
--- a/src/provider/model/FinEnvironment.ts
+++ b/src/provider/model/FinEnvironment.ts
@@ -1,6 +1,8 @@
 import {WindowEvent} from 'openfin/_v2/api/events/base';
-import {injectable} from 'inversify';
+import {injectable, id} from 'inversify';
 import {Identity, Window} from 'openfin/_v2/main';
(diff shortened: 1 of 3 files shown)


src/provider/model/FinEnvironment.ts
```diff
@@ -75,6 +112,18 @@ export class FinEnvironment extends AsyncInit implements Environment {
         const info = await fin.Application.wrapSync(identity).getInfo();
         this.windowCreated.emit(identity, info.manifestUrl);
     }
+
+    private async isExternalWindow(identity: Identity): Promise<boolean> {
+        const extendedIdentity = identity as (Identity & {entityType: string | undefined});
+
+        if (extendedIdentity.entityType) {
+            return extendedIdentity.entityType === 'external window';
```
lesjas: this should be 'external connection'

pjbroadbent (approved):

PR #122, opened 2019-07-15
[SERVICE-541] Add interim app directory customisation

rypete:
Adds the ability to set a custom application directory via the application manifest.


src/provider/model/AppDirectory.ts
```diff
@@ -1,25 +1,45 @@
...
 
 enum StorageKeys {
     URL = 'fdc3@url',
     APPLICATIONS = 'fdc3@applications'
 }
 
 // Demo development app directory url
-const devUrl = 'http://localhost:3923/provider/sample-app-directory.json';
-
+export const devAppDirectoryUrl = 'http://localhost:3923/provider/sample-app-directory.json';
```
bryangaleopenfin: Should we make this all caps to match other `export const`s?

src/provider/model/ConfigStore.ts
```diff
@@ -8,23 +8,27 @@ import {AsyncInit} from '../controller/AsyncInit';
...
 export class ConfigStore extends AsyncInit {
     private _store: Store<ConfigurationObject>;
-    private _loader: Loader<ConfigurationObject>;
+    private _loader!: Loader<ConfigurationObject>;
 
     constructor() {
         super();
         this._store = new Store(require('../../../gen/provider/config/defaults.json'));
-        this._loader = new Loader(this._store, 'fdc3');
+        if (global.hasOwnProperty('fin')) {
```
bryangaleopenfin: Rather than having these checks for `fin`, should we have a different `ConfigStore` that gets used in tests?

src/provider/common/Injector.ts
```diff
@@ -70,7 +70,7 @@ export class Injector {
             if (typeof Bindings[key] === 'function') {
                 container.bind(Inject[key]).to(Bindings[key] as any).inSingletonScope();
 
-                if ((Bindings[key] as Function).prototype.hasOwnProperty('init')) {
+                if ((Bindings[key] as Function).prototype.hasOwnProperty('init') && global.hasOwnProperty('fin')) {
```
bryangaleopenfin: Feel like I don't understand this? Only wait for dependency initialization if we're not in tests?

I feel like being able so set the bindings before this gets called, so this isn't necessary, might be tider

src/provider/model/AppDirectory.ts
```diff
@@ -1,25 +1,45 @@
...
-    public constructor() {
-        // Set default values
-        this._url = devUrl;
+    public constructor(@inject(Inject.CONFIG_STORE) configStore: ConfigStore) {
+        if (process.env.NODE_ENV === 'development') {
+            // Set the application directory to our local copy during development.  In production it will be an empty string.
+            configStore.config.add({level: 'desktop'}, {applicationDirectory: devAppDirectoryUrl});
+        }
+
+        this._url = configStore.config.query({level: 'desktop'}).applicationDirectory;
```
bryangaleopenfin: Should this be `this.updateURL`?

src/provider/model/AppDirectory.ts
```diff
@@ -1,25 +1,45 @@
...
+
+        // Adding the follow watch logic as there is a race condition with the loading of the application directory.
+        // This logic will set the URL in the store as soon as it is retrieved.
```
bryangaleopenfin: What's the race condition? Could we make sure configStore `initialize` doesn't resolve until these it's ready to have the `applicationDirectory` queried

bryangaleopenfin (commented):
Feel like there's a lot I don't understand. Grab me when you've got a good point to talk me through this

src/provider/common/Injector.ts
tomrobertsOF: I agree with Bryan on this front, but appreciate that we may need to just get this in. As a compromise I'd be fine with making and referencing a separate tech debt story to find a way to get this working properly.

src/provider/model/AppDirectory.ts
rypete: Yes, I like that much better.

src/provider/model/AppDirectory.ts
rypete: Yes, thanks

rypete: Fixed

src/provider/model/AppDirectory.ts
rypete: Fixed

rypete:
The outstanding open comments are all caused by injection issues.  WIP

src/provider/model/AppDirectory.ts
rypete: So the comment here was a bit wrong.  Its not really a race condition, but rather a mix of a feature and a need in testing.  Basically, if the desktop owner config changes after the initial load, this will pick up any change to the `applicationDirectory`.  This is more or a less of a feature to come.  TBD.  Following similar in testing, we manually set the `applicationDirectory` after the store is initialized so this is also needed to pick up that change. 

I have moved this to `AppDirectory.init` method and added an await for the `ConfigStore` there as it still makes sense to do that.

src/provider/model/ConfigStore.ts
rypete: Removed!

src/provider/common/Injector.ts
rypete: The `fin` guards have been removed and in place we have updated the injector to rebind certain injectables in the unit tests.  A primary example of such would be in the `AppDirectory.unittest.ts` file.

src/provider/common/DeferredPromise.ts
```diff
@@ -0,0 +1,26 @@
+export class DeferredPromise<T = void> {
```
ottah, line 1: Swap the references to `deferredPromise` in `utils/async.ts` to the class version here.

rypete: Fixed

src/provider/controller/ResolverHandler.ts
```diff
@@ -38,6 +38,11 @@ export interface ResolverResult {
     app: Application;
 }
 
+export interface ResolverHandlerBinding {
```
bryangaleopenfin, line 41: Suggest either:
- Have `ResolverHandler` implement this, and have `IntentHandler` take a `ResolverHandlerBinding` rather than a `ResolverHandler`

or:

- Get rid of `ResolverHandlerBinding`, changing any references back to `ResolverHandler`, and add an `as ReolverHandler` cast where necessary to `AppDirectory.unittest`

At the moment, we're relying on `IntentHandler` not complaining when it gets an implementation of `ResolverHandlerBindings` rather than an actual `ResolverHandler`. I feel like let's make that explicit

src/provider/model/ConfigStore.ts
```diff
@@ -5,18 +5,21 @@ import {injectable} from 'inversify';
 import {ConfigurationObject} from '../../../gen/provider/config/fdc3-config';
 import {AsyncInit} from '../controller/AsyncInit';
 
+export interface ConfigStoreBinding {
```
bryangaleopenfin, line 8: Similar comment as for `ResolverHandler`. I think if we're setting up the injector type mapping so that `Inject.CONFIG_STORE` maps to type `ConfigStoreBinding`, then `AppDirectory` should take a `ConfigStoreBinding` rather than a `ConfigStore`.

src/provider/common/Injector.ts
```diff
@@ -98,6 +113,9 @@ export class Injector {
      * @param type Identifier of the type/value to extract from the injector
      */
     public static get<K extends Keys>(type: typeof Inject[K]): Types[K] {
+        if (!Injector._container) {
```
bryangaleopenfin: Do we ever expect this to be null? `_container` will be initialized as soon as `Injector` is first imported, right?

src/provider/model/AppDirectory.ts
```diff
@@ -1,25 +1,53 @@
...
+        this._configStore = configStore;
+    }
+
+    protected async init(): Promise<void> {
+        await this._configStore.initialized;
+
+        if (process.env.NODE_ENV === 'development') {
+            // Set the application directory to our local copy during development.  In production it will be an empty string.
+            this._configStore.config.add({level: 'desktop'}, {applicationDirectory: DEV_APP_DIRECTORY_URL});
+        }
```
bryangaleopenfin, line 40: Just want to confirm that this is what we want to do? Rather than using the tools we now have to configure the URL from within the test

src/provider/common/Injector.ts
```diff
@@ -62,32 +62,47 @@ export class Injector {
...
+        });
+
+        Injector._initialized = Promise.all(promises).then(() => {});
+
+        return Injector._initialized;
     }
 
     public static rebind<K extends Keys>(type: typeof Inject[K]): inversify.BindingToSyntax<Types[K]> {
-        return Injector._container.rebind<Types[K]>(Bindings[type] as inversify.Newable<Types[K]>);
+        return Injector._container.rebind<Types[K]>(type as unknown as inversify.Newable<Types[K]>);
```
bryangaleopenfin: Just `...rebind<Types[K]>(type)` works here without the casting - guessing once all the types are stripped away `Injector._container.rebind` sees this param as a string? Looking at the calls to `Injector.rebind`, I don't see that this param would ever satisfy the `inversify.Newable<Types[K]>` interface

test/provider/AppDirectory.unittest.ts
```diff
@@ -136,7 +163,7 @@ describe('AppDirectory Unit Tests', () => {
             it('Return [] when fetching directory fails & URLs do not match', async () => {
                 const spyFetch = jest.spyOn(global, 'fetch')
                     .mockImplementation(() => Promise.reject(new Error('')));
-                const appDirectory = new AppDirectory();
+                const appDirectory = Injector.get<'APP_DIRECTORY'>(Inject.APP_DIRECTORY);
                 appDirectory['_url'] = '**different_url**';
```
bryangaleopenfin: Do we need to do this now? We can change the URL in the config, rather than reaching inside `appDirectory` and changing the URL. And similar comment for the "Fetches & updates the directory when URLs do not match" test

test/provider/AppDirectory.unittest.ts
```diff
@@ -106,7 +133,7 @@ describe('AppDirectory Unit Tests', () => {
             });
 
             it('Fetches & updates the directory when URLs do not match', async () => {
-                const appDirectory = new AppDirectory();
+                const appDirectory = Injector.get<'APP_DIRECTORY'>(Inject.APP_DIRECTORY);
                 mockJson.mockResolvedValue([fakeApp1]);
                 const spyGetItem = jest.spyOn(global.localStorage, 'getItem');
                 const spyFetch = jest.spyOn(global, 'fetch');
```
bryangaleopenfin: Here and below - `spyFetch` unused

test/provider/AppDirectory.unittest.ts
```diff
@@ -88,16 +96,35 @@ beforeEach(() => {
...
+        cancel: async (): Promise<void> => {}
+    });
+
+    await Injector.init();
+});
+
 describe('AppDirectory Unit Tests', () => {
     describe('When querying the Directory', () => {
         beforeEach(() => {
             mockJson.mockResolvedValue(fakeApps);
```
bryangaleopenfin: There's an inner beforeEach that also has this line. Are both needed?

Actually, could this move to `beforeEach` at the top of the file, and we get rid of all other `mockJson.mockResolvedValue(fakeApps);` lines in this file?

test/provider/AppDirectory.unittest.ts
```diff
@@ -119,15 +146,15 @@ describe('AppDirectory Unit Tests', () => {
 
             it('Use cached applications when the URLs are the same', async () => {
```
bryangaleopenfin, line 149: Is this test meaningfully different from the one below? Are we trying to test that `fetch` doesn't get called, because the cache can be use? Because fetch does get called in this test

test/provider/AppDirectory.unittest.ts
```diff
@@ -106,7 +133,7 @@ describe('AppDirectory Unit Tests', () => {
             });
 
             it('Fetches & updates the directory when URLs do not match', async () => {
-                const appDirectory = new AppDirectory();
+                const appDirectory = Injector.get<'APP_DIRECTORY'>(Inject.APP_DIRECTORY);
```
bryangaleopenfin, line 139: Do we definitely want to re-use instances of `AppDirectory` across tests? Particularly since we want to test caching behaviour, we could wind up in a situation where one test depends on a previous test having done *something* first.

Especially feels dangerous since we're directly setting the `_url` inside the `AppDirectory` in some tests. It works out ok because of the line `configStore.config.add({level: 'desktop'}, {applicationDirectory: DEV_APP_DIRECTORY_URL})`, but we're depending on an `onAdd` signal being fired even when the value hasn't changed.

test/provider/AppDirectory.unittest.ts
```diff
@@ -88,16 +96,35 @@ beforeEach(() => {
...
+
+    await Injector.init();
+});
+
 describe('AppDirectory Unit Tests', () => {
     describe('When querying the Directory', () => {
         beforeEach(() => {
             mockJson.mockResolvedValue(fakeApps);
+            const configStore = Injector.get<'CONFIG_STORE'>(Inject.CONFIG_STORE);
+            configStore.config.add({level: 'desktop'}, {applicationDirectory: DEV_APP_DIRECTORY_URL});
```
bryangaleopenfin: Suggest set this in the beforeAll, before AppDirectory is created, and then also have some tests specifically around this changing

src/provider/model/AppDirectory.ts
```diff
@@ -1,25 +1,53 @@
...
+
+        this.updateURL(this._configStore.config.query({level: 'desktop'}).applicationDirectory);
```
bryangaleopenfin: Change to `updateUrl()`? Think we're using that style more often

bryangaleopenfin (commented):
Looks good. Apologies I've leapt on the unit tests a bit, understand that aren't new with this PR, but think there's some stuff in there worth cleaning up.

Also, is this warning in when running `AppDirectory.unittest.ts` any problem?

    addWatch: Watch already exists within store, ignoring... MaskWatch {
      onAdd:
       Signal2 { _length: 2, _slots: [ [Object] ], aggregator: null },
      onRemove: Signal2 { _length: 2, _slots: [], aggregator: null },
      _store:
       Store {
         _items: Map { 'service' => [Array] },
         _cache: Map { 'desktop' => [Object] },
         _watches: [ [Circular] ] },
      type: 'mask',
      mask: { applicationDirectory: true } }

src/provider/model/AppDirectory.ts
bryangaleopenfin: No, wait, ignore!

bryangaleopenfin: No, wait, maybe don't ignore entirely. Is it possible to have a developer service config, so we don't need this check within the `AppDirectory` itself? (Though don't regard this as a blocker for merging!)

rypete: Yes and no.  This config is supposed to live in the desktop owners file which is yet to come and will be fully implemented in a future story (This story being just the groundwork.) We in theory could add it to say an app.json but then we are creating an additional work around.  This check here is a bit ugly but is only temporary and is similar to the other check in the same file which sets the appDirectory apps to be an empty array when in production, which is also temporary, until the mentioned desktop owners file story is in place.

test/provider/AppDirectory.unittest.ts
rypete: Yes they are both meaningful, but not in it's current written state.  In summary this tests description should be changed to something like `Fetches and updates directory when URLs are the same` and add some logic to change what is returned from the fetch.  The cache should only be used if the fetch fails which the second test covers.  

Speaking with @pjbroadbent, we are going to evaluate all the of the tests once the final story (Desktop owners file - SERVICE-516) is implemented as it will change a lot of the existing logic and we can do cleanup then.

test/provider/AppDirectory.unittest.ts
rypete: We can add this in as part of SERVICE-516 as we discussed

src/provider/controller/ResolverHandler.ts
rypete: Have gone with option 1

src/provider/model/ConfigStore.ts
rypete: Resolved with same as option 1 above

src/provider/common/Injector.ts
rypete: This was in error and has now been changed.  The check is to ensure that all of the modules that extend `asyncInit` are finished before the injector can be used.

src/provider/common/Injector.ts
rypete: fixed

test/provider/AppDirectory.unittest.ts
rypete: fixed

test/provider/AppDirectory.unittest.ts
rypete: removed

test/provider/AppDirectory.unittest.ts
rypete: done

test/provider/AppDirectory.unittest.ts
rypete: Will be fixed with SERVICE-516

src/provider/model/AppDirectory.ts
rypete: done

test/provider/AppDirectory.unittest.ts
pjbroadbent: This is a good point - didn't realise that we were re-using instances between tests.

@ryanpetersonOF I've got some changes that'll rebuild the injector at the start of each test, to prevent any re-use of instances. Let's review tomorrow, and can push up to your fork if all looks good.

src/provider/model/ConfigStore.ts
bryangaleopenfin: Cool. Could you also change `Main` in `provider/index.ts` to take a `ConfigStoreBinding` rather than a `ConfigStore`?

bryangaleopenfin:
Changes will look good. Will take a final look once Phil's changes are in

src/provider/model/ConfigStore.ts
rypete: Done

src/provider/model/AppDirectory.ts
```diff
@@ -1,25 +1,53 @@
...
+import {Scope} from 'openfin-service-config/Types';
+import {ScopedConfig} from 'openfin-service-config';
 
+import {Inject} from '../common/Injectables';
 import {Application, AppName} from '../../client/directory';
 import {AppIntent} from '../../client/main';
+import {AsyncInit} from '../controller/AsyncInit';
+import {ConfigurationObject} from '../../../gen/provider/config/fdc3-config';
+
+import {ConfigStore, ConfigStoreBinding} from './ConfigStore';
```
bryangaleopenfin: The config store import is now unused

rypete: Fixed

test/provider/AppDirectory.unittest.ts
pjbroadbent: My laptop is in a slightly dodgy state due to a broken update. Will bump this to SERVICE-516.

For what it's worth, tests still all passed with that change in there - so whilst current setup isn't ideal, doesn't seem to be creating any false-positives.

bryangaleopenfin (commented):
Looks good. One last thing - would put the `beforeAll` before the `beforeEach` to match execution order, then I'm good to approve

rypete:
@bryangaleOF I have moved the `beforeAll` before the `beforeEach`

bryangaleopenfin (approved):

ottah (approved):

PR #123, opened 2019-07-17
[SERVICE-551] Fix potential race condition in Model

tomrobertsOF:
Went with a similar approach to what we did in layouts for windows which close while registering. 

1. At the start of `onWindowCreated` we create a `deferredPromise` and add it to the private map referenced by the id of the window being created. 
2. The promise is resolved when `onWindowCreated` returns having either registered the window or determined that registration is not needed on creation.
3. In `onAPIConnection` we can await the created promise (if it exists), thereby ensuring that the create event will always have been handled before we proceed with the main handling of the APIConnection.

### DIFF
--- a/src/provider/model/Model.ts
+++ b/src/provider/model/Model.ts
@@ -133,29 +133,22 @@ export class Model {
     private async onWindowCreated(identity: Identity, manifestUrl: string): Promise<void> {
         // Registration is asynchronous and sensitive to race conditions. We use a deferred promise
         // to signal to other sensitive operations that it is safe to proceed.
-        const [pendingRegistration, clearPending] = deferredPromise();
+        const [pendingRegistration, resolvePending] = deferredPromise();
         this._pendingRegistrations.set(getId(identity), pendingRegistration);
 
         const apps = await this._directory.getAllApps();
         const appInfoFromDirectory = apps.find(app => app.manifest.startsWith(manifestUrl));
 
         const id: string = getId(identity);
-        if (!appInfoFromDirectory) {
-            // If the app is not in directory we ignore it. We'll add it to the model if and when it connects to FDC3
-            clearPending();
-            this._pendingRegistrations.delete(getId(identity));
-            return;
+        // If the app is not in directory we ignore it. We'll add it to the model if and when it connects to FDC3
+        if (appInfoFromDirectory && !this._windowsById[id]) {
+            this.registerWindow(appInfoFromDirectory, identity, true);
         } else if (this._windowsById[id]) {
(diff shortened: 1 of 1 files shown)


src/provider/model/Model.ts
```diff
@@ -128,21 +131,32 @@ export class Model {
...
+        // Registration is asynchronous and sensitive to race conditions. We use a deferred promise
+        // to signal to other sensitive operations that it is safe to proceed.
+        const [pendingRegistration, clearPending] = deferredPromise();
+        this._pendingRegistrations.set(getId(identity), pendingRegistration);
+
         const apps = await this._directory.getAllApps();
         const appInfoFromDirectory = apps.find(app => app.manifest.startsWith(manifestUrl));
 
+        const id: string = getId(identity);
         if (!appInfoFromDirectory) {
```
bryangaleopenfin: Could this block be a single `if` statement - `if (appInfoFromDirectory && !this._windowsById[id])`?

Or in any case, I'd suggest have each path fall through to the clearPending and delete at the bottom, rather than `return`ing. This is one of my "I can't explain this properly, but the graph in my head looks neater if it's done that way" things

src/provider/model/Model.ts
```diff
@@ -128,21 +131,32 @@ export class Model {
...
+            clearPending();
+            this._pendingRegistrations.delete(getId(identity));
             return;
+        } else {
+            this.registerWindow(appInfoFromDirectory, identity, true);
         }
 
-        this.registerWindow(appInfoFromDirectory, identity, true);
+        // Registration finished, allow any other sensitive operations to proceed
+        clearPending();
```
bryangaleopenfin: Suggest resolve rather than clear, to stick with vocab we already have for promises

bryangaleopenfin (approved):
LGTM. Couple of comments, but nothing that I feels like you need to be waiting on me to get back and re-review.

Think the final bullet point of the story is worth keeping captured in JIRA though, whether that it's own story or 551 kept open

src/provider/model/Model.ts
tomrobertsOF: I flipped back and forth a couple of times on that. I'll put it back to resolve

src/provider/model/Model.ts
tomrobertsOF: I thought I had done that, but apparently it got messed up in git somewhere.

tomrobertsOF: I was also keen to keep the "Ignoring window" comment as it helped with debugging.

rypete (approved):
This ones a real bread winner! Great job!

PR #125, opened 2019-07-19
[SERVICE-554] Ensure all handshakes are unshook if client disconnects

ottah:
Remove all listeners from an app window in the model on disconnect.

I've also added a member list/count to the channel selector in the demo to view what other apps are in a channel. Apps also get the last channel context.


test/demo/utils/ofPuppeteer.ts
```diff
@@ -115,6 +127,20 @@ export class OFPuppeteerBrowser {
         return identity;
     }
 
+    /**
+     * Execute a function on the provider and return the result.
+     * @param fn Function to execute.
+     * @param args Arguments to provided to the function.
+     */
+    public async executeOnProvider<A extends any[], R, C extends ProviderWindow = ProviderWindow>(fn: (this: C, ...args: A) => R, ...args: A):
```
tomrobertsOF: As discussed, can we keep these specific utils separate from the `OFPuppeteer` parts to make it easier to extract down the line.

ottah: Pulled out the stuff from this file, just changed the context generic so its not specific for client windows.

tomrobertsOF: Ok, ideally I would want to align this with the newer `OSPuppeter` stuff we have in notifications, but that can probably wait for now.

tomrobertsOF (approved):

src/provider/model/Model.ts
```diff
@@ -52,6 +52,7 @@ export class Model {
         this._environment.windowClosed.add(this.onWindowClosed, this);
 
         apiHandler.onConnection.add(this.onApiHandlerConnection, this);
+        apiHandler.onDisconnection.add(this.onWindowClosed, this);
```
pjbroadbent: Is this the best way to handle this case? We need to track that the window has lost some of it's context, but we may still want to persist some state.

For example, if a window refreshes or navigates to a new page, shouldn't it stay in the same channel? Removing the window from the model will place it back in the default channel again, which doesn't really feel right to me from a user perspective.

It may take some app integration for that to work nicely, but we should be considering those kind of use-cases (and edge cases) as part of this story.

test/demo/disconnect.inttest.ts
```diff
@@ -0,0 +1,170 @@
...
+            await quitApps(app);
+        });
+
+        testSuite('When app navigates away', testAppNotInDirectory1, open, async (app) => {
+            await navigateTo(app, 'about:blank');
+        });
+    });
+});
+
+async function testSuite(
```
pjbroadbent: Does `describe.each` work for this setup? Seems like the more idiomatic way of doing this kind of parameterisation.

src/demo/components/ContextChannelSelector/ChannelMemberView.tsx
```diff
@@ -0,0 +1,84 @@
...
+export function ContextChannelView(props: ChannelViewProps): React.ReactElement {
+    const channel = props.channel;
+    const [members, setMembers] = React.useState<Identity[]>([]);
+    const [color, setColor] = React.useState<number>(0xFFFFFF);
+    const [visible, setVisible] = React.useState(false);
+
+    const handleVisible = (event: React.MouseEvent) => {
+        setVisible(!visible);
+    };
+
```
bryangaleopenfin, line 20: Double line break here and at end of file (the linter update is in service tooling now 😛)

src/demo/components/ContextChannelSelector/ChannelMemberView.tsx
```diff
@@ -0,0 +1,84 @@
...
+                }
+            </ul>
+        </div>
+    );
+}
+
+/**
+ *
+ * @param num
+ */
```
bryangaleopenfin: Does this comment block add anything? Suggest remove

src/demo/apps/ChartsApp.tsx
```diff
@@ -26,6 +27,11 @@ export function ChartsApp(props: AppProps): React.ReactElement {
     }, []);
 
     React.useEffect(() => {
+        getCurrentChannel().then(async channel => {
```
bryangaleopenfin, line 30: This is to grab the new context when the window changes channel, right? I feel surprised that the context listener isn't triggered automatically. Oh, or is this for the case when the window contents get reloaded?

(Same comment for other demo apps)

src/demo/apps/ChartsApp.tsx
```diff
@@ -26,6 +27,11 @@ export function ChartsApp(props: AppProps): React.ReactElement {
...
+            const context = await channel.getCurrentContext();
+            if (context && context.type === 'security')
```
bryangaleopenfin: I don't think we're doing if's without braces elsewhere, would avoid introducing here

(Same comment for other demo apps)

src/provider/model/FinEnvironment.ts
```diff
@@ -22,17 +22,11 @@ interface PendingWindow {
...
 
-interface ContextMap {
-    [key: string]: boolean;
-}
+type ContextMap = Map<string, boolean>;
 
-interface ChannelEventMap {
-    [channelId: string]: {[eventId: string]: boolean};
-}
+type ChannelEventMap = Map<string, {[eventId: string]: boolean}>;
```
bryangaleopenfin: Can this be `{[eventId: FDC3EventType]: boolean}}` now we're using Maps? Was `string` before due to Typescript restrictions

src/provider/model/FinEnvironment.ts
```diff
@@ -22,17 +22,11 @@ interface PendingWindow {
...
 
-interface IntentMap {
-    [key: string]: boolean;
-}
+type IntentMap = Map<string, boolean>;
 
-interface ContextMap {
-    [key: string]: boolean;
-}
+type ContextMap = Map<string, boolean>;
```
bryangaleopenfin: Could this and the above `Map`s be `Set`s?

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Actually, suggest this be `Map<string, Map<FDC3EventType, boolean>>` (so we're not mixing an matching how we're mapping from a string to a boolean), or  `Map<string, Set<FDC3EventType>>`

src/provider/model/FinEnvironment.ts
```diff
@@ -221,45 +215,47 @@ class FinAppWindow implements AppWindow {
...
 
-        this._channelEventListeners[channel.id][eventType] = true;
+        this._channelEventListeners.get(channel.id)![eventType] = true;
     }
 
     public removeChannelEventListener(channel: ContextChannel, eventType: FDC3ChannelEventType): void {
-        if (this._channelEventListeners[channel.id]) {
-            delete this._channelEventListeners[channel.id][eventType];
+        if (this._channelEventListeners.has(channel.id)) {
+            const events = this._channelEventListeners.get(channel.id) || {};
```
bryangaleopenfin: We've already established in the above line that `this._channelEventListeners.get(channel.id)` should be defined, so would just write `const events = this._channelEventListeners.get(channel.id)!`

test/demo/disconnect.inttest.ts
```diff
@@ -0,0 +1,174 @@
...
+    await delay(500);
+}
+
+async function navigateTo(target: Identity, url: string): Promise<void> {
+    await ofBrowser.executeOnWindow(target, async function (location) {
+        const window = this.fin.Window.getCurrentSync();
+        await window.navigate(location);
+    }, url);
+
+    await delay(500); // wait for page reload
```
bryangaleopenfin: Suggest do something similar to notifications - `await delay(Duration.PAGE_RELOAD)`

test/demo/disconnect.inttest.ts
```diff
@@ -0,0 +1,174 @@
...
+type TestParam = [
+    string,
+    TestAppData,
+    (app: TestAppData) => Promise<void>,
+    (app: TestAppData) => Promise<void>
+];
+
+type TestCatagoryParam = [string, TestParam[]];
+
+async function openDirectoryApp(app: TestAppData) {
```
bryangaleopenfin: Suggest move to the bottom with other helper functions

test/demo/disconnect.inttest.ts
```diff
@@ -0,0 +1,174 @@
...
+            beforeEach(async () => {
+                await openMethod(application);
+                redChannel = await fdc3Remote.getChannelById(application, 'red');
+                blueChannel = await fdc3Remote.getChannelById(application, 'blue');
+                await fdc3Remote.addContextListener(application);
+                await fdc3Remote.addEventListener(application, 'channel-changed');
+                await fdc3Remote.addIntentListener(application, TEST_INTENT);
+                await redChannel.join(application);
+                await blueChannel.addContextListener();
+                await blueChannel.addEventListener('window-added');
```
bryangaleopenfin, line 101: Would divide these lines up with some whitespace for readability

test/demo/disconnect.inttest.ts
```diff
@@ -0,0 +1,174 @@
...
+
+describe('Disconnecting windows', () => {
+    // Directory Apps or NonDirectory Apps
+    describe.each(testCatagories)('%s Apps', (catagoryTitle: string, tests: TestParam[]) => {
+        // E.g. When an app is closed.
+        describe.each(tests)('When an app is %s', (
+            testTitle: string,
+            application: TestAppData,
+            openMethod: (app: TestAppData) => Promise<void>,
+            disconnectMethod: (app: TestAppData) => Promise<void>
```
bryangaleopenfin: Minor terminology quibble - method says to me "function on a class instance". Consider just going with `openFunction`/`disconnectFunction`?

src/provider/model/AppWindow.ts
```diff
@@ -48,4 +48,6 @@ export interface AppWindow {
     focus(): Promise<void>;
 
     isReadyToReceiveIntent(intent: IntentType): Promise<boolean>;
+
+    removeAllListeners(): void;
```
bryangaleopenfin, line 52: Does `test/mocks.ts` need to be updated to have this?

src/demo/apps/ChartsApp.tsx
ottah: The currentContext gets sent when you join a channel, and yes this is so when you reload/navigate-back on a demo app it gets the current context.

src/provider/model/FinEnvironment.ts
ottah: Went with `Map<string, Set<FDC3EventType>>` as it looks cleaner and since we don't ever use a `false` value, can just do `.has(eventType)`.

src/provider/model/FinEnvironment.ts
ottah: Yep

test/demo/disconnect.inttest.ts
ottah: Switched it

test/demo/disconnect.inttest.ts
bryangaleopenfin: Cool, though I think pull it out to a shared file. Notifs puts this in delay.ts

test/demo/disconnect.inttest.ts
```diff
@@ -0,0 +1,182 @@
...
+setupTeardown();
+
+export type ProviderWindow = Window & {
+    model: Model;
+    intentHandler: IntentHandler;
+    channelHandler: ChannelHandler;
+    eventHandler: EventHandler;
+}
+
+enum DURATION {
```
pjbroadbent: By convention, enums are named in CamelCase - can we rename this `Duration`? The enum keys should still be CONST_CASE.

src/provider/model/AppWindow.ts
ottah: Added to the mocks.

pjbroadbent (approved):

bryangaleopenfin (approved):

PR #127, opened 2019-07-23
[SERVICE-533] Make breaking tidying API changes

ottah:
- Rename `INTENT` & `CONTEXT` to prefix `RECEIVE-*` (Can discuss what this should be).
- Change payload `Context` -> `ContextPayload`

### DIFF
--- a/src/client/internal.ts
+++ b/src/client/internal.ts
@@ -56,9 +56,8 @@ export enum APIFromClientTopic {
  * Enum containing all and only actions that the client can accept.
  */
 export enum APIToClientTopic {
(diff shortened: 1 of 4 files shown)


pjbroadbent:
I'm not sure if the issues tagged in code were meant as an exhaustive list, or just a starting point.

We also need to figure out if we're actually ready to make these changes. There's some complex dependencies in place - let's discuss, and decide how best to proceed.

ottah:
> We also need to figure out if we're actually ready to make these changes. There's some complex dependencies in place - let's discuss, and decide how best to proceed.

I'll close this PR for now.

PR #128, opened 2019-07-23
[SERVICE-523] Add Signal package

ottah:
- Remove signal 
- Switch to signal package


tomrobertsOF (approved):

src/provider/APIHandler.ts
```diff
@@ -63,8 +63,8 @@ type APIActionReturn<T> = T | Readonly<T> | Promise<T> | Promise<Readonly<T>>;
  */
 @injectable()
 export class APIHandler<T extends Enum> {
-    public readonly onConnection: Signal1<Identity> = new Signal1<Identity>();
-    public readonly onDisconnection: Signal1<Identity> = new Signal1<Identity>();
+    public readonly onConnection = new Signal<[Identity]>();
+    public readonly onDisconnection = new Signal<[Identity]>();
```
pjbroadbent: Can we keep the type declarations? I know the end effect is the same, but for consistencies sake.

If a concrete type declaration is given, then (I think?) the generics aren't actually required on the constructor.

eg:
`public readonly onConnection: Signal<[Identity]> = new Signal();`

Ditto for other files.

rypete (approved):

src/provider/APIHandler.ts
ottah: Added them back. Going through and readding to the other PRs.

PR #130, opened 2019-08-19
[SERVICE-576][SERVICE-551] Gracefully handle newly-created windows that have not yet been registered

bryangaleopenfin:
This PR
* Adds an `expect` system to model, that generates a promise that either quickly resolves to a window in the model, or rejects
* Modifies `provider/index.ts` to use `expect` rather than `getWindow`

For the promise to resolve, the service must be aware of the window within 100ms of the call to `expect`, and window registration must complete within 5s of the service becoming aware of the window. This is a long timeout, but I believe necessary due to the app directory query that occurs as part of registration, which may be an HTTP call.

This PR also resolves the last unaddressed bullet point of SERVICE-551


src/provider/model/Model.ts
```diff
@@ -7,14 +7,24 @@ import {Inject} from '../common/Injectables';
...
+import {DeferredPromise, withStrictTimeout} from '../utils/async';
+import {Boxed} from '../utils/types';
 
 import {AppWindow} from './AppWindow';
 import {ContextChannel, DefaultContextChannel, DesktopContextChannel} from './ContextChannel';
 import {Environment} from './Environment';
 import {AppDirectory} from './AppDirectory';
 
+interface ExpectedWindow {
+    pendingDeferredPromise: DeferredPromise<Boxed<Promise<AppWindow>>>;
```
pjbroadbent: Can you explain the `Boxed` here? I'm guessing it's to prevent two promises from "collapsing" into each other, but not sure why that's necessary?..

test/provider/expect.unittest.ts
```diff
@@ -0,0 +1,340 @@
...
+    callTime: number,
+    finalizeTime: number,
+    result: 'resolve' | 'reject-timeout' | 'reject-closed'
+}
+
+type ExpectCallResult = {
+    promise: Promise<AppWindow>,
+    time: number,
+    call: ExpectCall
+};
```
pjbroadbent, line 40: The members in the type definitions above should be `;` delimited, not `,`.

The first two should also have a `;` at the end of the declaration.

src/provider/constants.ts
```diff
@@ -20,7 +20,7 @@ export const Timeouts = {
     /**
      * Time service allows for a window to go from 'pending' to being fully registered with the model
```
pjbroadbent: Update description, since this is no longer called "pending"?

Same for the naming of `WINDOW_EXPECT_TO_PENDING`, just to keep the terminology consistent.

src/provider/model/Model.ts
```diff
@@ -225,4 +243,53 @@ export class Model {
...
+                })
+            ]);
+
+            // Create a promise that resolves when the window has connected, or rejects if it is closed
+            const connected = Promise.race([
+                untilTrue((args) => {
+                    return !!args && getId(args[0]) === id;
+                }, this._apiHandler.onConnection),
+                untilTrue((args) => {
+                    return !!args && getId(args[0]) === id;
```
pjbroadbent: Can `untilTrue` spread it's input args rather than taking an array?

So that we can have named parameters here that match a listener on the signal, rather than `args[0]`.

src/provider/model/Model.ts
```diff
@@ -17,9 +17,14 @@ import {Environment} from './Environment';
 import {AppDirectory} from './AppDirectory';
 
 interface ExpectedWindow {
-    pendingDeferredPromise: DeferredPromise<Boxed<Promise<AppWindow>>>;
-    createdDeferredPromise: DeferredPromise<void>;
-    registerDeferredPromise: DeferredPromise<AppWindow>;
+    // Resolves when the window has been seen by the environment. Resolves to the `regsitered` promise wrapped in a timeout
```
ottah: regsitered -> registered

src/provider/model/Model.ts
```diff
@@ -277,17 +244,48 @@ export class Model {
...
+                })
+            ]);
+
+            // Create a promise that resolves when the window has connected, or rejects if it is closed
+            const connected = Promise.race([
+                untilTrue((args) => {
+                    return !!args && getId(args[0]) === id;
+                }, this._apiHandler.onConnection),
+                untilTrue((args) => {
+                    return !!args && getId(args[0]) === id;
```
ottah: Could you change the predicate signature in `untilTrue` to be:
```ts
predicate: (...args: A) => boolean
```
And then here we can explicitly list the args by name instead of using array indexing, makes it easier to read:
```ts
untilTrue((value: string) => {
	return !!value && getId(value) === id;
...
})
```

src/provider/model/Model.ts
```diff
@@ -277,17 +244,48 @@ export class Model {
...
+                    throw new Error(EXPECT_CLOSED_MESSAGE);
+                })
+            ]);
+
+            // Create a promise that resovles once the window has been seen, to a promise representing successful registration within our timeout
+            const seen = (windowSeen === 'seen' ? Promise.resolve() : untilTrue((args) => {
+                return args !== undefined && getId(args[0]) === id;
+            }, this._environment.windowSeen)).then(() => {
+                return {value: withStrictTimeout(Timeouts.WINDOW_SEEN_TO_REGISTERED, registered, EXPECT_TIMEOUT_MESSAGE).then(() => {
+                    return this._windowsById[id];
```
ottah: Hard to follow the logic here, maybe split this into 2 statements? 1 creating the promise/then and then the return
```ts
() => {
	const value = withStrictTimeout...
	return {value}
}
```

ottah (commented):
Looks good, still wrapping my head around some of the logic, just a few comments about making the code more readable.

src/provider/model/Model.ts
bryangaleopenfin: ✅

src/provider/model/Model.ts
bryangaleopenfin: Would prefer to leave as is - this means in all cases the predicate function can tell whether it's being called in response to the signal or not, and we're not relying on the signal payload not taking on certain values

src/provider/model/Model.ts
bryangaleopenfin: Changed this bit quite a lot. Should be easy to follow now

bryangaleopenfin:
@TheoOttahOF Ready for another look
@pjbroadbent I think the changes address most of your concerns, even though it's not going the route you suggested.

* Timeout/promise logic is kept with `expectWindow` and `onWindowCreated`.

* There's no longer any deferred promises, which were confusing things before, and instead we're piggybacking off existing signals.

* I'm still kicking off a timeout when the window is first seen, rather than storing a timestamp and using that to shorten the delay. I think it's preferable to have the one timer per window (for this stage of the timeout), than potentially multiple timers that we want to all fire simultaneously, but will probably only fire approximately simultaneously.

* I've also revised the registration logic so it's now really obvious what the flow is, since it's all inside `onWindowCreated`, and obvious there's no odd race condition that can lead to `registerWindow` being called twice

src/provider/model/Model.ts
```diff
@@ -131,73 +167,57 @@ export class Model {
...
-        if (appInfoFromDirectory && !this._windowsById[id]) {
+        if (appInfoFromDirectory) {
+            // If the app is in directory, we register it immediately
             this.registerWindow(appInfoFromDirectory, identity, true);
-        } else if (this._windowsById[id]) {
-            console.info(`Ignoring window created event for ${id} - window was already registered`);
-        }
+        } else {
+            // If the app is not in directory, we'll add it to the model if and when it connects to FDC3
+            allowReject(this.getOrCreateExpectedWindow(identity).connected.then(async () => {
```
pjbroadbent, line 183: Is the lack of `await` here intentional? Since the promises within `ExpectedWindow` don't include timeouts, `connected` could hang here, and this function basically ends up being a no-op. But I suppose in this case, that's ok?..

What is the `allowReject` trying to capture here? Seems the `connected` should never reject, only resolve or hang - since there's no timeout on it, and it's also defined as being an "`allowReject`" promise.

src/provider/model/Model.ts
```diff
@@ -228,4 +251,53 @@ export class Model {
...
+            });
+
+            // Create a promise that resolves when the window has registered
+            const registered = untilTrue(this._onWindowRegisteredInternal, (args) => {
+                return !!this._windowsById[id];
+            }).then(() => {
+                return this._windowsById[id];
+            });
+
+            // A promise that never resolves but rejects when the window has closed
```
pjbroadbent, line 278: `rejects if the window closes`?

What do we expect to the state of the `ExpectedWindow` if it's closed before we finish registering it? Seems a little convoluted that we reject the promise, but then suppress any rejections. Is there some useful side-effect of doing this over a `resolve()`?

src/provider/model/Model.ts
```diff
@@ -228,4 +251,53 @@ export class Model {
...
+            const closed = allowReject(untilTrue(this._environment.windowClosed, (args) => {
+                return !!args && getId(args[0]) === id;
+            }).then(() => {
+                throw new Error(EXPECT_CLOSED_MESSAGE);
```
pjbroadbent, line 282: Something I'm having a little trouble understanding... What do we expect to the state of the `ExpectedWindow` if it's closed before we finish registering it?

Seems a little convoluted to reject here, and then immediately suppress that rejection. Is there some useful side-effect of doing this over a `resolve()`?

May just have been staring at this PR a bit too long... :-)

src/provider/utils/async.ts
```diff
@@ -9,3 +13,48 @@ export function withTimeout<T>(timeoutMs: number, promise: Promise<T>): Promise<
...
     return Promise.race([timeout, p]);
 }
+
+/**
+ * Races a given promise against a timeout, and either resolves to the value the the promise resolved it, if it resolved before the
+ * timeout, or rejects
+ * @param timeoutMs Timeout period in ms
+ * @param promise Promise to race against the timeout
+ */
+export function withStrictTimeout<T>(timeoutMs: number, promise: Promise<T>, rejectMessage: string): Promise<T> {
```
pjbroadbent, line 23: Is there a convention behind when to use `withTimeout` and when to use `withStrictTimeout`?

This approach (to reject if the operation times-out) feels a bit more natural and idiomatic, any reason to not use this util throughout?

test/demo/utils/time.ts
```diff
@@ -0,0 +1,33 @@
+const MAX_PROMISE_CHAIN_LENGTH = 100;
+
+const realDateNow = Date.now;
+const boxedTime = {value: 0};
```
pjbroadbent: Why boxed?..

test/provider/expect.unittest.ts
```diff
@@ -0,0 +1,333 @@
...
+let mockApiHandler: jest.Mocked<APIHandler<APIFromClientTopic>>;
+
+beforeEach(async () => {
+    mockAppDirectory = new AppDirectory(null!) as jest.Mocked<AppDirectory>;
+    mockEnvironment = createMockEnvironmnent();
+    mockApiHandler = new APIHandler<APIFromClientTopic>() as jest.Mocked<APIHandler<APIFromClientTopic>>;
+    (mockApiHandler as PartiallyWritable<typeof mockApiHandler, 'onConnection'>).onConnection = new Signal<[Identity]>();
+    (mockApiHandler as PartiallyWritable<typeof mockApiHandler, 'onDisconnection'>).onDisconnection = new Signal<[Identity]>();
+
+    model = new Model(mockAppDirectory, mockEnvironment, mockApiHandler);
```
pjbroadbent, line 69: Can we use the injector to do this wiring?

eg: https://github.com/HadoukenIO/fdc3-service/blob/develop/test/provider/AppDirectory.unittest.ts#L83-L96

(Will also be simplified a fair bit with SERVICE-593)

test/provider/expect.unittest.ts
```diff
@@ -0,0 +1,333 @@
...
+                closeTime: 1000,
+                appType: 'directory'
+            },
+            3000,
+            [
+                [
+                    'When a window is expected shortly before it is created, the promise rejects',
+                    {callTime: 950, finalizeTime: 1000, result: 'reject-closed'}
+                ]
+            ]
```
pjbroadbent, line 123: Whilst the block above are pretty easy to follow, the tuple here is much more obfuscated.

Should these be objects, rather than tuples? So the keys of the object make it easier to see the test setup?

src/provider/constants.ts
bryangaleopenfin: ✅

test/provider/expect.unittest.ts
bryangaleopenfin: Oh man typescript is the *worst*

✅

src/provider/model/Model.ts
bryangaleopenfin: Note that predicate gets called when the `untilTrue` promise is first created. Can I have named params here without losing a 100% reliable test of whether `predicate` has been called due to a signal or not?

src/provider/model/Model.ts
bryangaleopenfin: My view is that it makes sense for `expectWindow` to reject if the window closes before the window finishes registering.

`allowReject` doesn't suppress rejections, it prevents rejections from causing warnings in tests and the console

src/provider/model/Model.ts
bryangaleopenfin: Semi duplicate of above comment?

src/provider/utils/async.ts
bryangaleopenfin: I tried getting rid of `withTimeout` a while ago, and it made the places that do use it messier, and didn't seem like an improvement

test/demo/utils/time.ts
bryangaleopenfin: Ah, leftover. Fixed

✅

test/provider/expect.unittest.ts
bryangaleopenfin: The difference between this block and the above blocks is the `TestWindow` not being declared inline with the `expectTest` call. I've fixed that, so things are consistent. Does that simplify things? The idea with this being a tuple is sticking to jest conventions for parameterized tests

src/provider/model/Model.ts
bryangaleopenfin: The lack of `await` here is intentional, but I don't *think* it'd make any difference in practice to have an await. I'm saying "in response to this event, set this thing up to happen", rather than "in response to this event, wait, possibly a very long time, for this to happen". Like, hypothetically, if `EventEmitter` await-ed on all listeners, would we want it to await until the window has connected? My feeling is no.

`connected` will reject if the window is closed, which makes sense, since we know it will never connect, and is consistent with the behaviour needed here to have `expectWindow` behave as expected

Note `allowReject` is around the promise returned by `then` (not `connected` itself), which I believe will itself reject if `connected` rejects

src/provider/model/Model.ts
bryangaleopenfin: The comment I've added indicates it's a promise that resolves to another promise. That that needs some boxing is a case of "that's just how typescript works". Does what I've put here suffice?

test/provider/expect.unittest.ts
bryangaleopenfin: It doesn't seem to work straightforwardly, and I could spend a lot of time trying to do in a dozen lines what I was doing in one...

src/provider/model/Model.ts
ottah: Could use the local variable `arguments` to check if theres any args passed.

Or prepend a flag argument so you can tell when its called by a signal or just the initial invocation?
```ts
predicate: (isSignal: boolean, ...args: A) => boolean
```

bryangaleopenfin: The former - there's still ambiguity if we've got a signal that fires no parameters

The latter, I gave that a try and couldn't get it to work. What would the function call look like in the "false" case?

`predicate(false, ???)`

What can go there that would compile?

pjbroadbent: Updated approach LGTM

src/provider/model/Model.ts
pjbroadbent: Yeah, answered above

test/provider/expect.unittest.ts
pjbroadbent: OK, yeah - maybe a bit verbose without having those changes first. Let's defer to SERVICE-593.

pjbroadbent (approved):

ottah (approved):

PR #134, opened 2019-09-09
[SERVICE-516] Allow App Directory location to be customised

rypete:
This is blocked by the runtime upgrades, but the work is done, so ignore all of the version changes in this PR (hence the draft PR).  This PR allows customizable application directory via the desktop owners file.


res/provider/config/fdc3-config.schema.json
```diff
@@ -132,7 +132,7 @@
             "properties": {
                 "applicationDirectory": {
                     "type": "string",
-                    "default": "",
+                    "default": "https://app-directory.openfin.co/api/v1/apps/",
```
bryangaleopenfin, line 135: Who in practice is populating this?

res/test/provider.json
```diff
@@ -12,5 +12,8 @@
     "runtime": {
         "arguments": "--remote-debugging-port=9222",
         "version": "13.76.43.31"
+    },
+    "serviceConfiguration": {
+        "applicationDirectory": "http://localhost:3923/provider/sample-app-directory.json"
```
bryangaleopenfin, line 17: Should this be `test/sample-app-directory.json`? Or is there behind the scenes magic happening?

src/provider/model/FinEnvironment.ts
```diff
@@ -133,14 +133,16 @@ export class FinEnvironment extends AsyncInit implements Environment {
             this.windowClosed.emit(identity);
         });
 
-        // Register windows that were running before launching the FDC3 service
-        const windowInfo = await fin.System.getAllWindows();
+        Injector.initialized.then(async () => {
```
bryangaleopenfin, line 142: Can this be just `await Injector.initialized`

bryangaleopenfin: And why do we need to wait for injector to be initialized? To give Model a chance to attach its signal listener first?

res/demo/provider.json
```diff
@@ -0,0 +1,19 @@
+{
```
bryangaleopenfin, line 1: Having this as a new file comes from the service tooling version bump, right?

src/provider/common/Injector.ts
```diff
@@ -97,11 +113,12 @@ export class Injector {
             }
         });
 
-        Injector._initialized = Promise.all(promises).then(() => {
+        Promise.all(promises).then(() => {
```
bryangaleopenfin: Can this just be:

    await Promise.all(promises)
    Injector._ready = true;
    Injector._initialized.resolve();

src/provider/common/Injector.ts
```diff
@@ -58,10 +59,7 @@ type Keys = (keyof typeof Inject & keyof typeof Bindings & keyof Types);
  * Wrapper around inversify that allows more concise injection
  */
 export class Injector {
-    private static _initialized: Promise<void>;
-    private static _ready: boolean = false;
-
-    private static _container: Container = (() => {
+    private static createContainer(): Container {
```
bryangaleopenfin: Can all static methods be together?

bryangaleopenfin: It's confusing me that this is only being called from reset, which doesn't seem to be called at all

src/provider/model/ConfigStore.ts
```diff
@@ -13,22 +13,19 @@ export interface ConfigStoreBinding {
...
         return this._store;
     }
 
     protected async init() {
-        const manifest = await fin.Application.getCurrentSync().getManifest();
-
-        if (manifest.config) {
-            this._store.add({level: 'desktop'}, manifest.config);
-        }
+        await this._loader.initialized;
```
bryangaleopenfin, line 29: This is another thing coming from the service tooling version bump, right?

bryangaleopenfin:
`package-lock.json` seems to be corrupted? Had to delete to get `npm i` to work, and after that project wouldn't compile

src/provider/model/AppDirectory.ts
```diff
@@ -97,54 +73,69 @@ export class AppDirectory extends AsyncInit {
...
-        await this.updateUrl(currentUrl);
     }
 
-    /**
-     * Fetch the AppDirectory.
-     * @param url Location of the AppDirectory.
-     * @param storedUrl Cache url
-     */
-    private async fetchData(url: string, storedUrl: string | null): Promise<Application[]> {
+    private async fetchOnline(url: string): Promise<Application[]|null> {
```
bryangaleopenfin: Can the naming be made symmetric? `fetchOnlineData` and `fetchCacheData`? (`fetchLiveData` and `fetchCachedData`?)

src/provider/model/AppDirectory.ts
```diff
@@ -33,32 +26,16 @@ export class AppDirectory extends AsyncInit {
...
-            // Set the application directory to our local copy during development.  In production it will be an empty string.
-            this._configStore.config.add({level: 'desktop'}, {applicationDirectory: DEV_APP_DIRECTORY_URL});
-        }
-
-        this.updateUrl(this._configStore.config.query({level: 'desktop'}).applicationDirectory);
-
-        const watch = new MaskWatch(this._configStore.config, {applicationDirectory: true});
-
-        watch.onAdd.add((rule: ScopedConfig<ConfigurationObject>, source: Scope) => {
-            this.updateUrl(this._configStore.config.query({level: 'desktop'}).applicationDirectory);
```
bryangaleopenfin, line 47: Why do we no longer want this?

src/provider/model/AppDirectory.ts
```diff
@@ -97,54 +73,69 @@ export class AppDirectory extends AsyncInit {
...
+
+        return null;
     }
 
     /**
      * Update the application directory in memory and storage.
      * @param applications To place into the directory.
      */
-    private async updateDirectory(applications: Application[]): Promise<void> {
+    private updateDirectory(applications: Application[]): void {
```
bryangaleopenfin: Suggest roll these two methods into one, with the url possibly being null, so we can easily see that the data and URL stay in sync

src/provider/model/AppDirectory.ts
```diff
@@ -97,54 +73,69 @@ export class AppDirectory extends AsyncInit {
...
-    /**
-     * Refresh the AppDirectory.
-     */
-    private async refreshDirectory(): Promise<void> {
-        // If using the demo app directory in production, set an empty array.
-        if (process.env.NODE_ENV === 'production') {
-            await this.updateDirectory([]);
-            return;
+    private async initializeDirectoryData(): Promise<void> {
+        this._url = this._configStore.config.query({level: 'desktop'}).applicationDirectory;
```
bryangaleopenfin, line 84: Just want to double check this is correct, wrt Chris's comment on the JIRA ticket

res/demo/provider.json
```diff
@@ -0,0 +1,19 @@
...
+        "url": "http://localhost:3923/provider/provider.html",
+        "uuid": "fdc3-service",
+        "autoShow": true,
+        "defaultHeight": 400,
+        "defaultWidth": 500
+    },
+    "runtime": {
+        "version": "13.76.43.31"
+    },
+    "serviceConfiguration": {
```
tomrobertsOF, line 16: This is just a custom key that we're using to pass some extra data through, not an official manifest param, right? Not super important, just want to make sure we're all on the same page about it.

test/provider/AppDirectory.unittest.ts
```diff
@@ -113,95 +99,113 @@ beforeEach(() => {
...
         ok: true,
-        json: mockJson
+        json: mockFetchReturnJson
     });
-
-    mockJson.mockResolvedValue(fakeApps);
+    mockFetchReturnJson.mockResolvedValue(fakeApps);
 });
 
 describe('AppDirectory Unit Tests', () => {
```
bryangaleopenfin: Do we need this outer level? Not seeing how it's distinct from "Given an App Directory with apps"

src/provider/model/AppDirectory.ts
tomrobertsOF, line 84: Just checking, but does the query return the default value if there's no desktop setting?

test/provider/AppDirectory.unittest.ts
```diff
@@ -113,95 +99,113 @@ beforeEach(() => {
...
+                }));
+
+                test('Data is not fetched from live app directory', () => {
+                    expect(appDirectory.getAllApps()).not.toEqual(fakeApps);
+                });
             });
         });
+    });
+
+    describe('When querying the Directory', () => {
```
bryangaleopenfin: Could these all go under "when we're online"? Think one of the tests is a duplicate?

src/provider/model/AppDirectory.ts
```diff
@@ -97,54 +73,69 @@ export class AppDirectory extends AsyncInit {
...
+    private async fetchOnline(url: string): Promise<Application[]|null> {
         const response = await fetch(url).catch(() => {
             console.warn(`Failed to fetch app directory @ ${url}`);
         });
 
         if (response && response.ok) {
-            return response.json();
+            try {
+                // validate the response is actually JSON
+                const validate = await response.json();
```
tomrobertsOF, line 103: Likely for another story, but is it also worth validating not only that it is a JSON, but also that it matches the shape of data we expect from an app directory?

src/provider/model/AppDirectory.ts
```diff
@@ -97,54 +73,69 @@ export class AppDirectory extends AsyncInit {
...
+    private fetchCacheData(): Application[]|null {
+        if (localStorage.getItem(StorageKeys.URL) === this._url) {
+            const cache = localStorage.getItem(StorageKeys.APPLICATIONS);
+
+            if (cache) {
+                try {
+                    const validate = JSON.parse(cache);
+                    return validate;
+                } catch (error) {
+                    // Not likely to get here but figured it's better to safely to handle it.
```
tomrobertsOF, line 122: Sensible choice IMO 👍

bryangaleopenfin (commented):
Struggling to get this to run on this branch, but possible I'm doing something silly.

Biggest thing for me is hierarchy of the unit tests - feel would be clearer being split into 'online' and 'offline', with the bulk being under 'online', then the few that are there to make sure things don't fall over 'offline'

src/provider/model/FinEnvironment.ts
tomrobertsOF, line 142: is this intentionally not awaited?

tomrobertsOF (commented):
Got a couple small comments for ya bud, but overall, it's some really top-notch typescript.

res/provider/config/fdc3-config.schema.json
rypete: @pjbroadbent may have more details, but eventually the public app directory will be located at this URL (or something similar) and maintained by OpenFin.

res/test/provider.json
rypete: We have the middleware, for some reason, that redirects `provider/sample-app-directory` to the test version.  It seemed a bit funny pointing it to test version in the demo even though they are the same thing.

res/demo/provider.json
rypete: Correct. It lets us have a custom provider manifest without changing the one that goes to production

src/provider/model/ConfigStore.ts
rypete: This one comes from a `service-config` bump.  Basic pattern as you see elsewhere to let us know the loader is ready to go

src/provider/model/AppDirectory.ts
rypete: This was a sort of workaround if you remember before.  We no longer need it because we are setting the application directory elsewhere (no longer checking if we are in dev). Also the application directory location cannot change once the service has started so we don't need to monitor it.

res/demo/provider.json
rypete: You are correct.  This is actually a secret key that an end user shouldn't use.

src/provider/model/AppDirectory.ts
rypete: Yes, the query will use the default value if none other has been set

res/demo/provider.json
tomrobertsOF: Cool 👍

src/provider/model/AppDirectory.ts
tomrobertsOF: Awesome, thanks 👍

src/provider/model/AppDirectory.ts
rypete: 😁

src/provider/model/AppDirectory.ts
rypete: I'm not sure we want to do that.  These are only used in one place. We need to only update the URL if the fetch succeeds, otherwise we would lose the cache if the application were to be reloaded as we check the URL provided from the application directory against the one in storage.

bryangaleopenfin: How about, in `initializeDirectoryData`

    if (fetchedData)
    {
        this.updateCache(this._url, fetchedData);
    }
    this.updateDirectory(fetchedData || cachedData || []);

with `updateDirectory` not changing the cache?

Or update the cache from within `fetchOnline` instead of `updateDirectory`?

src/provider/model/FinEnvironment.ts
rypete: Yes this is intentional.  Awaiting it `Injector.initialized` will never resolve because it itself awaits this init method.  I will add a comment explaining this above this line

rypete: We can't await the injector because we are calling this from within an `init` method that the injector calls, so it's a bit circular.  The reason it is needed is because we were running into the issue of window creation logic running before in the model before all of the injectables were ready

rypete: I've put a comment above this line explaining the await issue

src/provider/common/Injector.ts
rypete: Sure, that would make sense

src/provider/common/Injector.ts
rypete: This was in error!

src/provider/model/AppDirectory.ts
rypete: Verified this is correct.  The context of manifest in Chris's comment is the providers manifest, not application.  The config loader accounts for this when populating the desktop level

src/provider/model/AppDirectory.ts
rypete: I still feel that is not much better as we are now writing the cache twice in the case of truthy fetchedData (which I noticed is also the case with calling it with cachedData...) and we are just replacing one method with another.  

Would you settle with: 
```
        if (fetchedData) {
            this.updateUrl(this._url);
            this.updateDirectory(fetchedData);
        } else {
            this._directory = cachedData || []
        }

```
Or keep the above structure and remove the methods completely, and update the members / call storage directly in the if?

bryangaleopenfin: Let's talk about this one, grab me when you've got two minutes

src/provider/model/AppDirectory.ts
rypete: It crossed my mind but its not a very straightforward thing to do.  Its definitely something to look at in the future

rypete: Also keep in mind the structure is an agreed upon thing by working group https://fdc3.finos.org/1.0/appd-spec#tag/Application/paths/~1v1~1apps~1{appId}/get

tomrobertsOF: The structure is agreed, but what a user puts in the config is not. I think a guard confirming that the data we get is compliant with the spec you linked is highly advisable, but agree that it is likely out of scope of this story. If we can throw something in the backlog to make sure it's not forgotten then I'd be happy to mark this resolved.

src/provider/common/Injector.ts
rypete: Done

src/provider/common/Injector.ts
rypete: Done

rypete: Fixed

src/provider/model/AppDirectory.ts
rypete: Fixed

src/provider/model/AppDirectory.ts
rypete: Fixed as discussed

src/provider/common/Injector.ts
bryangaleopenfin: Wait, actually, I'm confused about why this couldn't stay as it was, without the deferred promise

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Huh. Makes sense. Has something changed that's made this a problem? Or did we just not see this problem before?

rypete: I think it was masked by the way we were setting the application directory before using the service config loader

src/provider/common/Injector.ts
rypete: I'm not entirely sure what you mean.  We use `Injector.initialized` for some promise based stuff like as seen with the circular reference, and `_ready` internally for non-async methods

test/provider/AppDirectory.unittest.ts
rypete: The online stuff is in context of 'is the app directory url we are provided online', not the actual service. We have logic to handle both cases (caching etc) which is why we test it.  Actual offline functionality in the service hasnt been considered as we dont even have the service worker yet

src/provider/common/Injector.ts
bryangaleopenfin: I mean, before `_initialized` was just a promise, and the end of `init()` was written in a way that a `DeferredPromise` wasn't needed.

Though ah, I see, this way avoids any problems if someone does `await Injector.initialized` *before* `Injector.init()` is called

test/provider/AppDirectory.unittest.ts
rypete: Youre right, we don't.  It was from before my time on this ticket so I left, but have removed now

src/provider/model/AppDirectory.ts
rypete: OK I have tagged with SERVICE-620

tomrobertsOF: Awesome! Thanks buddy!

tomrobertsOF (approved):

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Couple of paranoid suggestions - 
- Move `const windowInfo = await fin.System.getAllWindows();` outside the `then` block, so that we do this as soon as possible after adding the event listeners (or put it right before we add the event listeners), so we avoid a badly timed window creation being processed twice

- Put `await Injector.initialized` as the first line of the two event listeners, so we guarantee that not signals get dispatched until the service is properly ready

rypete: Added both

bryangaleopenfin (approved):

PR #136, opened 2019-09-11
[SERIVCE-596] Update Channels API to latest WG-approved terminology

tomrobertsOF:
Changes to bring us inline with slight changes made to the draft specification for FDC3 channels. 

Lots of small changes, all focused around the following main changes:
-  Move from `name` and `color` properties on the channel object itself to a `visualIdentity` object which contains these (and possibly other) properties.
- Rename `DesktopChannel` to `SystemChannel`

Note: while the provider now includes a `glyph` in the builtin channels' `visualIdentity`s, no work has been done to display this in the demo apps. Also, the glyphs currently point to the openfin favicon. Both of these are likely to change in the near future as part of a separate effort.


src/client/contextChannels.ts
```diff
@@ -21,10 +21,32 @@ import {getEventRouter} from './EventRouter';
...
+ * this is their meaning.
+ */
+export interface DisplayMetadata {
+    /**
+     * A user-readable name for this channel, e.g: `"Red"`
+     */
+    name: string;
+
+    /**
+     * The color that should be associated within this channel when displaying this channel in a UI, e.g: `0xFF0000`.
```
bryangaleopenfin: Should this be "#FF0000"? And emphasise this is in HTML colour format

src/provider/model/ContextChannel.ts
```diff
@@ -81,7 +80,7 @@ export class DesktopContextChannel extends ContextChannelBase {
         this._context = null;
     }
 
-    public serialize(): DesktopChannelTransport {
-        return {name: this.name, color: this.color, ...super.serialize() as {id: string, type: 'desktop'}};
+    public serialize(): SystemChannelTransport {
+        return {visualIdentity: this.visualIdentity, ...super.serialize() as {id: string, type: 'system'}};
```
bryangaleopenfin, line 84: Worth doing `visualIdentity: {...this.visualIdentity}`?

src/client/main.ts
```diff
@@ -199,8 +199,8 @@ export async function findIntentsByContext(context: Context): Promise<AppIntent[
  *
  * @throws `TypeError`: If `context` is not a valid Context
  */
-export function broadcast(context: Context): void {
-    tryServiceDispatch(APIFromClientTopic.BROADCAST, {context: parseContext(context)});
+export async function broadcast(context: Context): Promise<void> {
```
bryangaleopenfin, line 202: This scares me a bit, but sounded like this has been fully thrashed out?

src/client/contextChannels.ts
```diff
@@ -21,10 +21,32 @@ import {getEventRouter} from './EventRouter';
  */
 export type ChannelId = string;
 
+/**
+ * A system channel will be global enough to have a presence across many apps. This gives us some hints
+ * to render them in a standard way. It is assumed it may have other properties too, but if it has these,
+ * this is their meaning.
+ */
```
bryangaleopenfin, line 26: Would rewrite to start with a one-sentence description of what this is. "Defines the visual appearance of a system channel when presented in an app, for example, as part of a channel selector"? The "if' in the last sentence also feels odd, since these aren't optional properties

src/client/contextChannels.ts
tomrobertsOF: Yes! Missed that one.

src/provider/model/ContextChannel.ts
tomrobertsOF: Yea, probably.

src/client/main.ts
tomrobertsOF: Well..... the result of the discussions was that any other side-effects in the provider from this would be handled in some of the "pre 1.0 cleanup" stories (like the "make things consistent" one)

src/client/contextChannels.ts
tomrobertsOF: I copied all the doc comments on these from the FDC3 spec docs. They could probably do with some general improvements.

src/client/main.ts
tomrobertsOF: But as far as this story is concerned, yes, it is a settled matter.

src/client/contextChannels.ts
tomrobertsOF: Also, the "if" is because they are optional in the spec.

bryangaleopenfin (approved):

src/client/contextChannels.ts
```diff
@@ -321,34 +341,28 @@ export abstract class ChannelBase {
...
-    /**
-     * A user-readable name for this channel, e.g: `"Red"`
-     */
-    public readonly name: string;
+export class SystemChannel extends ChannelBase {
+    public readonly type!: 'system';
 
     /**
-     * The color that should be associated within this channel when displaying this channel in a UI, e.g: `0xFF0000`.
+     * SystemChannels may well be selectable by users. Here are the hints on how to see them.
```
pjbroadbent: > may well be selectable by users

Can we word this a little more strongly? This is the main intention of these types of channels.

src/provider/constants.ts
```diff
@@ -23,35 +23,53 @@ export const Timeouts = {
...
+export const SYSTEM_CHANNELS = [
     {
         id: 'red',
-        name: 'Red',
-        color: 0xFF0000
+        visualIdentity: {
+            name: 'Red',
+            color: '#FF0000',
+            glyph: 'https://openfin.co/favicon.ico'
+        }
```
pjbroadbent, line 33: Thinking forward to when we make this desktop-configurable (since this is essentially temporary, until we replace this with a config store query), would we want `visualIdentity` within the config schema, or the previous "flat" style?

src/provider/model/ContextChannel.ts
```diff
@@ -81,7 +80,7 @@ export class DesktopContextChannel extends ContextChannelBase {
         this._context = null;
     }
 
-    public serialize(): DesktopChannelTransport {
-        return {name: this.name, color: this.color, ...super.serialize() as {id: string, type: 'desktop'}};
+    public serialize(): SystemChannelTransport {
+        return {visualIdentity: {...this.visualIdentity}, ...super.serialize() as {id: string, type: 'system'}};
```
pjbroadbent: Do we need the clone on `visualIdentity`? I think the only thing we ever do with these serialised objects is immediately send them over the wire.

Should the return types of all these `serialize` functions be `Readonly<T>` types, so we know we can safely return without copying?

src/provider/constants.ts
tomrobertsOF: I think matching it to the spec as closely as possible is the best way forward, but if you think differently then just say and I'll change it.

src/provider/model/ContextChannel.ts
tomrobertsOF: Argue with @bryangaleOF about this since he told me to add it. 😛

src/client/contextChannels.ts
tomrobertsOF: How would you word it? This was just taken from the spec.

bryangaleopenfin: "How a client application should present this channel in any UI" (Sat with Phil and we like this wording)

src/provider/model/ContextChannel.ts
bryangaleopenfin: Discussed with Phil - we think get rid of the `...`, but change the return type to `Readonly<SystemChannelTransport>`, with similar change to the base class

src/client/contextChannels.ts
tomrobertsOF: 👍

src/provider/model/ContextChannel.ts
tomrobertsOF: 👍

tomrobertsOF:
@pjbroadbent, @bryangaleOF, pushed up changes base on your most recent comments. Should be ready to go now.

src/provider/model/ContextChannel.ts
bryangaleopenfin: Also change on the interface?

tomrobertsOF: How many places do we need to define this?!?

👍

pjbroadbent (approved):

PR #137, opened 2019-09-11
[SERVICE-533] Deferred breaking API clean-up changes

pjbroadbent:
Minor clean-up to the types/constants used for client-provider communication. Largely issues flagged as tech debt where the fix was deferred until we were making a breaking change for other reasons.

Note: Not all changes are actually breaking, PR also contains just some general misc clean-up.

Summary of changes:
- More consistent naming of provider-to-client message keys
- More consistent naming of payload types (now 1-1 mapping between API's and payload types)
  - All values passed over the IAB channel are wrapped in payload objects
- Removed redundant app directory property
- Explicit typing of the listener object returned when listening directly to a channel
- Removed unimplemented 'identity' property from `addEventListener`


tomrobertsOF (approved):

src/client/contextChannels.ts
```diff
@@ -126,7 +126,7 @@ export interface ChannelChangedEvent {
     previousChannel: Channel|null;
 }
 
-interface ChannelContextListener extends ContextListener {
+export interface ChannelContextListener extends ContextListener {
```
bryangaleopenfin, line 149: Developer facing interface without docs?

bryangaleopenfin (commented):
Do we want to unify EventRouter between notifications and fdc3 as part of this?

fdc3 uses {type: 'main', id: 'main'}, notifs uses 'default' for top level events

bryangaleopenfin (approved):

PR #141, opened 2019-09-24
Create NOTICE file with content aligned to FINOS project blueprint

agitana:


mcleo-d:
Hi @chuckdoerr,

I've noticed the build checks of this PR have failed. Is it possible to work with @mcleo-d and @aitana16 to resolve so we can get the PR merged?

Many thanks,

James.

PR #143, opened 2019-09-26
[SERVICE-578] The FDC3 service respects the optional context field on the findIntents call

ngunday:
The optional context parameter in findIntents call in the FDC3 now filter out the apps that satisfy the specified intent type but not the context.

If a context is specified, the non-directory apps will not be included as there is no way to know the intent context info of these app in provider at the moment.


src/provider/index.ts
```diff
@@ -128,7 +128,7 @@ export class Main {
     private async findIntent(payload: FindIntentPayload): Promise<AppIntent> {
         let apps: Application[];
         if (payload.intent) {
-            apps = await this._model.getApplicationsForIntent(payload.intent);
+            apps = await this._model.getApplicationsForIntent(payload.intent, payload.context && parseContext(payload.context).type);
```
bryangaleopenfin, line 131: Do we want anything  to happen in the case where `payload.context` exists but not `payload.intent`?

test/demo/findIntent.inttest.ts
```diff
@@ -0,0 +1,38 @@
+import 'jest';
```
bryangaleopenfin, line 1: We have `resolveByIntent.inttest.ts`. Suggest rename that file to `findIntent.inttest.ts`, and integrate these tests into that file (there's a "todo" at the bottom of that file). Or if it's worth splitting into two files, follow the same pattern as we're following with `raiseIntent/`

src/provider/model/Model.ts
```diff
@@ -140,27 +140,34 @@ export class Model {
...
-                    // AppInfo has already been added by another window on the same app also listening for the same intent
-                    return apps;
-                }
-                return apps.concat([appWindow.appInfo]);
-            }, []);
+        const directoryAppsWithIntent = await this._directory.getAppsByIntent(intentType, contextType);
+
+        // If context is specified don't use apps in model as there's no way to know about the context of non-directory
+        // app intents at the moment.
+        if (!contextType) {
```
bryangaleopenfin: Would do this as

    if (contextType) {
        return directoryAppsWithIntent;
    } else {
        ....
    }

Keeps the comment with the branch it applies to

src/provider/model/Model.ts
```diff
@@ -140,27 +140,34 @@ export class Model {
      *
      * Includes windows that are not in the app directory but have registered a listener for it
      * @param intentType intent type
+     * @param contextType context type
```
bryangaleopenfin, line 143: Update comment to match new behaviour. That second line seems misleading now

src/provider/model/AppDirectory.ts
```diff
@@ -35,9 +35,12 @@ export class AppDirectory extends AsyncInit {
         }) || null;
     }
 
-    public async getAppsByIntent(intentType: string): Promise<Application[]> {
+    public async getAppsByIntent(intentType: string, contextType?: string): Promise<Application[]> {
         return this._directory.filter((app: Application) => {
-            return app.intents && app.intents.some(intent => intent.name === intentType);
+            return app.intents && app.intents.some(intent => {
+                const isContextRelevant = contextType && intent.contexts && intent.contexts.length > 0;
+                return (intent.name === intentType) && (!isContextRelevant || intent.contexts.some(context => context === contextType));
```
bryangaleopenfin: Can do `intents.contexts.includes(contextType)` instead of `intent.contexts.some(context => context === contextType))`

src/provider/model/AppDirectory.ts
```diff
@@ -35,9 +35,12 @@ export class AppDirectory extends AsyncInit {
         }) || null;
     }
 
-    public async getAppsByIntent(intentType: string): Promise<Application[]> {
+    public async getAppsByIntent(intentType: string, contextType?: string): Promise<Application[]> {
```
bryangaleopenfin, line 38: Consider do this as two filter steps to cut down the nesting. Thoughts on something like:

        const appsWithIntent = this._directory.filter((app: Application) => {
            return app.intents && app.intents.some(intent => intent.name === intentType);
        });

        if (contextType) {
            return appsWithIntent.filter(...);
        } else {
            return appsWithIntent;
        }

?

bryangaleopenfin, line 38: Has the behaviour where we essentially treat an empty list of context types on the intent as a special case been thrashed out?

res/test/sample-app-directory.json
```diff
@@ -87,7 +90,10 @@
...
     ],
     "intents": [
-        {"name": "DialCall"},
+        {
+            "name": "DialCall",
+            "contexts": ["fdc3.organization"]
+        },
         {"name": "StartCall"},
         {"name": "SaveContact"},
         {"name": "ViewChart"}
```
bryangaleopenfin: Think the fake data here would be more robust if we did something like `{"name": "test.IntentWithContexts"}`, `{"name": "test.IntentWithoutContexts"}`, etc. (like we have `{"name": "test.IntentNamePreregistered"}` right now). It'd at least make it more obvious why these particular apps have these differences and what we're testing for.

test/demo/findIntent.inttest.ts
```diff
@@ -0,0 +1,38 @@
...
+    describe('When calling findIntent with an intent type and context type accepted by some directory app', () => {
+        const intent = {
+            type: 'DialCall'
+        };
+        const context = {
+            type: 'fdc3.contact'
+        };
+        test('The promise resolves to an array of all AppIntents which has the specified intent with the specified context or no context at all', async () => {
+            const receivedAppIntents = await findIntent(intent.type, context);
+            expect(receivedAppIntents.apps).toHaveLength(4);
```
bryangaleopenfin: Suggest test for specific apps being present/not being present. Going off total length feels like it'll break easily if the sample app directory gets added to

src/provider/model/AppDirectory.ts
bryangaleopenfin: Think update the interface to specify that `contexts` is optional on an `Intent`. This seems to match the spec - https://fdc3.finos.org/appd-spec#tag/Application/paths/~1v1~1apps/post

bryangaleopenfin (commented):
Should the documentation on `findIntents` in `main.ts` be updated to match the new behaviour? Right now it's a copy & paste from the spec. Do we want to keep it that way? (@pjbroadbent?)

src/provider/model/AppDirectory.ts
noyangunday: I think I might be missing some info on that one. I thought where there's an empty list of context types it should mean the intent is interested in all contexts?

bryangaleopenfin: Could well be the case, any idea where that came from? Though it's not consistent with the behaviour of `getAppIntentsByContext`.

noyangunday: this wouldn't cut down the nesting as in your example the initial filter() doesn't really give me an array of intents to work on, it only filters out the irrelevant apps ; the main problem stays the same where contextType is defined, meaning still need to iterate through each intent of each `appsWithIntent ` to find the relevant ones to me and then contexts of those intents

src/provider/index.ts
noyangunday: that's essentially getAppIntentsByContext() with it's dedicated API `APIFromClientTopic.FIND_INTENTS_BY_CONTEXT`. should we replicate it (or redirect to it from) here?

src/provider/model/Model.ts
noyangunday: agreed, cleaner.

test/demo/findIntent.inttest.ts
noyangunday: Wouldn't that mean replicating the whole App data for 6 apps in sample app directory in this test which probably would, again, make it so easily breakable when something in the app-directory data has changed?

I believe that's a risk we run as long as we use a shared app directory data for all tests and not encapsulate them per test? The tests will be breakable regardless when the data is changed in a specific way, would it be reasonable to accept it's the future test writer's responsibility to make sure their change in test data hasn't broken anything else/fix it if needed rather than trying to exhaust all the possible test failures prospectively? 

But I believe your {"name": "test.IntentWithContexts"} approach makes it much less likely to be breakable as that is specific naming of the intent for this test.

src/provider/model/AppDirectory.ts
ngunday: It's purely a guess actually. I couldn't think of any logical reasons why contexts being an empty array and contexts being undefined should be different in behaviour?

bryangaleopenfin: Think we might be at crossed wires here. Let's discuss in person

test/demo/findIntent.inttest.ts
bryangaleopenfin: You wouldn't need to replicate the app data - have a look at how the other tests in `resolveByIntent.inttest.ts` use `extractSortedAppNames`.

You could do `expect(extractSortedAppNames(receivedAppIntents.apps)).toEqual(['test-app-1', ...]);`

src/provider/model/AppDirectory.ts
bryangaleopenfin: Ah, yeah, I see. Had another go at simplifying. This feels more glance-able to me, but your call on this.

        const appsWithIntent = this._directory.filter(app => app.intents && app.intents.some(intent => intent.name === intentType));
    
        if (contextType) {
            return appsWithIntent.filter(app => app.intents!.some(intent => !intent.contexts || intent.contexts.includes(contextType)));
        } else {
            return appsWithIntent;
        }

res/test/sample-app-directory.json
bryangaleopenfin: Cool. Think follow the `test.` convention here, so it's glance-able what's test specific

src/provider/index.ts
bryangaleopenfin: Hmm, actually, digging a bit deeper I'm more confused. Going by the interface, `payload.intent` is required.

Hey @pjbroadbent - the workaround here where `findIntent` can end up returning all apps. Can we get rid of it? AFAICT, it's unused and undocumented.

And I know this is scope creeping now, but the value we're giving the the `displayName` for the intent. We should be smarter about this, right?

src/provider/model/Model.ts
bryangaleopenfin: Ah, I didn't mean just update the second line, something like - "Get apps that can handle an intent and optionally with specified context type" (struggling to word this to make sense briefly, but something like that) - for the first sentence?

test/demo/findIntent.inttest.ts
bryangaleopenfin: Cool. Worth doing a pass to bring the new tests more in line with the existing tests

src/provider/model/AppDirectory.ts
noyangunday: ...`appsWithIntent.filter(app => app.intents!.some(intent => !intent.contexts ||`...

I believe this bit will return a false positive if the app has an unrelated intent which has no context (or has the related context in an unrelated intent), higher up in its intents list.

e.g:
```
app: {
    ... ,
    "intents": [
            {"name": "Irrelevant Intent"},
            {
                "name": "Relevant Intent",
                "contexts": ["some.context"]
            }
    ],
    ...
}
```

test/demo/findIntent.inttest.ts
noyangunday: The existing tests seem to be testing app specific (such as listener-related behaviour etc) whereas these ones are testing the API behaviour directly on app-directory. Couldn't think of a nice way to make them look very similar, I am open for some suggestions with this one :)

test/demo/findIntent.inttest.ts
noyangunday: I have changed this test to check for the names similarly.

While taking a look at other tests tho, I think I might be a little sceptical about the way we use getDirectoryAppsForIntent() to get the expected names for the test. We're basically replicating the functionality of the provider within the test with this method, essentially turns these tests into tautological tests. I'd prefer to hard-code these names in the test to expected values than to use this method.

src/provider/model/AppDirectory.ts
bryangaleopenfin: Aah, yeah, I see 👍

test/demo/findIntent.inttest.ts
bryangaleopenfin: Oh, yeah, I raised an eyebrow at that too. I'll be embarrassed if it turns out I wrote those. Definitely prefer check for specific names

test/demo/findIntent.inttest.ts
bryangaleopenfin: I think the first and last test in the "without context" block are essentially doing the same thing, and even if you want to keep both, they both belong in the 'Without ad-hoc apps running' block

bryangaleopenfin: Though, looking at those tests, it does raise questions about how we want this new behaviour to interact with `addIntentListener`.

I feels like it should be the case that `findIntent(intent)` will always return more results than `findIntent(intent, context)` (so context is refining the results, never expanding them), but that isn't necessarily the case with the current implementation.

Grab me when you're at a good point to discuss?

bryangaleopenfin: Try working through `getApplicationsForIntent`, imagining we have a single app, "Dialer". Imagine it's in the app directory with the "fdc3.DialCall" intent and no contexts, and it's running, but hasn't added an intent listener.

Let's say we call `getApplicationsForIntent('fdc3.DialCall', 'any-context")`.
- `directoryAppsWithIntent` will contain "Dialer"
- Since we've specified a context `directoryAppsWithIntent` will be returned as is, so the result contains "Dialer"

Let's say we call `getApplicationsForIntent('fdc3.DialCall')`.
- `allAppWindows` will contain "Dialer" (because "Dialer" is running)
- `directoryAppsWithIntent` will contain "Dialer"
- Since we've not specified the context, the execution will take the `else` branch
- `appsInModelWithIntent` **won't** contain "Dialer" (because "Dialer" hasn't added an intent listener)
- `directoryAppsNotInModel` **won't** contain "Dialer" (because "Dialer" *is* in the model)
- So our return result `[...appsInModelWithIntent, ...directoryAppsNotInModel]` won't contain "Dialer"

So this is a case where specifying the context grows rather than shrinks what gets returned. Have I got anything wrong here?

pjbroadbent: Fully agree that `getApplicationsForIntent(intent, context)` should be a strict subset of `getApplicationsForIntent(intent)` - in the former, can we "validate" the results we get from the directory against the model, and remove any where the app hasn't (yet?) registered listeners for intent types that it claimed support for within the directory entry?

src/provider/model/AppDirectory.ts
pjbroadbent: Just to summarise earlier conversation - we should treat an empty array in the same way as an `undefined` contexts list. This is because otherwise the intent definition would be unusable, and we would have `findIntents` (when called without a context type) returning intent definitions that aren't possible to actually use with `raiseIntent`.

test/demo/findIntent.inttest.ts
noyangunday: `directoryAppsNotInModel` should be a difference set of directory apps and `appsInModelWithIntent`. 

Given "Dialer" is not included in `appsInModelWithIntent` it should be included in `directoryAppsNotInModel`

this should guarantee `getApplicationsForIntent(intent, context)` is a strict subset of `getApplicationsForIntent(intent)`.

noyangunday: I am suggesting to change to this:
```            
const directoryAppsNotInModel = directoryAppsWithIntent.filter(directoryApp => !appsInModelWithIntent.some(app => app.appId === directoryApp.appId));```

Comment (bryangaleopenfin) [reply test/demo/findIntent.inttest.ts:1]: So that's a change in behaviour of the "without-context" case, which previously would not have returned "Dialer", but now will do. And so I think the test "When calling findIntent with the intent, the app is NOT returned" will fail.

I think it's more we want to reduce what gets returned in the "with-context" case, not expand what gets returned in the "without-context" case.

Unless you're intentionally proposing that we change the existing behaviour here.

Comment (noyangunday) [reply test/demo/findIntent.inttest.ts:1]: @bryangaleOF Is there a specific reason why we don't want to return "Dialer" in "without-context" case if it declares "fdc3.DialCall" as an intent? I thought the idea of `getApplicationsForIntent` was to return all the application that matches the specified intent (and context if eligible)

Comment (bryangaleopenfin) [reply test/demo/findIntent.inttest.ts:1]: Since Dialer is running, and we know it hasn't registered any intent listeners, we know in practice that it's unable to handle the the intent. That's given precedence over what's in the app directory

Comment (noyangunday) [reply test/demo/findIntent.inttest.ts:1]: oh, gotcha!

File: src/provider/index.ts (hunk 1 above)
Comment (noyangunday) [reply src/provider/index.ts:131]: @bryangaleOF @pjbroadbent 
So, in our latest conversation we concluded, the workaround Bryan mentions here is still needed as that's currently the only way to get all apps regardless of intent which is undocumented because it still requires the workgroup agreement and a seperate api but is needed in the near future, and is used by the demo apps. So I am reverting this back to include the workaround. Are we all happy with this?

Comment (bryangaleopenfin) [reply src/provider/index.ts:131]: Yeah, fine with the workaround staying since something's using it. Though then there's still the question of how to handle the "context specified, but intent not" case

Comment (noyangunday) [reply src/provider/index.ts:131]: I think the only case we use "no intent" is when we want to get all apps (as this is the main reason for having this temporary workaround) maybe we shouldn't worry about context at all in that case?

Comment (bryangaleOF) [reply src/provider/index.ts]: Yeah, I'm good with that

Comment (bryangaleopenfin) [reply src/provider/index.ts:131]: Yeah, I'm good with that

File: test/demo/findIntent.inttest.ts (hunk 2 above)
Comment (ngunday) [reply test/demo/findIntent.inttest.ts:1]: 👍

File: src/provider/index.ts (hunk 1 above)
Comment (ngunday) [reply src/provider/index.ts:131]: 👍

File: src/provider/model/AppDirectory.ts (hunk 6 above)
Comment (ngunday) [reply src/provider/model/AppDirectory.ts:38]: 👍

File: src/provider/model/Model.ts [hunk 9]
```diff
@@ -136,37 +136,51 @@ export class Model {
...
-     * Gets apps that can handle an intent
+     * Get apps that can handle an intent and optionally with specified context type
      *
-     * Includes windows that are not in the app directory but have registered a listener for it
+     * Includes windows that are not in the app directory but have registered a listener for it if contextType is not specified
      * @param intentType intent type
+     * @param contextType context type
      */
-    public async getApplicationsForIntent(intentType: string): Promise<Application[]> {
+    public async getApplicationsForIntent(intentType: string, contextType?: string): Promise<Application[]> {
```
bryangaleopenfin, line 145: I'm wondering if in the context-provided case, we should include
- Running non-directory apps that have registered a listener for the intent
- Running directory apps that have registered a listener for the intent, where the app directory says they don't handle that intent at all

Those are both cases where we know an app handles an intent (because it has that listener) but don't know which contexts it handles for that intent (because the app directory doesn't tell us). Should we, for consistency, handle those the same as we'd handle an app whose context list for that intent is undefined?

This would still respect the constraint that specifying a `contextType` should narrow down the result. And avoids the weird situation that FDC3 could say that an app supports a particular intent, but would say it supports no particular context types for that intent.

Thoughts?

src/provider/index.ts
bryangaleopenfin: Will resolve with once we have a story in for a better displayName. Shout if you want me to write that up

test/demo/findIntent.inttest.ts
```diff
@@ -0,0 +1,142 @@
...
+
+            // Resolve the valid intent
+            const appIntent = await findIntent(validIntent);
+
+            // Compare the two arrays (sorted as order doesn't matter)
+            expect(extractSortedAppNames(appIntent)).toEqual(expectedAppNames);
+        });
+
+        test('When resolving an intent without a specific context, the returned object lists all \
+directory applications accepts the intent regardless of their context specifications.', async () => {
```
bryangaleopenfin: "directory applications accepts" -> "directory applications that accept"

test/demo/findIntent.inttest.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        test('When calling resolve with an intent which no directory applications accept, the promise resolves to an empty array', async () => {
+            // Resolve the invalid intent
+            const appIntent = await findIntent(invalidIntent);
+
+            // Expect no apps returned
+            expect(extractSortedAppNames(appIntent)).toEqual([]);
+        });
+
+        describe('When a directory app is running', () => {
```
bryangaleopenfin, line 66: Now that running apps affect the result in the "with context" case, that's worth testing

src/provider/index.ts
noyangunday: 👍

src/provider/model/Model.ts
noyangunday: I recall discussing these specific cases with @pjbroadbent . He thinks we should not include them when the context is provided but the intent-context information about the app is unknown (in non-app directory case) or unreported (in directory case). My personal opinion is the same as Phil's but happy to change it to that if there's consensus for the opposite.

test/demo/findIntent.inttest.ts
noyangunday: 👍

src/provider/model/Model.ts
bryangaleopenfin: 👍

test/demo/findIntent.inttest.ts
noyangunday: 👍

bryangaleopenfin (approved):

tomrobertsOF (approved):
All looks fine to me from a code perspective. 

Assuming the comments around expected behaviour in edge cases have been discussed and agreed on? Either way, I have no strong feelings on that, so approved from me.

PR #144, opened 2019-09-27
[SERVICE-599] Verify proposed "Channel Restoration" flows

bryangaleopenfin:
This PR:
- Bumps up the `WINDOW_EXPECT_TO_SEEN` timeout (just over 100 seen in testing, so going 100 -> 250 this seems safe)
- Improves our standard channel selector.
- Adds an integration test to maintain ensure that we can interact with windows as soon as they are created

Please also regard this doc as part of this PR - https://docs.google.com/document/d/1Kn8gLqm36mKx5bjgdUjvQXMr6I92VyYwSsYm5B4d-bs/

This has been updated with code fixes, and to remove an old flow that listened for `channel-changed` rather than `window-created`, which is now unnecessary with the expect system.

Suggest leave doc comments in the doc itself

### DIFF
--- a/src/provider/constants.ts
+++ b/src/provider/constants.ts
@@ -15,7 +15,7 @@ export const Timeouts = {
     /**
      * Time service allows for a `window-created` event after a client expects the window to exist
      */
-    WINDOW_EXPECT_TO_SEEN: 100,
+    WINDOW_EXPECT_TO_SEEN: 250,
 
     /**
      * Time service allows for a window to go from first seen to being fully registered with the model
--- a/src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
+++ b/src/demo/components/ContextChannelSelector/ContextChannelSelector.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 
-import {Channel, defaultChannel, getCurrentChannel, getSystemChannels, SystemChannel} from '../../../client/contextChannels';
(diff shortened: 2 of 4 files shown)


tomrobertsOF (approved):
Looks good to me. Read through the document a while back and have no major objections.

src/provider/constants.ts
```diff
@@ -15,7 +15,7 @@ export const Timeouts = {
     /**
      * Time service allows for a `window-created` event after a client expects the window to exist
      */
-    WINDOW_EXPECT_TO_SEEN: 100,
+    WINDOW_EXPECT_TO_SEEN: 250,
```
ngunday, line 18: interesting naming :)

ngunday (approved):
Looks good to merge

src/provider/constants.ts
bryangaleopenfin: Yeah, reads a bit odd, but follows from the terminology in `Environment`. Was about to change to MAX_WINDOW_EXPECT_TO_SEEN_INTERVAL, but feel like that doesn't add anything that isn't implicit from it being under `Timeouts`

bryangaleopenfin:
Made some small doc changes to account for app channels. It still refers to `allChannels` in the code, even though that's not accurate, but I'm trying to keep within the 71 character line length the doc gives me!

Thumbs up before I merge?

ngunday:
> Made some small doc changes to account for app channels. It still refers to `allChannels` in the code, even though that's not accurate, but I'm trying to keep within the 71 character line length the doc gives me!
> 
> Thumbs up before I merge?

Changes look good. LGTM

PR #145, opened 2019-10-01
[SERVICE-597] Implement "App" channels

bryangaleopenfin:
This PR
- Adds support for app channels
- Adds integration tests - existing tests are expanded to cover app channels where makes sense, plus a new test file for app channel specific
- Tidies channel tests, to use bookend functions from `common.ts`
- Adds unit tests
- Sets a consistent order when working with channel types - default, system, app
- Adds user-facing doc comments

### DIFF
--- a/src/client/contextChannels.ts
+++ b/src/client/contextChannels.ts
@@ -416,7 +416,7 @@ export const DEFAULT_CHANNEL_ID: ChannelId = 'default';
  * All windows will belong to exactly one channel at all times. If they have not explicitly been placed into a channel
  * via a {@link ChannelBase.join} call, they will be in this channel.
  *
- * If an app wishes to leave a system/app channel it can do so by re-joining this channel.
+ * If an app wishes to leave any other channel, it can do so by (re-)joining this channel.
  */
 export const defaultChannel: DefaultChannel = new DefaultChannel();
 
--- a/src/client/contextChannels.ts
+++ b/src/client/contextChannels.ts
@@ -383,8 +383,8 @@ export class SystemChannel extends ChannelBase {
  *
  * Applications can create these for specialised use-cases.  These channels should be obtained by name by calling
  * {@link getOrCreateAppChannel} and it is up to your organization to decide how applications are aware of this name.
- * It is recommended that channel names are prefixed with your organization's reverse domain name to avoid name
- * collisions, e.g., 'com.company-name.channel-name'.
+ * As with organization defined contexts, app channel names should have a prefix specific to your organization to avoid
+ * name collisions, e.g. 'company-name.channel-name'.
  *
  * App channels can be joined by any window, but are only indirectly discoverable if the name is not know.
  */
@@ -468,9 +468,9 @@ export async function getCurrentChannel(identity?: Identity): Promise<Channel> {
 /**
  * Returns an app channel with the given name. Either creates a new channel or returns an existing channel.
  *
- * It is up to your organization to decide how to share knowledge of these custom channels. It is recommended that
- * channel names are prefixed with your organization's reverse domain name to avoid name collisions, e.g.,
- * 'com.company-name.channel-name'.
+ * It is up to your organization to decide how to share knowledge of these custom channels. As with organization
+ * defined contexts, app channel names should have a prefix specific to your organization to avoid name collisions,
+ * e.g. 'company-name.channel-name'.
  *
  * The service will assign a unique ID when creating a new app channel, but no particular mapping of name to ID should
  * be assumed.
--- a/src/client/contextChannels.ts
+++ b/src/client/contextChannels.ts
@@ -465,11 +465,11 @@ export async function getCurrentChannel(identity?: Identity): Promise<Channel> {
  *
  * It is up to applications to manage how to share knowledge of these custom channels across windows and to manage
  * channel ownership and lifecycle.
- * @param channelId the identity of this channel
+ * @param name the name of the channel
  */
 
-export async function getOrCreateAppChannel(channelId: ChannelId): Promise<AppChannel> {
-    const channelTransport = await tryServiceDispatch(APIFromClientTopic.GET_OR_CREATE_APP_CHANNEL, {id: parseChannelId(channelId)});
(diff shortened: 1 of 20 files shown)


bryangaleopenfin:
@pjbroadbent - Could you take a look at the comments in ` src/client/contextChannels.ts`, specifically the instructions I've given about choosing channel names?

src/client/contextChannels.ts
```diff
@@ -443,6 +464,24 @@ export async function getCurrentChannel(identity?: Identity): Promise<Channel> {
...
+ * It is up to your organization to decide how to share knowledge of these custom channels. It is recommended that
+ * channel names are prefixed with your organization's domain name to avoid name collisions.
+ *
+ * The service will assign a unique ID when creating a new app channel, but no particular mapping of name to ID should
+ * be assumed.
+ *
+ * @param name The name of the channel
+ */
+
+export async function getOrCreateAppChannel(name: string): Promise<AppChannel> {
```
ottah, line 481: Should this not be split into 2 functions `get` & `create` (or just add these 2)? If I want to see if a channel exists calling `getOrCreateAppChannel` will create the channel which is quite limiting.

ottah: On a second look at this, is this to avoid collisions by different organisations/vendors trying to create the same channel with the same name?

src/client/validation.ts
```diff
@@ -64,12 +64,23 @@ export function parseContext(context: Context): Context {
...
     }
 
     return channelId;
 }
 
+/**
+ * Validates the provided app channel name
+ */
+export function parseAppChannelName(name: string): ChannelId {
+    if (typeof name !== 'string') {
```
ottah: Are empty strings `''` allowed as a channel name?

src/client/contextChannels.ts
```diff
@@ -391,10 +412,10 @@ export const DEFAULT_CHANNEL_ID: ChannelId = 'default';
 /**
  * The channel in which all windows will initially be placed.
  *
- * All windows will belong to exactly one channel at all times. If they have not explicitly
- * been placed into a channel via a {@link ChannelBase.join} call, they will be in this channel.
+ * All windows will belong to exactly one channel at all times. If they have not explicitly been placed into a channel
+ * via a {@link ChannelBase.join} call, they will be in this channel.
  *
- * If an app wishes to leave a system channel it can do so by (re-)joining this channel.
+ * If an app wishes to leave a system/app channel it can do so by (re-)joining this channel.
```
ottah: > If an app wishes to leave a system/app channel

> If an app wishes to leave a channel

bryangaleopenfin: I put it as "system/app channel" rather than just" channel" as otherwise it's inaccurate if you're already in the default channel.

Would "If an app wishes to leave any other channel, it can do so by (re-)joining this channel." be an improvement?

src/client/contextChannels.ts
bryangaleopenfin: The idea is that, let's say FinTechCo have a suite of apps that all use "com.FinTechCo-taupe", "com.FinTechCo-magnolia", "com.FinTechCo-salmon", no single app can rely on another one of those apps having created the channel. It's whoever get there first that creates it, so we either provider `getOrCreate`, or the most common use-case of a seperate `get` and `create` is going to be emulating `getOrCreate` (with added atomicity problems).

It might be too limiting eventually, but this has come out of Phil's convos with FactSet, and distinct `get`/`create`s are something we can always add if there's demand

src/client/validation.ts
bryangaleopenfin: Good question! I'm gonna say no. Will change

bryangaleopenfin: ✅

src/client/contextChannels.ts
bryangaleopenfin: Though realised typing this out, the comment should be "*reverse* domain name". Fixed

src/client/contextChannels.ts
ottah: Yeah that sounds better.

src/client/contextChannels.ts
ottah: That makes sense, in that case only a `hasAppChannel` would need to be added in future.

src/client/contextChannels.ts
bryangaleopenfin: ✅

src/provider/model/ContextChannel.ts
```diff
@@ -86,3 +86,38 @@ export class SystemContextChannel extends ContextChannelBase {
...
+
+    public constructor(id: ChannelId, name: string) {
+        super(id, 'app');
+
+        this.name = name;
+
+        this._context = null;
+    }
+
+    public getStoredContext(): Context | null {
```
noyangunday: is there a specific reason why we don't use a getter for this?

src/provider/model/ContextChannel.ts
```diff
@@ -86,3 +86,38 @@ export class SystemContextChannel extends ContextChannelBase {
...
+        return this._context;
+    }
+
+    public setLastBroadcastContext(context: Context) {
```
noyangunday: `setLastBroadcastContext`, `clearStoredContext`, `getStoredContext` methods and `_context` property feel like they will be relevant to (and identical in) every subclass extension apart from `DefaultContextChannel` from this point on. Is this a fair assumption?

Would it be cleaner to define all these behaviour in `ContextChannelBase` and make `DefaultContextChannel` a special case where you override the behaviour?

bryangaleopenfin: Not keen on overrides - (having to reason that the thing you're stopping being invoked is ok not to invoke, and in this case there'd be unused state on the default channel) but have moved shared behaviour into a base class.

src/provider/model/ContextChannel.ts
bryangaleopenfin: Just vague intuition that it felt more appropriate, given the other methods on the class and that it's not quite as straightforward as a variable being read.

Have given changing it at try. No strong feeling, happy to go either way on this

ngunday (approved):

ottah (approved):

PR #146, opened 2019-10-02
[SERVICE-617] Allow custom app uuid/name alias field within app directory

ngunday:
- Replaces manifestUrl-based mapping when an app is created with `appId`->`uuid` mappings.
- Provides the logic to use `appUuid` as `customConfig` field for cases where `appId` cannot be used to map `uuid`.


bryangaleopenfin:
Draft review: Looks good, though a couple of things that jump out.

- Understand the changes in `sample-app-directory` means the new code path does get exercised, but can this be explicitly tested? I _think_ a near-identical test just below the first test in `open.inttest.ts`, but with an app that uses the fallback customConfig would cover it

- I think we can collapse `windowSeen` and `windowCreated` into a single signal. We no longer need the call to `getInfo` in `FinEnvironment`, so there's no gap between the window first being seen by the service, and having enough info for the model to start registering it

ngunday:
> * I think we can collapse `windowSeen` and `windowCreated` into a single signal. We no longer need the call to `getInfo` in `FinEnvironment`, so there's no gap between the window first being seen by the service, and having enough info for the model to start registering it

working on this

src/provider/model/FinEnvironment.ts
```diff
@@ -162,7 +162,7 @@ export class FinEnvironment extends AsyncInit implements Environment {
         this.windowSeen.emit(identity);
 
         const info = await fin.Application.wrapSync(identity).getInfo();
-        this.windowCreated.emit(identity, info.manifestUrl);
+        this.windowCreated.emit(identity);
```
ottah: `info` can be removed.

ngunday: done

src/provider/model/AppDirectory.ts
```diff
@@ -31,7 +33,7 @@ export class AppDirectory extends AsyncInit {
 
     public async getAppByName(name: AppName): Promise<Application | null> {
         return this._directory.find((app: Application) => {
-            return app.name === name;
+            return (app.name === name || checkCustomConfigField(app, CustomConfigFields.OPENFIN_APP_NAME) === name);
```
bryangaleopenfin: Do we definitely want to allow a custom field in this case? I can see the logic in the UUID case - a developer for some reason can't have their app UUID match the directory appId, but I don't see what the use case is here.

src/provider/model/Model.ts
```diff
@@ -175,13 +175,11 @@ export class Model {
...
-    private onWindowSeen(identity: Identity): void {
-        this.getOrCreateExpectedWindow(identity);
-    }
-
-    private async onWindowCreated(identity: Identity, manifestUrl: string): Promise<void> {
+    private async onWindowSeen(identity: Identity): Promise<void> {
         const apps = await this._directory.getAllApps();
-        const appInfoFromDirectory = apps.find(app => app.manifest.startsWith(manifestUrl));
+        const appInfoFromDirectory = apps.find((app) => {
+            return app.appId === identity.uuid || checkCustomConfigField(app, CustomConfigFields.OPENFIN_APP_UUID) === identity.uuid;
```
bryangaleopenfin: Do we definitely want to check both `app.appId` and the custom fields. My intuition is that if a directory entry has the `appUuid` custom field, then we should ignore the `app.appId`. As it is, there's potentially two different apps that would get matched to this directory entry - the one that matches `app.appId`, and the one that matches the custom field.

Could do as something like 

    return checkCustomConfigField(app, CustomConfigFields.OPENFIN_APP_UUID, app.appId) === identity.uuid;`

where `checkCustomConfigField` (would need renaming I think) uses the custom field if available, or just returns the fallback third param we pass in.

bryangaleopenfin: Thoughts on moving this logic into `AppDirectory`?

src/provider/model/Model.ts
```diff
@@ -175,13 +175,11 @@ export class Model {
         }, []);
     }
 
-    private onWindowSeen(identity: Identity): void {
-        this.getOrCreateExpectedWindow(identity);
```
bryangaleopenfin, line 179: I think we want to keep this call as the first thing the remaining seen/created listener does

test/provider/expect.unittest.ts
```diff
@@ -51,7 +51,7 @@ type TestParam = [
 const FAKE_TEST_DURATION = 10000;
 
 const REGISTRATION_TIMEOUT = 5000;
-const PENDING_TIMEOUT = 250;
+const PENDING_TIMEOUT = 100;
```
bryangaleopenfin: This should be the same value as `Timeouts.WINDOW_EXPECT_TO_SEEN` - read as "how soon after the window is expected should the promise reject, if the window is never created".

I didn't simply import Timeouts originally, as then I felt like I was leaning on part of implementation of the thing I was testing, but maybe it should just be imported 🤷‍♀️. Or just a comment here?

test/provider/expect.unittest.ts
```diff
@@ -282,7 +282,6 @@ function expectTest(testWindow: TestWindow, appDirectoryResultTime: number, resu
         });
 
         maybeSetTimeout(() => mockEnvironment.windowSeen.emit(identity), testWindow.seenTime);
-        maybeSetTimeout(() => mockEnvironment.windowCreated.emit(identity, manifestUrl), testWindow.createdTime);
```
bryangaleopenfin, line 285: You can get rid of `createdTime` from `TestWindow`

bryangaleopenfin (commented):
Some general things
- Do a pass to completely remove "seen" terminology, and use "created" everywhere instead
- The `expect` unit tests weren't working for me. They are kinda byzantine, so shout if you want me to have a go explaining them

src/provider/model/Model.ts
noyangunday: makes sense

src/provider/model/AppDirectory.ts
noyangunday: Ok, that was actually my original idea, but then got probably a bit sidetracked by this comment:

```
Understand the changes in sample-app-directory means the new code path does get
exercised, but can this be explicitly tested? I think a near-identical test just below the first
test in open.inttest.ts, but with an app that uses the fallback customConfig would cover it
```

the first test in open.inttest.ts tests the name field indirectly since `open` makes a call to `getAppByName`

noyangunday: I will remove this and add a unit test for `AppDirectory ` to test the `customConfig` entries.

src/provider/model/Model.ts
```diff
@@ -18,13 +19,13 @@ import {Environment, EntityType} from './Environment';
...
-    // Resolves when the window has been seen by the environment. Resolves to the `registered` promise wrapped in a timeout
-    seen: Promise<Boxed<Promise<AppWindow>>>;
+    // Resolves when the window has been created by the environment. Resolves to the `registered` promise wrapped in a timeout
+    created: Promise<Boxed<Promise<AppWindow>>>;
 
     // Resolves when the window has connected to FDC3
     connected: Promise<void>;
 
-    // Resolves to the AppWindow when the window has been fully registered and is ready for use outside the Model
+    // Resolves to the AppWindow when the window has been fully regstered and is ready for use outside the Model
```
bryangaleopenfin: Typo

src/provider/model/AppDirectory.ts
```diff
@@ -35,6 +37,12 @@ export class AppDirectory extends AsyncInit {
         }) || null;
     }
 
+    public async getAppByUuid(uuid: string): Promise<Application | null> {
+        return this._directory.find((app: Application) => {
+            return checkCustomConfigField(app, CustomConfigFields.OPENFIN_APP_UUID) === uuid || app.appId === uuid;
```
bryangaleopenfin: It's still the case that there's potentially two input `uuids` that'll give the same app.

My claim is that if an appId is specified in the custom data, the other appId should be ignored entirely here. So something like

    const customValue = checkCustomConfigField(app, CustomConfigFields.OPENFIN_APP_UUID);
    return (customValue !== undefined)  ? (customValue === uuid) : (app.appId === uuid)

would work. And I think it'd be worth explicitly testing that if a directory entry specifies a UUID as a custom data field, appId gets igored by `getAppByUuid`

src/provider/model/Model.ts
```diff
@@ -175,13 +175,9 @@ export class Model {
         }, []);
     }
 
-    private onWindowSeen(identity: Identity): void {
+    private async onWindowCreated(identity: Identity): Promise<void> {
         this.getOrCreateExpectedWindow(identity);
```
bryangaleopenfin: Can store the result and resuse, rather than calling the same function again a few lines down

bryangaleopenfin:
I think I see why the `expect` unit tests are failing - the test is mocking `getAllApps`, where now it wants to be mocking `getAppByUuid`, since that's what's use in `onWindowCreated`

Also suggest change the "990"s to "1000"s, to be less mysterious - it was "990" before to indicate 'seen' happening slightly before 'create', but now we only have one of those, it can be a nicer value.

ngunday:
> I think I see why the `expect` unit tests are failing - the test is mocking `getAllApps`, where now it wants to be mocking `getAppByUuid`, since that's what's use in `onWindowCreated`
> 
> Also suggest change the "990"s to "1000"s, to be less mysterious - it was "990" before to indicate 'seen' happening slightly before 'create', but now we only have one of those, it can be a nicer value.

done

src/provider/model/FinEnvironment.ts
```diff
@@ -59,7 +46,7 @@ export class FinEnvironment extends AsyncInit implements Environment {
     public readonly windowClosed: Signal<[Identity]> = new Signal();
 
     private _windowsCreated: number = 0;
-    private readonly _seenWindows: {[id: string]: SeenWindow} = {};
+    private readonly _createdWindows: {[id: string]: CreatedWindow} = {};
```
ottah: Might as well change this to a `Map`.

bryangaleopenfin (approved):

src/provider/model/Model.ts
```diff
@@ -7,9 +7,10 @@ import {Inject} from '../common/Injectables';
 import {ChannelId, DEFAULT_CHANNEL_ID} from '../../client/main';
 import {APIHandler} from '../APIHandler';
 import {APIFromClientTopic} from '../../client/internal';
-import {SYSTEM_CHANNELS, Timeouts} from '../constants';
+import {SYSTEM_CHANNELS, Timeouts, CustomConfigFields} from '../constants';
 import {withStrictTimeout, untilTrue, allowReject, untilSignal} from '../utils/async';
 import {Boxed} from '../utils/types';
+import {checkCustomConfigField} from '../utils/helpers';
```
bryangaleopenfin: Sorry, one last thing! Just some unused imports in this file

ngunday: good catch thanks

src/provider/model/FinEnvironment.ts
ngunday: done

ottah (approved):

PR #149, opened 2019-10-03
[N/A] Demo app updates - split instruments into 'name' and 'ticker'

pjbroadbent:
Made instruments slightly more "realistic", removed shortcut from teal charts app.

### DIFF
--- a/src/demo/apps/BlotterApp.tsx
+++ b/src/demo/apps/BlotterApp.tsx
@@ -6,24 +6,25 @@ import {ContextChannelSelector} from '../components/ContextChannelSelector/Conte
 
 export interface Symbol {
     name: string;
+    ticker: string;
 }
 
 const symbols: Symbol[] = [
-    {'name': 'AAPL'},
-    {'name': 'AMD'},
-    {'name': 'AMZN'},
-    {'name': 'CMCSA'},
-    {'name': 'CSCO'},
-    {'name': 'FB'},
-    {'name': 'GOOG'},
-    {'name': 'INTC'},
-    {'name': 'MSFT'},
-    {'name': 'NFLX'},
-    {'name': 'NVDA'},
-    {'name': 'PYPL'},
-    {'name': 'QCOM'},
-    {'name': 'ROKU'},
-    {'name': 'TSLA'}
+    {ticker: 'AAPL', name: 'Apple'},
+    {ticker: 'AMD', name: 'Advanced Micro Devices'},
+    {ticker: 'AMZN', name: 'Amazon.com'},
+    {ticker: 'CMCSA', name: 'Comcast Corporation'},
+    {ticker: 'CSCO', name: 'Cisco Systems'},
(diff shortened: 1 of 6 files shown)


src/demo/components/blotter/SymbolsRow.tsx
```diff
@@ -83,7 +83,8 @@ export function SymbolsRow(props: SymbolsRowProps): React.ReactElement {
             type: 'fdc3.instrument',
             name: item.name,
             id: {
-                default: item.name
+                ticker: item.ticker,
+                default: item.ticker
```
tomrobertsOF: If we're changing it to use `ticker` do we still need `default`? That was just a leftover thing from an old version of the spec right? I don't see it on the fdc3 spec site anymore.

tomrobertsOF (approved):

src/demo/apps/BlotterApp.tsx
```diff
@@ -6,24 +6,25 @@ import {ContextChannelSelector} from '../components/ContextChannelSelector/Conte
...
-    {'name': 'GOOG'},
-    {'name': 'INTC'},
-    {'name': 'MSFT'},
-    {'name': 'NFLX'},
-    {'name': 'NVDA'},
-    {'name': 'PYPL'},
-    {'name': 'QCOM'},
-    {'name': 'ROKU'},
-    {'name': 'TSLA'}
+    {ticker: 'AAPL', name: 'Apple'},
```
bryangaleopenfin, line 13: "Apple", or "Apple Inc" as elsewhere?

src/demo/apps/ChartsApp.tsx
```diff
@@ -12,11 +12,15 @@ interface AppProps {
 }
 
 export function ChartsApp(props: AppProps): React.ReactElement {
-    const [symbolName, setSymbolName] = React.useState('AAPL');
+    const [symbolName, setSymbolName] = React.useState('Apple Inc (AAPL)');
```
bryangaleopenfin: Think symbolDisplayName would be clearer now (ditto in `NewsApp,tsx`)

src/demo/apps/BlotterApp.tsx
pjbroadbent: Now normalised, but went with "Apple" - to normalise the other way would mean adding "Inc" to most of the other symbols too, which resulted in some rather long titles and the duplication looked a bit weird in the blotter view.

src/demo/apps/ChartsApp.tsx
pjbroadbent: Went with a slightly more generic `title`/`setTitle`

src/demo/components/blotter/SymbolsRow.tsx
pjbroadbent: Hmm, didn't realise that got removed. Yeah, let's do that.

bryangaleopenfin (approved):

PR #150, opened 2019-10-03
[DRAFT] [SERVICE-553, SERVICE-613?] Add handshake to context listeners

tomrobertsOF:
I think it is worth having a draft PR up to discuss the general approach I took to this story and allow people the opportunity to point out any edge cases I might have missed beforeI go deeper into cleanup/testing.

This started out as a branch purely for SERVICE-553, but I found in doing it that SERVICE-613 came more or less free as a side effect of the changes (unless I've missed some subtleties, which is possible, if not likely).

I mostly duplicated the approach taken for `IntentListener`s with the main differences being that there is no distinction between various listeners, allowing some simplifications.

Some of the noteworthy changes are:
- AppWindow now has a private member `_contextListener` similar to `_contextListeners`, but is a `boolean` instead of a `Set`
- Added duals the other `intentListener` related methods for contexts, e.g. add/remove/has
- The async method `isReadyToRecieveContext` follows the same pattern as the intent version but has its own timeout constant which, for now, is set to the same duration. I thought there was value in having these independently changeable.


tomrobertsOF:
Had some new ideas about this. Will close for now, and reopen once I'm done tinkering.

PR #153, opened 2019-10-03
[N/A] Fix potential memory leak

bryangaleopenfin:
Fixes issue discovered while working on SERVICE-552

There are circumstances where listeners can be added to `Model`'s signals, but never removed (eg., a window that opens and closes, but never connects to the service). This fixes that issue


tomrobertsOF (approved):
Really good catch! Hadn't even considered that case. 

I'm always a bit wary of optional parameter proliferation, but willing to overlook it here.

rypete (approved):
Looks good

PR #154, opened 2019-10-03
[SERVICE-737] Interim .NET app support

pjbroadbent:
First-pass of adding support for non-JS applications.

Also includes some further encapsulation of `APIHandler`.


src/provider/model/Model.ts
```diff
@@ -91,6 +92,12 @@ export class Model {
     }
 
     public async expectWindow(identity: Identity): Promise<AppWindow> {
+        // Temporary work-around until .NET channel aliases are also applied to 'dispatch' calls, same as they are on channel connection events.
+        // Note that 'secondWindow' in .NET demo will not fully work whilst this workaround is in place.
+        if (identity.name && identity.name.split('-').length === 5) {
+            identity.name = 'mainwin';
+        }
```
pjbroadbent: Will remove this before marking PR as ready, currently still doing some final testing - which requires this workaround.

src/provider/APIHandler.ts
```diff
@@ -81,6 +77,24 @@ export class APIHandler<T extends Enum> {
...
+            throw error;
+        });
+    }
+
+    public publish(action: string, payload: any): Promise<any>[] {
+        const connections = this._providerChannel.connections.slice();
+        return this._providerChannel.publish(action, payload).map((promise, index) => promise.catch(error => {
+            // Log and re-throw
+            // We don't know which connection had the error, but assume that the indices of the promises match the indices of the channel connections.
+            console.error(`Error when publishing '${action}' (probably from connection ${index + 1}/${connections.length}: ${connections[index].uuid} / ${connections[index].name})\n`, payload);
```
bryangaleopenfin: How did the linter allow a line this long?

src/provider/model/FinEnvironment.ts
```diff
@@ -117,6 +118,12 @@ export class FinEnvironment extends AsyncInit implements Environment {
         }
     }
 
+    public async getEntityType(identity: Identity): Promise<EntityType> {
+        const entityInfo = await fin.System.getEntityInfo(identity.uuid, identity.uuid);
```
bryangaleopenfin: Is this definitely correct? Not `await fin.System.getEntityInfo(identity.uuid, identity.name);`. Suggest comment if correct, definitely looks like an error

src/provider/model/FinEnvironment.ts
```diff
@@ -77,7 +77,8 @@ export class FinEnvironment extends AsyncInit implements Environment {
         identity = parseIdentity(identity);
         const id = getId(identity);
 
-        const {creationTime, index} = this._seenWindows[id];
+        const seenWindow = this._seenWindows[id] || {creationTime: 0, index: this._windowsCreated++};
```
bryangaleopenfin: Would comment need for this. And would it make more sense for creation time to be time of connection?

src/provider/model/Model.ts
```diff
@@ -208,6 +215,13 @@ export class Model {
         }
     }
 
+    private async onApiHandlerConnection(identity: Identity): Promise<void> {
+        if (await this._environment.getEntityType(identity) === EntityType.EXTERNAL_CONNECTION) {
```
bryangaleopenfin, line 227: Since we won't ever get a window closed event for .NET connections (right?), de-register on disconnect?

And I'm not convinced things will work right with the expect system? E.g., we don't get a seen event for  these windows, and the close promise will never reject

bryangaleopenfin (commented):
Any way of doing integration tests for this? There's not a nice way to fake a .NET app for an integration test is there?

Could unit test with a mocked fin.System.getEntityInfo though?

src/provider/model/Model.ts
bryangaleopenfin: (And if you're fiddling around there, expect merge conflicts - sorry!)

ngunday, line 227: is a .net application guaranteed to be EntityType.EXTERNAL_CONNECTION? what happens to the iframes inside the .net apps? I couldn't find the full description of getEntityInfo() behaviour in api docs.

src/provider/APIHandler.ts
pjbroadbent: Hmm, not sure how I managed that :-)

I was playing around with some tooling stuff separately (though nothing that affected the linter...), may have something to do with it.

Now seeing the linter complain though, will split into two lines.

src/provider/APIHandler.ts
```diff
@@ -81,6 +77,24 @@ export class APIHandler<T extends Enum> {
...
+            throw error;
```
ngunday: alternatively (maybe a bit uglier but should save you from that slice() and the assumption:
`const catchHandler = ((uuid, name, error) => { <catch body goes here> }).bind(undefined, connectionUuid, connectionUuid)`
then
`promise.catch(catchHandler);`

ngunday: well, then I realised that would mean implementing your own publish :) ignore that.

src/provider/APIHandler.ts
ngunday: that looks like a `-n` to me :P

src/provider/model/Model.ts
pjbroadbent: I'll defer to the second phase of this story, if there's no objections?

I'd want to find a way to store the entity type on the "window" object itself, so that we can keep the disconnect handler synchronous. Since some key parts of this are in flux due to other PR, maybe makes sense for those changes to be finalised first?

pjbroadbent: Raised this with #dc-adapters people - yeah, seems that OpenFin iframes and .NET embedded views will both be identified as `"iframe"`.

Will need to think about the implications of this - we may even need a test .NET app before we can solve properly. Will add to the second part of this story, or log separately in the backlog.

ngunday: alright

ngunday (approved):

bryangaleopenfin:
Could you put in appropriate todos for anything that's not going to go in with this PR. Even if you're creating an empty test file with a comment/`.todo` test

src/provider/model/Model.ts
bryangaleopenfin: No objection, but put in TODOs

bryangaleopenfin (approved):

PR #155, opened 2019-10-04
[SERVICE-552] Resolve unnecessary inconsistency between how we handle “directory” and “ad-hoc” apps

bryangaleopenfin:
This PR:
- Changes the registration process so that both directory and non-directory app windows are not registered until they connect to FDC3
- Adds an uses new methods `expectWindowsForApp` and `ensureRunning` to `Model`.
- Adds and uses a test app that is in the directory, but never connects to FDC3

I don't like the knock-on changes around "open" and "raiseIntent" calls, but I don't think they're making anything worse, just making existing problems more obvious. I think these problems are best addressed in SERVICE-556


src/provider/model/FinEnvironment.ts
```diff
@@ -61,6 +61,15 @@ export class FinEnvironment extends AsyncInit implements Environment {
     private _windowsCreated: number = 0;
     private readonly _seenWindows: {[id: string]: SeenWindow} = {};
 
+    public async isRunning(appInfo: Application): Promise<boolean> {
+        // TODO: This can be simplified once we change how we map between OpenFin apps and directory entries [SERVICE-617]
+        const runningApplications = await Promise.all((await fin.System.getAllApplications())
+            .filter(info => info.isRunning)
```
tomrobertsOF: Does `getAllApplication` ever return non-running apps?

src/provider/controller/IntentHandler.ts
```diff
@@ -109,8 +109,12 @@ export class IntentHandler {
     }
 
     private async fireIntent(intent: Intent, appInfo: Application): Promise<IntentResolution> {
-        const appWindows = await this._model.findOrCreate(appInfo);
-        // to decide between focus nothing or apps with intent listener
+        await this._model.ensureRunning(appInfo);
+
+        // TODO: This will never resolve if no windows of this app connect to FDC3 [SERVICE-556]
```
tomrobertsOF: How concerned should we be about this possibility before 556 goes in? What affects will it have on the rest of the service?

Also the same question for the similar comments in other files.

src/provider/model/FinEnvironment.ts
```diff
@@ -61,6 +61,15 @@ export class FinEnvironment extends AsyncInit implements Environment {
     private _windowsCreated: number = 0;
     private readonly _seenWindows: {[id: string]: SeenWindow} = {};
 
+    public async isRunning(appInfo: Application): Promise<boolean> {
+        // TODO: This can be simplified once we change how we map between OpenFin apps and directory entries [SERVICE-617]
```
tomrobertsOF: 617 is currently in review. It's probably worth liaising with Noyan to make sure this gets resolved either here or there so we don't end up with a dangling `TODO`

src/provider/model/FinEnvironment.ts
bryangaleopenfin: If I had to guess, probably not, but since the docs are silent on this, and since `isRunning` is specifically returned (and this isn't a shape that's used elsewhere - this is a different `ApplicationInfo` to the one returned by `app.getInfo()`), this feels like worthwhile safety

tomrobertsOF: Fair enough. I guess since it doesn't require another call into the core to check it's not much of a difference to keep it.

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Ah, realised my " I don't think they're making anything worse" line was wrong. Have revised these so there's an explicit timeout around the `expectWindowsForApp` call

tomrobertsOF: Cool. I'd say that's good enough for now. As long as we aren't going to end up deadlocked.

test/demo/connection.inttest.ts
```diff
@@ -12,76 +12,133 @@ import {IntentHandler} from '../../src/provider/controller/IntentHandler';
...
 
-type TestParam = [
+type ConnectTestParam = [
+    string,
+    TestAppData,
+    RegistrationStatus,
+    (app: TestAppData) => Promise<void>,
+];
+
+type ConnectTestCatagoryParam = [string, ConnectTestParam[]];
```
rypete: `Category` not `Catagory`. Same with all variants found throughout this file.

test/demo/connection.inttest.ts
```diff
@@ -171,3 +232,9 @@ async function hasEventListeners(identity: Identity, eventType: ChannelEvents['t
     }, eventType);
     return identities.some(id => id.name === identity.name && id.uuid === identity.uuid);
 }
+
+async function isRegistered(identity: Identity): Promise<boolean> {
```
rypete, line 236: Could this be useful as a util for any other tests?

src/provider/controller/IntentHandler.ts
```diff
@@ -109,21 +111,29 @@ export class IntentHandler {
...
+        dispatchResults = await withTimeout(Timeouts.ADD_INTENT_LISTENER, (async () => {
+            const appWindows = await this._model.expectWindowsForApp(appInfo);
+
+            // Wait for windows to add intent listener, then dispatch payload
+            return Promise.all(appWindows.map(async (window: AppWindow): Promise<boolean> => {
+                if (await window.isReadyToReceiveIntent(intent.type)) {
+                    const payload: ReceiveIntentPayload = {context: intent.context, intent: intent.type};
+
+                    // TODO: Implement a timeout so a misbehaving intent handler can't block the intent raiser [SERVICE-555]
+                    if (dispatchResults === undefined) {
```
rypete: Just to get my head around this one - When would dispatchResults be `undefined` if we are setting it be the promise this is called from?

test/demo/connection.inttest.ts
bryangaleopenfin: ✅

test/demo/connection.inttest.ts
bryangaleopenfin: My feeling is no.  This and the similar functions in this file are all kind cheeky in the context of integration tests. Rather than looking at the API output, we're looking at the internal state of the provider, which is dodgy, but kinda necessary in these tests. I wouldn't want to communicate "this is a thing we should be doing" by putting it in a common util file

src/provider/controller/IntentHandler.ts
bryangaleopenfin: It'll be undefined until the promise returned by withTimeout on line 118 resolves. This check is to prevent dispatching the intent happening after we've already reported that it's failed.

Though it feels kind weird and dodgy, have revised.

✅

test/demo/connection.inttest.ts
tomrobertsOF: I'm with Bryan on this one. Integration tests should really only be looking at the state on either end of some action, not really concerned with internal provider state.

rypete (approved):

src/provider/model/FinEnvironment.ts
bryangaleopenfin: TODO resolved now that 617 is in

✅

tomrobertsOF (approved):

PR #156, opened 2019-10-07
[N/A] Move getId out of Model.ts to a util file

ottah:
Small patch to fix demo apps not working due to a change in #144 
Previously getId was being imported by the demo apps causing them to import the whole injector.

### DIFF
--- a/src/provider/model/Model.ts
+++ b/src/provider/model/Model.ts
@@ -80,7 +80,7 @@ export class Model {
         return Object.values(this._channelsById);
     }
 
-    public getWindow(identity: Identity): AppWindow | null {
+    public getWindow(identity: Identity): AppWindow|null {
         return this._windowsById[getId(identity)] || null;
     }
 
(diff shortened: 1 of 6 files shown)


src/provider/model/Model.ts
```diff
@@ -87,7 +80,7 @@ export class Model {
         return Object.values(this._channelsById);
     }
 
-    public getWindow(identity: Identity): AppWindow|null {
+    public getWindow(identity: Identity): AppWindow | null {
```
rypete: Is this linted? I think we enforce `a|b`

rypete (changes requested):

src/provider/model/Model.ts
ottah: Aren't we changing to `a | b` anyway with the new linting rules? I'll change it back since my VSCode auto formatted it.

bryangaleopenfin (approved):
Woops! Sorry, approved!

src/provider/model/Model.ts
rypete: Maybe, but it will cause problems until then.

rypete (approved):

PR #157, opened 2019-10-09
[SERVICE-553] Add handshake to context listeners 

tomrobertsOF:
I mostly duplicated the approach taken for IntentListeners with the main differences being that there is no distinction between various listeners, allowing some simplifications.

Some of the noteworthy changes are:

- AppWindow now has a private member _contextListener similar to _contextListeners, but is a boolean instead of a Set
- Added duals the other intentListener related methods for contexts, e.g. add/remove/has
- The async method isReadyToRecieveContext follows the same pattern as the intent version but has its own timeout constant which, for now, is set to the same duration. I thought there was value in having these independently changeable.

I also decided to refactor AppWindow slightly to make it more testable. All of the non-openfin-specific code no lives in an abstract base class which is extended by `FinAppWindow` with the fin-specific calls for the real service, and  by `TestAppWindow` with stubbed out fin calls for unit testing. In this PR I've added unit tests covering the parts of `AppWindow` that I've added/changed in story, but left it for future stories to backfill testing for any other functionality


src/client/internal.ts
```diff
@@ -217,6 +221,14 @@ export interface RemoveIntentListenerPayload {
     intent: string;
 }
 
+export interface addContextListenerPayload {
```
tomrobertsOF: There is precedent for empty payload interfaces earlier in the file. Figured this would make things easier if we ever need to add any new properties here.

src/provider/constants.ts
```diff
@@ -7,6 +7,11 @@ export const Timeouts = {
      */
     ADD_INTENT_LISTENER: 5000,
 
+    /**
+     * Time for an app to register a listener after opening
+     */
+    ADD_CONTEXT_LISTENER: 5000,
```
tomrobertsOF: Set this to the same as Intents, but no objection to changing if people have other opinions.

test/demo/utils/globalSetup.ts
```diff
@@ -5,14 +5,29 @@ import * as path from 'path';
...
+import * as jsAdapter from 'hadouken-js-adapter';
 
 declare const global: NodeJS.Global&{__BROWSER_GLOBAL__: puppeteer.Browser};
 
 const DIR = path.join(os.tmpdir(), 'jest_puppeteer_global_setup');
 
 module.exports = async () => {
-    // Openfin should be running at this point
-    const browserWSEndpoint = (await (await fetch('http://localhost:9222/json/version')).json()).webSocketDebuggerUrl;
+    // Get the devtools port of the currently running openfin instance
```
tomrobertsOF, line 15: I figured I would pull these improvements in from Notifications, since it seems to improve the connection errors at least a bit.

src/provider/controller/ContextHandler.ts
```diff
@@ -31,8 +31,11 @@ export class ContextHandler {
      */
     public async send(window: AppWindow, context: Context): Promise<void> {
         const payload: ReceiveContextPayload = {context};
-
-        return this._apiHandler.dispatch(window.identity, APIToClientTopic.RECEIVE_CONTEXT, payload);
+        if (await window.isReadyToReceiveContext()) {
+            return this._apiHandler.dispatch(window.identity, APIToClientTopic.RECEIVE_CONTEXT, payload);
```
ottah, line 36: Add a timeout for the dispatch or mark with a TODO SERVICE-555.

src/provider/model/FinAppWindow.ts
```diff
@@ -0,0 +1,28 @@
+import {Identity, Window} from 'openfin/_v2/main';
+
+import {Application} from '../../client/main';
+
+import {AppWindow, AbstractAppWindow} from './AppWindow';
```
ottah: `AppWindow` is unused.

test/demo/open.inttest.ts
```diff
@@ -11,6 +11,7 @@ import {
     testManagerIdentity, testAppInDirectory1, testAppInDirectory2,
     testAppWithPreregisteredListeners1, testAppWithPreregisteredListeners2
 } from './constants';
+import {delay, Duration} from './utils/delay';
```
ottah: Duration is unused.

test/provider/AppWindow.unittest.ts
```diff
@@ -0,0 +1,164 @@
...
+    name: 'Test App',
+    manifest: 'test.manifest',
+    manifestType: 'openfin'
+};
+
+beforeEach(() => {
+    jest.resetAllMocks();
+});
+
+describe('When querying if a window has a context listener', () => {
```
ottah, line 42: As file is just simple generic `AppWindow` tests might aswell add in some intent tests aswell.

test/provider/ContextHandler.unittest.ts
```diff
@@ -25,7 +25,15 @@ let mockGetChannelMembers: jest.Mock<AppWindow[], [ContextChannel]>;
...
 
 function createMockAppWindowWithName(name: string): AppWindow {
-    return {...createMockAppWindow(), identity: {uuid: 'test', name}};
+    const mockedWindow = createMockAppWindow();
+    mockedWindow.isReadyToReceiveContext.mockResolvedValue(true);
+    return {...mockedWindow, identity: {uuid: 'test', name}};
+}
+
+function createMockAppWindowNotReadyWithName(name: string): AppWindow {
+    const mockedWindow = createMockAppWindow();
```
ottah: Can `createMockAppWindow()` be changed to take a `Partial<AppWindow>` and just:
```ts
export function createMockAppWindow(options: Partial<AppWindow> = {}): jest.Mocked<AppWindow> {
    return {
        id: '',
        // existing...
        ...options
    };
}
```

Then we wont need to create all these types of functions.

ottah (commented):
Looks good, just a bit of cleanup.

test/provider/AppWindow.unittest.ts
tomrobertsOF: I explained that in the description: 
> In this PR I've added unit tests covering the parts of AppWindow that I've added/changed in story, but left it for future stories to backfill testing for any other functionality

My intention is definitely that this file be used for a much larger set of tests of all the relevant behaviours of the class, but I would say that falls under another story.

src/provider/controller/ContextHandler.ts
tomrobertsOF: 👍 on the TODO

tomrobertsOF: Also, unlikely to be as big of a problem in this case, since we don't expect any kind of response here, but will tag it anyway as its better to be safe.

test/demo/open.inttest.ts
tomrobertsOF: 👍

test/provider/ContextHandler.unittest.ts
tomrobertsOF: I'd rather not. This mock is used in instances where we need to pass an `AppWindow` into some dependent module. There is no guarantee that it will only ever use those properties, so I think it makes sense to expose a fully mocked version that implements the whole interface.

src/provider/model/FinAppWindow.ts
tomrobertsOF: 👍

test/provider/ContextHandler.unittest.ts
ottah: `// existing...` was meant to be shortcut for what is currently there in `createMockAppWindow`. It would be the same as it is now, but allow overrides for to be passed in via `options`.

tomrobertsOF: I may have misread this a bit. Will circle back in a bit once I'm clear what's being asked.

tomrobertsOF: Ah, ok. Think I get it. That does make decent sense. I mostly did it like this because that's how the original helper was. It felt a bit gross having the giant, very specific function name, but its hardly the first one like that in the tests. 

I'll change it now.

tomrobertsOF:
@TheoOttahOF, ready for another look

src/client/internal.ts
bryangaleopenfin: Capitalize interface names

test/mocks.ts
```diff
@@ -31,7 +34,10 @@ export function createMockAppWindow(): jest.Mocked<AppWindow> {
         bringToFront: jest.fn<Promise<void>, []>(),
         focus: jest.fn<Promise<void>, []>(),
         isReadyToReceiveIntent: jest.fn<Promise<boolean>, [IntentType]>(),
-        removeAllListeners: jest.fn<void, []>()
+        isReadyToReceiveContext: jest.fn<Promise<boolean>, []>(),
+        removeAllListeners: jest.fn<void, []>(),
+        // Apply any custom overrides
+        ...options
```
bryangaleopenfin, line 40: Hmm. Could we do this everywhere or nowhere?

We've got a few instances of eg. `{...createMockChannel(), id: 'test-3', type: 'system'}` which takes the inverse approach.

src/provider/model/AppWindow.ts
```diff
@@ -53,5 +64,192 @@ export interface AppWindow {
...
+
+            const [didTimeout] = await withTimeout(Timeouts.ADD_INTENT_LISTENER - age, deferredPromise.promise);
+
+            slot.remove();
+
+            return !didTimeout;
+        }
+    }
+
+    public async isReadyToReceiveContext(): Promise<boolean> {
```
bryangaleopenfin, line 222: Can shared code be pulled out and used by both `isReadyToReceiveIntent` and `isReadyToReceiveContext`? or does it wind up being too messy to be worthwhile?

test/provider/AppWindow.unittest.ts
```diff
@@ -0,0 +1,164 @@
...
+        await Promise.resolve();
+
+        // Promise should have resolved immediately, so spy should have been invoked
+        expect(timingSpy).toHaveBeenCalledWith(true);
+    });
+
+    describe('When the window does not have a listener registered', () => {
+        test('If the window was created longer than the timeout in the past, the promise resolves false immediately', async () => {
+            // @ts-ignore Updating the creation time to test old windows
+            testAppWindow._creationTime = Date.now() - 10000;
```
bryangaleopenfin: Cleaner to fastforward faketime rather than rewind the window?

test/provider/AppWindow.unittest.ts
```diff
@@ -0,0 +1,164 @@
...
+            testAppWindow.isReadyToReceiveContext().then(timingSpy);
+
+            jest.advanceTimersByTime(1000);
+            await Promise.resolve();
+            expect(timingSpy).not.toHaveBeenCalled();
+
+            testAppWindow.addContextListener();
+
+            await Promise.resolve();
+            jest.advanceTimersByTime(5000);
```
bryangaleopenfin: Copy over the fake time code form notification, which wraps together advancing time and resolving promises?

test/demo/open.inttest.ts
```diff
@@ -184,6 +185,63 @@ and does not trigger the context listener of the already open app', async () =>
...
+            await quitApps(testAppDelayedPreregisterShort);
+        });
+
+        test('The promise resolves and the app opens', async () =>{
+            await openPromise;
+
+            await expect(fin.Application.wrapSync(testAppDelayedPreregisterShort).isRunning()).resolves.toBe(true);
+        });
+
+        test('the context is received by the listener', async () =>{
```
bryangaleopenfin: Captialization

test/demo/open.inttest.ts
```diff
@@ -184,6 +185,63 @@ and does not trigger the context listener of the already open app', async () =>
...
+        afterEach(async () => {
+            await quitApps(testAppDelayedPreregisterLong);
+        });
+
+        test('The promise resolves and the app opens', async () =>{
+            await openPromise;
+            await expect(fin.Application.wrapSync(testAppDelayedPreregisterLong).isRunning()).resolves.toBe(true);
+        });
+
+        test('the context is not received by the listener', async () =>{
```
bryangaleopenfin: Capitalization

ottah (approved):
Looks good :+1:

test/demo/utils/delay.ts
```diff
@@ -10,5 +10,6 @@ export async function delay(duration: number) {
 
 export enum Duration {
     PAGE_RELOAD = 500,
-    PAGE_NAVIGATE = 500
+    PAGE_NAVIGATE = 500,
+    LISTENER_TIMEOUT = 10000
```
bryangaleopenfin: Is timeout the correct term here? This is just the long preregister delay, right? Would have both this and the short delay here, or just hardcode both where used

src/client/internal.ts
tomrobertsOF: Whoops. That's why we need Ryan's linter changes. Will change

test/mocks.ts
tomrobertsOF: Sure. Makes sense to keep it consistent.

src/provider/model/AppWindow.ts
tomrobertsOF: Hmmm, maybe. My concern is that there are subtle differences pretty much throughout these functions, which would make whatever common function have several `if context` / `if intent` block, at which point I think we lose whatever value we gain from consolidating. Specifically, we would need to switch for the `has[x]Listener` function, the timeout constant, and the `on[x]ListenerAdded` signal, which to me pushes it into the "too messy" region.

test/provider/AppWindow.unittest.ts
tomrobertsOF: Yea, probably. I'm a bit wary of fake time and the ordering of when we create the window vs when we stop time, but I'll give I a go and see how it works.

test/provider/AppWindow.unittest.ts
tomrobertsOF: Ah, didn't know we had that. Will take a look.

test/demo/open.inttest.ts
tomrobertsOF: 👍

test/demo/open.inttest.ts
tomrobertsOF: 👍

test/demo/utils/delay.ts
tomrobertsOF: Sure. I wasn't really sure what to call it and probably rushed a bit getting it in. I'll go back and have a look.

test/provider/AppWindow.unittest.ts
tomrobertsOF: Your suggestion started working properly once I swapped in the fake time stuff from Notes.

tomrobertsOF:
@bryangaleOF, ready for a second look

test/provider/AppWindow.unittest.ts
bryangaleopenfin: There's a couple of leftover instances of `await Promise.resolve();` - one that can be replaced with `resolvePromiseChain`, and one that's commented one so I assume can just be deleted

test/provider/AppWindow.unittest.ts
bryangaleopenfin: Looks good, but commented out code still in place

tomrobertsOF:
@bryangaleOF ok, actually ready this time.

test/demo/utils/delay.ts
tomrobertsOF: Cleaned this up as discussed.

test/provider/AppWindow.unittest.ts
bryangaleopenfin: There's still line 107 using the raw `await Promise.resolve();`

tomrobertsOF: Ok, that should be fixed.

tomrobertsOF: And I searched through for any others and didn't see any.

bryangaleopenfin (approved):

PR #158, opened 2019-10-09
[SERVICE-750] Ensure consistency between`raiseIntent`, `findIntent`, and apps shown in resolver

bryangaleopenfin:
Draft PR for "is this the outward behaviour we want?" review. Lots of files changed, but the main ones to focus on for now are `AppDirectory.ts`, `Model.ts`, `IntentHandler.ts`, and `provider/index.ts`.

(This was branched off the SERVICE-552 PR, so will have some irrelevant changes)

All ways of querying/presenting what apps handle what intents/contexts now go through `_model.getApplicationsForIntent`.

The big knock-on effects that deviate from the doc agreed with Phil are:
- Raising an intent with a target isn't totally trusting that the user knows what they're doing, and will check that the app will support it
- The result of `findIntentsByContext` will be influenced by what's running/has registered intent listeners.

I've also had to put contexts on most of the intents in `test/sample-app-directory.json`, otherwise with the new logic, they match against everything


src/client/directory.ts
```diff
@@ -1,3 +1,5 @@
+import {Intent} from './internal';
```
bryangaleopenfin, line 1: This is moved out as we want to use it in `Model` and `AppDirectory`, but if we export it directly from here, we'll get name collisions from `main.ts`'s wildcard imports.

src/provider/model/AppDirectory.ts
```diff
@@ -16,6 +16,44 @@ enum StorageKeys {
 
 @injectable()
 export class AppDirectory extends AsyncInit {
+    /**
+     * Test if an app *might* support an intent - i.e., were the app running with the appropriate listener added,
+     * should we regard it as supporting this intent (and optionally context)
+     */
+    public static mightAppSupportIntent(app: Application, intentType: string, contextType?: string): boolean {
```
bryangaleopenfin, line 23: Open to better naming on `mightAppSupportIntent` and `shouldAppSupportIntent`, as current naming makes test descriptions/titles awkward.

Tempted by `hopeAppSupportsIntent` and `expectAppSupportsIntent`...

PR #160, opened 2019-10-11
[SERVICE-750] Ensure consistency between`raiseIntent`, `findIntent`, and apps shown in resolve

bryangaleopenfin:
This PR:
- Adds to AppDirectory the notion of what intents an app "might" and "should" support
- Has `Model.getApplicationsForIntent` build off of these
- Has (almost) everywhere we query intent support go through `Model.getApplicationsForIntent`
- Adds `Model.getAppIntentsByContext`, which for efficiency doesn't go though `Model.getApplicationsForIntent`, but gives results consistent with it
- Adds contexts to most of the intents in the test app directory, which is necessary to avoid a lot of false matches in our tests, and which (I think) is more representative of a real app directory
- Moves towards following Notifications conventions in unit tests
- Fixes a bug in AppWindow where the list of intents/channel context listeners would remain empty when they should be populated


src/provider/model/AppDirectory.ts
```diff
@@ -16,6 +16,44 @@ enum StorageKeys {
 
 @injectable()
 export class AppDirectory extends AsyncInit {
+    /**
+     * Test if an app *might* support an intent - i.e., were the app running with the appropriate listener added,
+     * should we regard it as supporting this intent (and optionally context)
+     */
+    public static mightAppSupportIntent(app: Application, intentType: string, contextType?: string): boolean {
```
bryangaleopenfin, line 23: Open to better naming on `mightAppSupportIntent` and `shouldAppSupportIntent`, as current naming makes test descriptions/titles awkward.

Tempted by `hopeAppSupportsIntent` and `expectAppSupportsIntent`...

src/client/directory.ts
```diff
@@ -1,3 +1,5 @@
+import {Intent} from './internal';
```
bryangaleopenfin, line 1: This is moved out as we want to use it in `Model` and `AppDirectory`, but if we export it directly from here, we'll get name collisions from `main.ts`'s wildcard imports.

src/provider/model/AppDirectory.ts
tomrobertsOF: I think as long as the meanings are well documented in both cases, there's no problem with keeping them `might` and `should`.

src/provider/model/AppDirectory.ts
```diff
@@ -16,6 +16,44 @@ enum StorageKeys {
...
+            return false;
+        } else {
+            const intent = app.intents.find(intent => intent.name === intentType);
+            return intent !== undefined && (contextType === undefined || intentSupportsContext(intent, contextType));
+        }
+    }
+
+    public static getIntentDisplayName(apps: Application[], intentType: string): string {
+        for (const app of apps) {
+            const intent = app.intents && app.intents.find(intent => intent.name === intentType);
```
tomrobertsOF, line 47: I imagine there was some discussion that I missed around how we handle the fact that there's multiple names, but is "return the first from some arbitrarily sorted list" really the right way? I can't think of any better way to be honest, but still....

src/provider/controller/IntentHandler.ts
```diff
@@ -40,74 +40,76 @@ export class IntentHandler {
...
                     `Couldn't resolve intent target '${intent.target}'. No matching app in directory or currently running.`
                 );
-            }
-
-            // Target app is in directory -> ensure that it handles intent
-            if (!(appInfo.intents || []).some(appIntent => appIntent.name === intent.type)) {
-                throw new FDC3Error(ResolveError.TargetAppDoesNotHandleIntent, `App '${intent.target}' does not handle intent '${intent.type}'`);
+            } else {
+                throw new FDC3Error(
+                    ResolveError.TargetAppDoesNotHandleIntent,
```
tomrobertsOF, line 71: Does this branch always correspond to the app not handling the intent? Are we sure there's no other case were we would fall into this section?

src/provider/model/AppWindow.ts
```diff
@@ -126,11 +126,11 @@ export abstract class AbstractAppWindow implements AppWindow {
     }
 
     public get channelContextListeners(): ReadonlyArray<ChannelId> {
-        return Object.keys(this._channelContextListeners);
+        return Array.from(this._channelContextListeners.values());
```
tomrobertsOF, line 129: What is the value of this change? Not sure I've fully understood what's happening here.

src/provider/model/Model.ts
```diff
@@ -31,6 +31,11 @@ interface ExpectedWindow {
     closed: Promise<void>;
 }
 
+interface WindowGroup {
```
tomrobertsOF: Not sure I like the naming here. There is already an established concept of a `WindowGroup` in OpenFin (not that we use it in FDC3, but still). Really this is just a single object containing an app and it's main/child windows, so couldn't we just have the windows array as a property of the application. I know that's probably a lot more work than this, but makes more sense to me in terms of the hierarchy given that an App is really just a collection of windows plus some other state.

test/provider/AppDirectory.unittest.ts
```diff
@@ -212,57 +204,205 @@ describe('When querying the Directory', () => {
...
                 }
-            ] as AppIntent[]);
+            }
         });
-    });
 
-    describe('When finding app intents by context with a context not implemented by any intent', () => {
-        it('Should return an empty array', async () => {
-            const intents = await appDirectory.getAppIntentsByContext('testContext.NonExistent');
+        it('For intents with no contexts, the app might support those intents with an arbitray context', () => {
```
ngunday: arbitray -> arbitrary

test/provider/AppDirectory.unittest.ts
```diff
@@ -212,57 +204,205 @@ describe('When querying the Directory', () => {
...
+            for (const intent of [intent3, intent4]) {
+                expect(AppDirectory.mightAppSupportIntent(app, intent.name, arbitraryContextType)).toBe(true);
+            }
         });
-    });
 
-    describe('When finding app intents by context with a context implemented by only 1 intent in 1 app', () => {
-        it('Should return 1 intent in 1 app', async () => {
-            const intents = await appDirectory.getAppIntentsByContext('testContext.Instrument');
+        it('For intents with contexts, the app will not support those intents with an arbitray context', () => {
```
ngunday: arbitray -> arbitrary

test/demo/connection.inttest.ts
```diff
@@ -210,9 +210,9 @@ async function navigateTo(target: Identity, url: string): Promise<void> {
     await delay(Duration.PAGE_NAVIGATE);
 }
 
-async function getIntentListeners(intentType: string): Promise<Application[]> {
-    return ofBrowser.executeOnWindow(SERVICE_IDENTITY, function (this: ProviderWindow, type: string): Promise<Application[]> {
```
ngunday, line 214: this change makes `Application` an unused import

src/provider/utils/async.ts
```diff
@@ -78,3 +78,23 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
     promise.catch(() => {});
     return promise;
 }
+
+export async function asyncFilter<T>(arr: T[], asyncF: (x: T) => Promise<boolean>): Promise<T[]> {
```
ngunday: think the symbol `asyncF` sounds a tiny bit ambiguous especially in a context where method name is `asyncFilter`. should this be `asyncFunc `?

src/provider/index.ts
```diff
@@ -3,9 +3,9 @@ import {inject, injectable} from 'inversify';
...
-import {AppIntent, IntentResolution, Application, Intent, Context} from '../client/main';
 import {FDC3Error, OpenError, IdentityError} from '../client/errors';
+import {RaiseIntentPayload, APIFromClientTopic, OpenPayload, FindIntentPayload, FindIntentsByContextPayload, BroadcastPayload, APIFromClient, AddIntentListenerPayload, RemoveIntentListenerPayload, GetSystemChannelsPayload, GetCurrentChannelPayload, ChannelGetMembersPayload, ChannelJoinPayload, ChannelTransport, SystemChannelTransport, GetChannelByIdPayload, ChannelBroadcastPayload, ChannelGetCurrentContextPayload, ChannelAddContextListenerPayload, ChannelRemoveContextListenerPayload, ChannelAddEventListenerPayload, ChannelRemoveEventListenerPayload, GetOrCreateAppChannelPayload, AppChannelTransport, AddContextListenerPayload, RemoveContextListenerPayload} from '../client/internal';
```
ngunday, line 7: do we have a solution for this line?

ngunday (commented):
first pass

src/provider/model/AppWindow.ts
bryangaleopenfin: `Object.keys` doesn't work as expected for `Map` (it's basically just another class, rather than anything `Object.keys` will give a meaningful result for), and was always returning an empty array

src/provider/model/AppDirectory.ts
bryangaleopenfin: There wasn't any real discussion around this, but I think this is what we were doing before somewhere, and it seemed like an improvement to just do that everywhere, rather than use the intent type for the display name.

We have https://appoji.jira.com/browse/SERVICE-751 if there's further improvements to be made

tomrobertsOF: Ok, that's fine by me.

src/provider/model/AppWindow.ts
tomrobertsOF: Ah, ok. Something leftover from when we switched from object to Map I guess.

test/provider/AppDirectory.unittest.ts
bryangaleopenfin: ✅

test/provider/AppDirectory.unittest.ts
bryangaleopenfin: ✅

test/demo/connection.inttest.ts
bryangaleopenfin: ✅

src/provider/utils/async.ts
bryangaleopenfin: Gone with `callback`, following https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

✅

src/provider/index.ts
bryangaleopenfin: I think when this has come up before, the solution has been 🤷‍♂️

src/provider/model/Model.ts
bryangaleopenfin: I don't think that would be a property of the `Application`, I'd want to keep that as a representation of the directory data. A new object to represent a running app could make sense, but that would mean big changes to `Model` and `Environment`, which feels beyond this PR. Open to push back on that though

src/provider/controller/IntentHandler.ts
bryangaleopenfin: Not that I can see. We know at this point it's either running or in the directory, so this seems like the remaining alternative. I have though changed how we test for if it's running, checking the environment, rather than for windows in the model, as that may give a misleading result if an app has windows that don't register with FDC3

src/provider/model/Model.ts
bryangaleopenfin: Renamed `WindowGroup` to `LiveApp`, which I think is an improvement, and consistent with terminology I've used elsewhere

ngunday (approved):

src/provider/controller/IntentHandler.ts
tomrobertsOF: Hmmm, ok.

src/provider/model/Model.ts
tomrobertsOF: Still think that the distinction would make sense, but happy with the name change for now.

tomrobertsOF (approved):

PR #161, opened 2019-10-16
[SERVICE-613]/[SERVICE-556] The FDC3 service ensures app will always receive "context" field when specified on the 'open' call/Lack of coherent philosophy for how long we give client to register intent/context listeners

bryangaleopenfin:
The first set of commits, approved by Ryan and Theo resolves SERVICE-613, and partially resolves SERVICE-556
- Adds `expectWindowsForApp` to `Model`, which is used when opening an app and raising intents
- Adds the concept of a `LiveApp`, used between `Environment` and `Model`, representing a running app (open to better naming on this - has to avoid conflicting with `Application` being used for static app data)
- Adds the concept of app "maturity" (i.e., the app is old enough that we can expect it to have added all its listeners), replacing separate timeouts for context listeners and intent listeners
- Changes `getApplicationsForIntent` and `getAppIntentsByContext` to take into account app maturity, and fall back to directory data for immature apps
- Expands tests for running apps to cover the "running mature" and "running immature" cases

The [second](https://github.com/HadoukenIO/fdc3-service/pull/161/files/ac9feb90e6b879df0267a8914ec248638b1a7b19..HEAD) [set](https://github.com/HadoukenIO/fdc3-service/pull/161/commits/2e5f91ba4a8c07d0871c2e51b085d5560f540280)* of commits resolves the remains of SERVICE-556, plus other changes
- Changes context handler to dispatch to windows that connect to FDC3 within our standard grace period
- Fixes an intermittent test failure caused by trying to get the info of an application before it had started
- Fixes other random test failures caused by relying on over-the-IAB API calls being "fast"
- Increases robustness of our test teardown method
- Miscellaneous tidying

*Two separate links as github is getting confused around the merge commit 🤷‍♂️


src/provider/model/Model.ts
```diff
@@ -109,42 +111,92 @@ export class Model {
...
-
-        if (matchingWindows.length === 0) {
-            const signalPromise = new Promise<AppWindow[]>(resolve => {
-                const slot = this._onWindowRegisteredInternal.add(() => {
-                    const matchingWindows = this.findWindowsByAppId(appInfo.appId);
-                    if (matchingWindows.length > 0) {
+    public async expectWindowsForApp(
+        appInfo: Application,
+        syncPredicate: (window: AppWindow) => boolean,
+        asyncPredicate: (window: AppWindow) => Promise<boolean>
```
ottah: You can use `Promise.resolve(...)` to wrap sync values/functions and then just use the `asyncPredicate` without having to have a duplicate predicate.

src/provider/index.ts
```diff
@@ -103,32 +108,52 @@ export class Main {
...
         if (!appInfo) {
             throw new FDC3Error(OpenError.AppNotFound, `No app in directory with name: ${payload.name}`);
         }
 
-        // This can throw FDC3Errors if app fails to open or times out
-        await this._model.ensureRunning(appInfo);
+        const promises: Promise<void>[] = [];
+
+        // Start the application if not already running
+        const startedPromise = (await this._model.getOrCreateLiveApp(appInfo)).startedPromise;
```
ottah: Just destructure this? `const {startedPromise}` or just directly push it into `promises`.

src/provider/index.ts
```diff
@@ -103,32 +108,52 @@ export class Main {
...
+        promises.push(startedPromise);
+
+        // If the app has open windows, bring all to front in creation order
+        const windows = this._model.findWindowsByAppName(appInfo.name);
+        if (windows.length > 0) {
+            const windowsToFocus = windows.sort((a, b) => a.appWindowNumber - b.appWindowNumber);
+
+            const bringToFrontPromise = Promise.all(windowsToFocus.map(window => window.bringToFront()));
+            const focusPromise = bringToFrontPromise.then(async () => {
+                if (windowsToFocus.length > 0) {
```
ottah: When would this never be true?

src/provider/model/Model.ts
ottah: After looking at how this function is called, these 2 predicates are 2 different tests might make sense to rename them so they make more sense. `readyNow` & `waitFor`?

src/provider/model/AppWindow.ts
```diff
@@ -86,7 +84,7 @@ export abstract class AbstractAppWindow implements AppWindow {
     private readonly _appInfo: Application;
     private readonly _appWindowNumber: number;
 
-    private readonly _creationTime: number | undefined;
+    private readonly _maturePromise: Promise<void>;
```
ottah, line 85: Is `ready` or `initialized` not a better word?

src/provider/model/AppWindow.ts
```diff
@@ -225,25 +216,18 @@ export abstract class AbstractAppWindow implements AppWindow {
...
-
-            return !didTimeout;
-        }
+            deferredPromise.resolve();
+        });
+
+        // App may be starting - give until app maturity to register a listener
+        return Promise.race([
+            this._maturePromise.catch(() => {}).then(() => false),
+            deferredPromise.promise.then(() => true)
```
ottah: Use `deferredPromise.resolve(true)`?

src/provider/model/LiveApp.ts
```diff
@@ -0,0 +1,84 @@
+import {Application} from '../../client/directory';
+import {DeferredPromise} from '../common/DeferredPromise';
+import {Timeouts} from '../constants';
+
+import {AppWindow} from './AppWindow';
+
+/**
+ * Represents a running appliction
```
ottah: appliction -> application

ottah (commented):
First pass.

src/provider/index.ts
bryangaleopenfin: ✅

src/provider/index.ts
bryangaleopenfin: Err...

✅

src/provider/model/AppWindow.ts
bryangaleopenfin: I've done a biggish re-work of this

src/provider/model/LiveApp.ts
bryangaleopenfin: ✅

src/provider/model/AppWindow.ts
bryangaleopenfin: Gonna say no. Think "mature" better communicates that this is something that occurs after some non-trivial period of time

src/provider/model/Model.ts
bryangaleopenfin: This prompted another reworking of `AppWindow`...

✅

src/provider/controller/ContextHandler.ts
```diff
@@ -31,12 +31,10 @@ export class ContextHandler {
      */
     public async send(window: AppWindow, context: Context): Promise<void> {
```
rypete: `async` is no longer needed here (hooray pre linting :))

src/provider/controller/IntentHandler.ts
```diff
@@ -117,31 +103,17 @@ export class IntentHandler {
...
-            }));
-        })());
-
-        dispatchingCompleted = true;
-
-        if (dispatchResults[0] || !dispatchResults[1]!.includes(true)) {
+        const listeningWindows = await this._model.expectWindowsForApp(
+            appInfo,
+            (window) => window.hasIntentListener(intent.type),
+            async (window) => window.waitForReadyToReceiveIntent(intent.type)
```
rypete: Does this need to be `async`?

src/provider/controller/IntentHandler.ts
```diff
@@ -117,31 +103,17 @@ export class IntentHandler {
...
+        );
+
+        if (listeningWindows.length > 0) {
```
rypete, line 131: Petty, but could just be `if (listeningWindows.length) {`

src/provider/index.ts
```diff
@@ -103,32 +108,50 @@ export class Main {
...
+        const promises: Promise<void>[] = [];
+
+        // Start the application if not already running
+        const {startedPromise} = await this._model.getOrCreateLiveApp(appInfo);
+
+        promises.push(startedPromise);
+
+        // If the app has open windows, bring all to front in creation order
+        const windows = this._model.findWindowsByAppName(appInfo.name);
+        if (windows.length > 0) {
```
rypete, line 128: Same with above

src/provider/index.ts
```diff
@@ -103,32 +108,50 @@ export class Main {
...
+            windows.sort((a, b) => a.appWindowNumber - b.appWindowNumber);
+
+            const bringToFrontPromise = Promise.all(windows.map(window => window.bringToFront()));
```
rypete: Keeping consistency with map elsewhere in your PR - can you put `windows` arg in parens?

src/provider/index.ts
```diff
@@ -103,32 +108,50 @@ export class Main {
...
-        if (payload.context) {
-            // TODO: Revisit timeout logic [SERVICE-556]
-            await withTimeout(Timeouts.ADD_CONTEXT_LISTENER, (async () => {
-                const appWindows = await this._model.expectWindowsForApp(appInfo);
+        // If a context has been provided, send to listening windows
+        if (context) {
+            const windowsPromise = this._model.expectWindowsForApp(
+                appInfo,
+                (window) => window.hasContextListener(),
+                async (window) => window.waitForReadyToReceiveContext()
```
rypete: Does this need to be `async`?

src/provider/index.ts
```diff
@@ -103,32 +108,50 @@ export class Main {
...
+            );
 
-                await Promise.all(appWindows.map(window => {
-                    return this._contextHandler.send(window, parseContext(payload.context!));
-                }));
-            })());
+            const sendContextPromise = windowsPromise.then(async (windows) => {
+                await Promise.all(windows.map(window => this._contextHandler.send(window, context)));
```
rypete: Same with above. Put `windows` arg in parens

src/provider/index.ts
bryangaleopenfin: Actually, changed `startedPromise` to be `waitForAppStarted()`, so destructuring not an option, and would prefer to keep this seperate from the push statement

src/provider/index.ts
```diff
@@ -103,32 +108,50 @@ export class Main {
...
+        promises.push(startedPromise);
+
+        // If the app has open windows, bring all to front in creation order
+        const windows = this._model.findWindowsByAppName(appInfo.name);
+        if (windows.length > 0) {
+            windows.sort((a, b) => a.appWindowNumber - b.appWindowNumber);
+
+            const bringToFrontPromise = Promise.all(windows.map(window => window.bringToFront()));
+            const focusPromise = bringToFrontPromise.then(async () => {
+                await windows[windows.length - 1].focus();
```
rypete: I think you can remove the `async` and `await` here and instead only return `windows[windows.length - 1].focus()`. The promise.all at the end should then await the focus resolve instead of the resolve of the focus resolve. This same logic can be applied to line 148 as well

src/provider/model/AppWindow.ts
```diff
@@ -190,60 +193,16 @@ export abstract class AbstractAppWindow implements AppWindow {
...
-                }
-            });
-
-            const [didTimeout] = await withTimeout(Timeouts.ADD_INTENT_LISTENER - age, deferredPromise.promise);
-
-            slot.remove();
-
-            return !didTimeout;
-        }
+    public async waitForReadyToReceiveIntent(intent: IntentType): Promise<void> {
```
rypete: No async

src/provider/model/AppWindow.ts
```diff
@@ -190,60 +193,16 @@ export abstract class AbstractAppWindow implements AppWindow {
...
-            const deferredPromise = new DeferredPromise();
-
-            const slot = this._onContextListenerAdded.add(() => {
-                deferredPromise.resolve();
-            });
-
-            const [didTimeout] = await withTimeout(Timeouts.ADD_CONTEXT_LISTENER - age, deferredPromise.promise);
-
-            slot.remove();
+    public async waitForReadyToReceiveContext(): Promise<void> {
```
rypete: No async

src/provider/model/AppWindow.ts
```diff
@@ -190,60 +193,16 @@ export abstract class AbstractAppWindow implements AppWindow {
...
+        return this.waitForListener(this._onContextListenerAdded, () => this.hasContextListener());
+    }
 
-            return !didTimeout;
-        }
+    public async waitForReadyToReceiveContextOnChannel(channel: ContextChannel): Promise<void> {
```
rypete: No async

src/provider/model/LiveApp.ts
```diff
@@ -0,0 +1,84 @@
...
+    private readonly _maturePromise: Promise<void>;
+
+    private _appInfo: Application | undefined = undefined;
+    private _mature: boolean = false;
+
+    /**
+     * Constructs a new LiveApp
+     *
+     * @param startedPromise A promise that resolves once the app fully started, or undefined if the app started "long ago" (e.g., was
+     * running before the servive started)
```
rypete: `servive` > `service`

src/provider/model/LiveApp.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    public get startedPromise(): Promise<void> {
+        return this._startedPromise;
+    }
+
+    public get maturePromise(): Promise<void> {
+        return this._maturePromise;
+    }
+
+    public async getAppInfo(): Promise<Application> {
```
rypete: No `async`

test/demo/open.inttest.ts
```diff
@@ -84,21 +81,98 @@ describe('Opening applications with the FDC3 client', () => {
...
+                    const childWindow = await fdc3Remote.createFinWindow(testAppInDirectory1, {url: testAppUrl, name: 'child-window'});
+                    const listener = await fdc3Remote.addContextListener(childWindow);
+
+                    await openPromise;
+
+                    // Check that the app received the context passed in open and nothing else
+                    await expect(listener).toHaveReceivedContexts([validContext]);
+                });
+
+                test('When the app adds listeners on mulitple windows, the app opens and the first windows\'s context listener is \
```
rypete: `window's?` Also how come punctuation here but no where else

test/provider/AppWindow.unittest.ts
```diff
@@ -89,47 +91,47 @@ describe('When querying if a window is ready to receive contexts', () => {
...
-        testAppWindow = new TestAppWindow(fakeIdentity, fakeAppInfo, mockChannel, Date.now(), 0);
+        testAppWindow = new TestAppWindow(fakeIdentity, fakeAppInfo, createMaturityPromise(), mockChannel, 0);
     });
 
     afterEach(() => {
         unmockTime();
     });
 
-    test('A window with a context listener already registered returns true immediately', async () => {
+    test('A window with a context listener already registered resovles immediately', async () => {
```
rypete: `resolves`

rypete (commented):
Logic looks solid overall. Some async items and spelling errors to fix up

test/demo/open.inttest.ts
ottah: `mulitple` as well

ottah (approved):

src/provider/controller/ContextHandler.ts
bryangaleopenfin: ✅

src/provider/controller/IntentHandler.ts
bryangaleopenfin: ✅

src/provider/index.ts
bryangaleopenfin: ✅

src/provider/model/AppWindow.ts
bryangaleopenfin: ✅

src/provider/model/AppWindow.ts
bryangaleopenfin: ✅

src/provider/model/AppWindow.ts
bryangaleopenfin: ✅

src/provider/model/LiveApp.ts
bryangaleopenfin: ✅

test/provider/AppWindow.unittest.ts
bryangaleopenfin: ✅

test/demo/open.inttest.ts
bryangaleopenfin: Spelling corrected. Not sure what punctuation you mean though

src/provider/model/LiveApp.ts
bryangaleopenfin: ✅

src/provider/controller/IntentHandler.ts
bryangaleopenfin: I could but I don't want to!

We're using `if (whatever.length > 0)` everywhere else, and I prefer this

src/provider/index.ts
bryangaleopenfin: Ditto

src/provider/index.ts
bryangaleopenfin: ✅

src/provider/index.ts
bryangaleopenfin: ✅

src/provider/index.ts
bryangaleopenfin: ✅

test/demo/open.inttest.ts
bryangaleopenfin: ✅

bryangaleopenfin:
Feedback addressed, and backported some fixes from a subsequent PR - https://github.com/HadoukenIO/fdc3-service/pull/161/commits/1d4870f7fdbe7bbedc511c86e4958e1bef0ce9eb - that are worth taking a look at specifically

rypete (approved):

ottah:
Ah in that case ignore that comment. I wish there was a different method name for `then(success, reject)` so it is easier to spit. Something like `Either()`.

src/provider/model/Model.ts
```diff
@@ -308,6 +297,8 @@ export class Model {
         const {uuid} = identity;
         this._liveAppsByUuid[uuid] = liveApp;
 
+        await liveApp.waitForAppStarted();
```
bryangaleopenfin, line 300: This missing line was the cause of all my problems

src/provider/controller/ContextHandler.ts
```diff
@@ -63,19 +75,46 @@ export class ContextHandler {
...
+                this._model.expectWindowsForApp(
+                    appInfo,
+                    (window: AppWindow) => window.hasContextListener(),
+                    async (window: AppWindow) => window.waitForReadyToReceiveContext()
+                ).then((windows) => {
+                    windows
+                        .filter(notSender)
+                        .filter(window => !memberWindows.includes(window))
+                        .filter(window => window.channel.id === channel.id)
+                        .forEach(window => this.send(window, context));
```
ottah: Can these filters be merged into 1? Or better yet use an if statement inside the forEach that filters. This is worse case looping 4 times. Same for the filter below.

test/provider/ContextHandler.unittest.ts
```diff
@@ -194,3 +185,24 @@ describe('When broadcasting a Context using ContextHandler', () => {
...
     });
 });
+
+function createCustomMockAppWindow(name: string, listensForContext: boolean): jest.Mocked<AppWindow> {
+    const mockWindow = createMockAppWindow({
+        identity: {uuid: 'test', name},
+        waitForReadyToReceiveContext: listensForContext ? jest.fn().mockResolvedValue(undefined) : jest.fn().mockRejectedValue(undefined)
+    });
+
+    mockWindows.push(mockWindow);
```
ottah, line 193: This seems like a weird side effect that you would miss when looking at how `mockWindows` is populated.

src/provider/index.ts
```diff
@@ -141,12 +139,12 @@ export class Main {
...
-                async (window) => window.waitForReadyToReceiveContext()
+                (window) => window.waitForReadyToReceiveContext()
             );
 
-            const sendContextPromise = windowsPromise.then(async (windows) => {
-                await Promise.all(windows.map(window => this._contextHandler.send(window, context)));
-            });
+            const sendContextPromise = windowsPromise.then((windows) => {
+                return Promise.all(windows.map((window) => this._contextHandler.send(window, context)));
+            }).then(() => {});
```
ottah: What does `.then(() => {})` add?

rypete (approved):
I am happy with these changes

src/provider/model/FinEnvironment.ts
```diff
@@ -114,48 +114,91 @@ export class FinEnvironment extends AsyncInit implements Environment {
...
+    }
+
+    private deregisterWindow(identity: Identity): void {
+        if (identity.uuid !== SERVICE_IDENTITY.uuid) {
+            this._windows.delete(getId(identity));
+            this.windowClosed.emit(identity);
+        }
+    }
+
+    private registerApplication(identity: Identity, startedPromise: Promise<void> | undefined): void {
```
ottah, line 185: I feel like using **?** is easier to read/notice 
`startedPromise?: Promise<void>`

src/provider/model/LiveApp.ts
```diff
@@ -0,0 +1,107 @@
...
+    private _started: boolean = false;
+    private _mature: boolean = false;
+
+    /**
+     * Constructs a new LiveApp
+     *
+     * @param startedPromise A promise that resolves once the app fully started, or undefined if the app started "long ago" (e.g., was
+     * running before the service started)
+     */
+    public constructor(startedPromise: Promise<void> | undefined) {
```
ottah, line 30: Same as above.

src/provider/index.ts
bryangaleopenfin: To make the promise type `Promise<void>` - have made a slight change to how I'm doing this

src/provider/model/FinEnvironment.ts
bryangaleopenfin: I'm against doing this, as I want the caller to *know* they're passing in undefined here - in this context, undefined should be a deliberate choice, rather than a fallback

src/provider/model/LiveApp.ts
bryangaleopenfin: Same response as above

test/provider/ContextHandler.unittest.ts
bryangaleopenfin: Could have the caller of `createCustomMockAppWindow` add the returned value to this array, but would rather this stay as a background detail that the called doesn't have to worry about.

Would changing name from `createCustomMockAppWindow` to `setupCustomMockAppWindow` help?

src/provider/index.ts
ottah: Yeah an await makes more sense 👌

test/provider/ContextHandler.unittest.ts
ottah: Yeah would prefer setupCustomMockAppWindow.

src/provider/controller/ContextHandler.ts
bryangaleopenfin: Have merged filters, but would rather keep distinct from the forEach for clarity

test/provider/ContextHandler.unittest.ts
bryangaleopenfin: ✅

PR #162, opened 2019-10-18
[N/A] Remove prefixes from intent names

pjbroadbent:
FDC3 specs only mention prefixes on content names, all intent types used in the spec docs lack prefixes - updating demo app and tests to match.


bryangaleopenfin:
Not sure I agree with removing the "test." prefixes, as these make it obvious what's there for a specific test, and there's no FDC3 defined intents to match in those cases

rypete (approved):
Nobody could remove those prefixes better than you!

bryangaleopenfin (approved):

PR #165, opened 2019-10-22
[SERVICE-558] Intent resolution data

ottah:
This PR populates the `IntentResolution.data` field with a value returned by a intent listener. The rules for deciding the return value is as follows: 
- When a listener return value is void, null is used.
- When listener returns a value, the value is used.
- When a listener throws an error, an FDC3 IntentHandlerException error thrown.
- When a listener is added, but takes longer than 5seconds to complete it will throw an error.
- When there is multiple listeners across multiple windows, the first **NON** error/null is used. The intent resolution will continue until a value is found, if only an error is found it will throw an error, if only null/void it will return null.
The tests in `intentResolution.inttest.ts` describe these cases.

Other changes:
- Bring over test framework changes from notifications.
- Enable mounting custom function intent listeners.


src/client/main.ts
```diff
@@ -298,14 +300,20 @@ function deserializeChannelChangedEvent(eventTransport: Transport<ChannelChanged
     return {type, identity, channel, previousChannel};
 }
 
+function returnValuePredicate(value?: any): boolean {
+    return (value === false || value === '' || !!value);
```
bryangaleopenfin: Are `0`, `null`, and `NaN` not valid return values? And can this be better named? `isValidIntentResult`?

src/provider/utils/async.ts
```diff
@@ -79,6 +79,32 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
     return promise;
 }
 
+export async function raceTilPredicate(items: any[], predicate: (v: any) => boolean): Promise<any | null> {
```
bryangaleopenfin: Suggest bring more in line with similar functions in this file
- Comment
- Think  `raceUntilTrue` would be a more consistent name
- Use generics rather than any

src/provider/utils/async.ts
```diff
@@ -79,6 +79,32 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
...
+            }
+            valueFound.resolve(null);
+        }
+        await valueFound.promise;
+    };
+    // Wrap `item` incase it is not a Promise
+    const racePromises = items.map(item => Promise.resolve(item).then(resolveCheck, resolveCheck));
+    racePromises.push(valueFound.promise);
+
+    return Promise.race(racePromises);
```
bryangaleopenfin: VSC seems to think the return type here will be `Promise<void>` (mouse over `racePromises`). If that's not the case, I think it's worth adding some explicit types

src/provider/utils/async.ts
```diff
@@ -79,6 +79,32 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
...
 }
 
+export async function raceTilPredicate(items: any[], predicate: (v: any) => boolean): Promise<any | null> {
+    let promisesCompleted = 0;
+    const errors: Error[] = [];
+    const valueFound = new DeferredPromise<any | null>();
+
+    const resolveCheck = async (value?: any) => {
+        promisesCompleted++;
+        if (value instanceof Error) {
```
bryangaleopenfin: Would rather have separate resolve and reject handlers, than merge the two results then try to separate them out again. If a promise rejects, even if it doesn't reject with an `Error`object, I think we want to treat it as an error. And the inverse for resolve - if it resolves with an `Error` object, that's still valid

test/demo/raiseIntent/withTarget.inttest.ts
```diff
@@ -283,8 +283,8 @@ async function raiseDelayedIntentWithTarget(intent: Intent, targetApp: Directory
     return raiseIntentPromise;
 }
 
-function raiseIntent(intent: Intent, target?: TestAppData): Promise<void> {
-    return fdc3Remote.raiseIntent(
+async function raiseIntent(intent: Intent, target?: TestAppData): Promise<void> {
+    await fdc3Remote.raiseIntent(
```
bryangaleopenfin, line 396: What difference does this change make?

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+    return async () => {
+        await delay(time);
+        return fn();
+    };
+}
+
+async function createChildWindows(target: Identity, number: number): Promise<Identity[]> {
+    const ids = new Array(number)
+        .fill(null)
+        .map((_, i) => fdc3Remote.createFinWindow(target, {name: `child-window-${i}`, url: 'http://localhost:3923/test/test-app.html'}));
```
bryangaleopenfin: We have a `testAppUrl` constant

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+function raiseIntent(intent: Intent, target: TestAppData): Promise<IntentResolution> {
+    return fdc3Remote.raiseIntent(
+        testManagerIdentity,
+        intent.type,
+        intent.context,
+        target.name
+    );
+}
+
+function createDelayedFunction(fn: (...args: any[]) => any, time: number): () => Promise<any> {
```
bryangaleopenfin: Can this use generics rather than `any`? We know the last two any's should be the same type

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+async function createChildWindows(target: Identity, number: number): Promise<Identity[]> {
+    const ids = new Array(number)
+        .fill(null)
+        .map((_, i) => fdc3Remote.createFinWindow(target, {name: `child-window-${i}`, url: 'http://localhost:3923/test/test-app.html'}));
+    return Promise.all(ids);
+}
+
+async function openDirectoryApp(app: TestAppData) {
```
bryangaleopenfin: I don't understand the return value here, and I think it's better to use `setupOpenDirectoryAppBookends` anyway

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
+import 'jest';
```
bryangaleopenfin: Unused imports

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+        });
+
+        describe('When there is a single window', () => {
+            test('And there is 1 intent handler that throws an error, an error is thrown', async () => {
+                await fdc3Remote.addIntentListener(intentHandlingApp, preregisteredIntent.type, errorFn);
+                const resolution = raiseIntent(preregisteredIntent, intentHandlingApp);
+                await expect(resolution).rejects.toThrowError();
+            });
+
+            test('And there is multiple intent handlers that do not throw an error, a value is returned', async () => {
```
bryangaleopenfin: "And there are" here and a few other places below

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+        .map((_, i) => fdc3Remote.createFinWindow(target, {name: `child-window-${i}`, url: 'http://localhost:3923/test/test-app.html'}));
+    return Promise.all(ids);
+}
+
+async function openDirectoryApp(app: TestAppData) {
+    await fdc3Remote.open(testManagerIdentity, app.name);
+    return testManagerIdentity;
+}
+
+async function openNonDirectoryApp(app: NonDirectoryTestAppData): Promise<Identity> {
```
bryangaleopenfin: This is unused, and think it's better to use `setupStartNonDirectoryAppBookends`

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+        });
+    }), 1],
+    ['a delayed async value', appHandlingIntent, jest.fn(async () => {
+        await delay(1000);
+        return 1;
+    }), 1],
+    ['a literal', appHandlingIntent, jest.fn(() => 1), 1]
+];
+
+// E.g. When an intent handler returns an error
```
bryangaleopenfin: Not sure what this comment is attached to

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+ */
+const preregisteredIntent: Intent = {
+    type: 'test.IntentNamePreregistered',
+    context: {type: 'preregistered.IntentNamePreregisteredContext'}
+};
+const appHandlingIntent: TestAppData = testAppInDirectory2;
+
+setupTeardown();
+
+type ContextListener = jest.Mock | (() => any);
```
bryangaleopenfin: I'm not sure `jest.Mock` adds anything here, or at least it seems to still compile without it. Or can we specify some generic params for on `jest.Mock`?

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+
+type TestParam = [
+    string,
+    TestAppData,
+    ContextListener | undefined,
+    any // The value IntentResolution.data should resolve to
+];
+
+const resolutionValueTests: TestParam[] = [
```
bryangaleopenfin: Convention elsewhere is to name this `somethingSomethingTestParams`

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+    ['a promise', appHandlingIntent, jest.fn(() => {
```
bryangaleopenfin: If `appHandlingIntent` doesn't vary, would leave it out of the test params

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+                });
+
+                test('And a child window has the only intent listener, the expected value gets resolved', async () => {
+                    await fdc3Remote.addIntentListener(children[1], preregisteredIntent.type, listener);
+                    const resolution = raiseIntent(preregisteredIntent, intentHandlingApp);
+
+                    await testExpectedResolution(resolution, expectedValue);
+                });
+
+                test('And all the windows return a value, the first to return is the value resolved', async () => {
```
bryangaleopenfin: This test doesn't seem like it relates to the test params, and would make more sense outside the `describe.each` block. Like, if `expectedValue` were anything other than `1`, there's no value you could give to the other test params to make this work, right?

test/demo/utils/jestExtensions.ts
```diff
@@ -115,8 +115,18 @@ expect.extend({
         };
     },
 
-    async toHaveReceivedContexts(listener: RemoteContextListener, expectedContexts: FDC3Context[]): Promise<CustomMatcherResult> {
-        const receivedContexts = await listener.getReceivedContexts();
+    async toHaveReceivedContexts(
```
bryangaleopenfin: Does this change get used at all?

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,215 @@
...
+                    await Promise.all(childListeners);
+                    const resolution = raiseIntent(preregisteredIntent, intentHandlingApp);
+                    await testExpectedResolution(resolution, expectedValue);
+                });
+            });
+        }
+    );
+
+    describe('Error handling', () => {
+        const intentHandlingApp: TestAppData = appHandlingIntent;
```
bryangaleopenfin: Could this just use `appHandlingIntent` directly?

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -23,26 +23,23 @@ export const ofBrowser = new OFPuppeteerBrowser();
 
 const remoteChannels: {[id: string]: RemoteChannel} = {};
 
-export interface RemoteContextListener {
+export interface RemoteListener {
```
bryangaleopenfin: Would leave unexported if we're not using elsewhere

test/demo/raiseIntent/intentResolution.inttest.ts
bryangaleopenfin: What do we gain from these being jest mocks, rather than just normal arrow functions?

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -327,9 +334,14 @@ function createRemoteIntentListener(executionTarget: Identity, id: number, inten
             }, intent, id);
         },
         getReceivedContexts: async (): Promise<Context[]> => {
-            return ofBrowser.executeOnWindow(executionTarget, function(this: TestWindowContext, intent: IntentType, id: number): Context[] {
-                return this.receivedIntents.filter(entry => entry.listenerID === id && entry.intent === intent).map(entry => entry.context);
-            }, intent, id);
+            if (!listener || !('mock' in listener)) {
```
bryangaleopenfin: Not properly got my head round this, but feels un-intuitive to a user of this that we'd treat mocks as a special case

src/provider/controller/IntentHandler.ts
```diff
@@ -166,3 +168,8 @@ interface IntentWithTarget extends Intent {
 function hasTarget(intent: Intent): intent is IntentWithTarget {
     return !!intent.target;
 }
+
+// Check if the intent handler return value is valid defined.
+function returnValuePredicate(value?: any) {
```
bryangaleopenfin: Same comment as for this function in client

src/provider/controller/IntentHandler.ts
```diff
@@ -118,42 +118,44 @@ export class IntentHandler {
...
+        let data: unknown;
+        try {
+            // Use the first handler return data that matches the predicate
+            const [didTimeout, result] = await withTimeout(Timeouts.INTENT_RESOLUTION, raceTilPredicate(promises, returnValuePredicate));
+            if (didTimeout) {
+                throw new Error('Race timed out');
+            }
+            data = result;
+        } catch (error) {
+            if (/Exceptions in all/.test(error.message)) {
```
bryangaleopenfin: Is there an alternative to matching against error messages here? Feels fragile

src/client/main.ts
bryangaleopenfin: (Feel fairly strongly that `0` should be a valid result, and once that's allowed, I think it's easiest to communicate to users as "any non-void/non-undefined" result)

bryangaleopenfin:
Looks like Jenkins failure is legit?

src/client/main.ts
ottah: Oh yeah forgot about 0. Can't wait for TS3.7 to add `??`.

src/provider/utils/async.ts
ottah: Good catch. Instead of returning it awaits the race then returns the `valueFound.promise`. Its the same behaviour as before but fixes the intellisense.

src/provider/controller/IntentHandler.ts
ottah: Reworked the flow of this.

test/demo/raiseIntent/intentResolution.inttest.ts
ottah: I was missing the intent listener on the `intentHandlingApp`. Fixed now.

src/provider/utils/async.ts
ottah: Reworked this whole function.

test/demo/raiseIntent/intentResolution.inttest.ts
ottah: Removed the args

test/demo/raiseIntent/withTarget.inttest.ts
bryangaleopenfin: Ah, n/m, I see!

test/demo/raiseIntent/intentResolution.inttest.ts
bryangaleopenfin: "And there are an error on the main window"

src/provider/utils/async.ts
```diff
@@ -79,6 +79,48 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
...
+        if (predicate(value)) {
+            valueFound.resolve(value);
+        }
+    };
+
+    const handleReject = (): void => {
+        errors++;
+    };
+
+    const completeCheck = async (value: T | void): Promise<T | null> => {
```
bryangaleopenfin: Can this just return void, given the result of racePromises isn't used?

src/provider/utils/async.ts
```diff
@@ -79,6 +79,48 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
...
+        if (value) {
+            return value;
+        }
+        if (items.length === promisesCompleted) {
+            if (items.length === errors) {
+                throw new Error('All promises rejected');
+            }
+            valueFound.resolve(null);
+        }
+        await valueFound.promise;
```
bryangaleopenfin: Is this needed?

test/demo/raiseIntent/intentResolution.inttest.ts
bryangaleopenfin: Ah, I see. The test would still pass if the third listener to return is the value we use 😛.

Suggest change to something like `createDelayedFunction(() => i, 2000 * (i + 1)));` to `createDelayedFunction(() => {not: expectedValue}, 2000 * (i + 1)));`

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,189 @@
...
+    ContextListener | undefined,
+    any // The value IntentResolution.data should resolve to
+];
+
+const resolutionTestParams: TestParam[] = [
+    ['a delayed value', async () => {
+        await delay(1000);
+        return 1;
+    }, 1],
+    ['a literal', () => 1, 1]
```
bryangaleopenfin: Could these `1`s be something like `"intent-result"`?

src/client/main.ts
bryangaleopenfin: Still want to push on the naming of this, and still think this should be more permissive about what gets accepted as a return value. I think anything other than `undefined` is an attempt to return something, and whatever the rule, it has to be communicated to developers, and "anything other than undefined" is most straightforward

src/provider/utils/async.ts
bryangaleopenfin: Would still suggest commenting similar to other functions, think behaviour is non-obvious enough to be worth it

src/client/main.ts
```diff
@@ -69,7 +71,7 @@ export interface ContextListener {
 
 export interface IntentListener {
     intent: string;
-    handler: (context: Context) => void;
+    handler: (context: Context) => any;
```
tim-dinsdale: Is this any or unknown? I think it's unknown!

src/provider/controller/IntentHandler.ts
tim-dinsdale: If that comment is 'please share duplicated code' I agree!

src/provider/utils/async.ts
```diff
@@ -79,6 +79,48 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
     return promise;
 }
 
+export async function raceUntilTrue<T>(items: T[], predicate: (v?: T) => boolean): Promise<T | null> {
```
tim-dinsdale: Can we have a function comment please, given this will end up a shared function?

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -99,19 +96,25 @@ export async function addContextListener(executionTarget: Identity): Promise<Rem
...
         if (this.intentListeners[intentRemote] === undefined) {
             this.intentListeners[intentRemote] = [];
         }
         const listenerID = this.intentListeners[intentRemote].length;
-        this.intentListeners[intentRemote][listenerID] = this.fdc3.addIntentListener(intentRemote, (context) => {
+        this.intentListeners[intentRemote][listenerID] = this.fdc3.addIntentListener(intentRemote, async (context) => {
             this.receivedIntents.push({listenerID, intent: intentRemote, context});
+            // Catch and throw so it rejects correctly client side
+            const result = listenerRemote ? await listenerRemote(context).catch(() => {
+                throw new Error();
```
tim-dinsdale: Should this guy populate the error with what comes from the catch?

test/demo/utils/uuidv4.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Generates a random uuid based on math.random and the system time
+ */
+export function uuidv4() {
+    return (Math.random().toString(36).substring(2) + Date.now().toString(36));
+}
```
tim-dinsdale: I get nervous when I see something claim it's uuidv4 and it doesn't use proper uuid code. Tell me in a convincing way that this is OK, or consider having a dev dependency on uuid?

src/client/main.ts
ottah: Changed to unknown

src/provider/utils/async.ts
ottah: Added documentation

test/demo/utils/fdc3RemoteExecution.ts
ottah: The caught error is thrown now.

test/demo/utils/uuidv4.ts
ottah: Used a different uuidv4 function.

test/demo/raiseIntent/intentResolution.inttest.ts
ottah: Changed to intent-result

src/provider/utils/async.ts
ottah: It needs to be awaited otherwise it will return and end the race.

src/provider/utils/async.ts
ottah: Returns `Promise<void>`

src/provider/controller/IntentHandler.ts
ottah: Changed the rules to be any defined value. Removed this predicate function.

src/provider/utils/async.ts
ottah: Added documentation.

src/client/main.ts
ottah: Removed the predicate function and now just do `value !== undefined`.

test/demo/raiseIntent/intentResolution.inttest.ts
ottah: Now its just a string so it can't pass. I do think its important to have a unique value that can be used to see what one is returning first.

bryangaleopenfin:
My robustness-against-client-errors PR is in. Sorry! Will resolve what I can then spend some time thinking about how best to reconcile

src/provider/controller/IntentHandler.ts
bryangaleopenfin: @tim-dinsdale Gonna assume you're fine for me to resolve this!

tim-dinsdale (approved):

src/provider/utils/async.ts
bryangaleopenfin: I wonder if this function should `Promise.all` rather than `Promise.race` then, then I think you *could* get rid of this line. And I thiiiink you could get rid of `racePromises.push(valueFound.promise);`.

But hold off changing until I've looked into reconciling this and develop!

bryangaleopenfin:
Places this is going to conflict:

- `ofPuppeteer` - we've both made the same change - copying code over from `notifications`, so should be easy enough to bring in one or the other

- `IntentHandler` - on develop this uses `collateClientResults` (which is the same method used wherever we do a service-to-client call), on this branch, it uses `raceUntilTrue`. Suggest keep using `collateClientResults`, which shouldn't need much modification to return the first non-undefined result. `collateClientResults` also differs from `raceUntilTrue` in that it'll wait for *all* promises to resolve or timeout, but I think for service-to-client calls that's the right approach - this guarantees we will have hit all target windows

- `main` - on develop this uses `invokeListeners` (which is the same method used by the two places we call context listeners) and on this branch it uses `raceUntilTrue`. I think keep using `invokeListeners`, since it provides some higher-level common logic, but bring in the logic from `raceUntilTrue`, so that it returns as soon as we have a single result, rather than waiting until all promises have resolved

- `constants` - on develop we now have a general `SERVICE_TO_CLIENT_API_CALL` constant. Preference would be to just use this, rather than also have a timeout specific for intent resolution, but could see an argument for keeping them both, and allowing `collateClientResults` to optionally take a timeout to override the `SERVICE_TO_CLIENT_API_CALL` default

- `errors` - have been renamed on develop to match actual calls, think should now have everything you need on develop

test/demo/raiseIntent/intentResolution.inttest.ts
```diff
@@ -0,0 +1,192 @@
+import 'jest';
```
bryangaleopenfin: Does this need to be its own file? What's the logic for what goes in here, and what goes in `withTarget` or `withoutTarget`?

test/demo/raiseIntent/intentResolution.inttest.ts
bryangaleopenfin: Cool. Think it would be clearer as `createDelayedFunction(() => 'delayed-value-${i}', 2000 * (i + 1)))`, and then it becomes obvious what `expect = 'delayed-value-0'` is about.

Would also change "the first to return" to "the first to return a non-undefined value" in the test title

test/demo/utils/jestExtensions.ts
ottah: Reset this to how it was before. I had planned to fully remove the remote stuff and just use mocks, but then realised that we have intents preregistered apps so its pointless not just using the remote get received context api.

test/demo/utils/fdc3RemoteExecution.ts
ottah: Removed

src/client/main.ts
```diff
@@ -5,6 +5,8 @@
 /**
  * Main entry point
  */
+// import {raceUntilTrue} from '../provider/utils/async';
```
bryangaleopenfin: Leftover comment

src/client/internal.ts
```diff
@@ -246,35 +246,43 @@ export interface ChannelReceiveContextPayload {
 /**
  * Invokes an array of listeners with a given context, allowing us to apply consistent error handling. Will throw an error if > 0 listeners
  * are given, and all fail
+ * The first only *defined* return value is returned
```
bryangaleopenfin: Can this fit better with the existing comment?

    are given, and all fail. Otherwise, the first *defined* value returned by a listener is returned, or undefined if no defined values are returned.

src/provider/utils/helpers.ts
```diff
@@ -37,6 +37,9 @@ export async function collateClientCalls<T = void>(promises: Promise<T>[]): Prom
         if (successes === 0) {
             result = promiseResult;
         }
+        if (result === undefined && promiseResult !== undefined) {
```
bryangaleopenfin, line 36: Suggest explicitly initialize `result` to `undefined`, and get rid of the `if (successes === 0)` block

test/demo/utils/fdc3RemoteExecution.ts
```diff
@@ -103,17 +100,25 @@ export async function addContextListener(executionTarget: Identity): Promise<Rem
     return createRemoteContextListener(executionTarget, id);
 }
 
-export async function addIntentListener(executionTarget: Identity, intent: IntentType): Promise<RemoteIntentListener> {
-    const id = await ofBrowser.executeOnWindow(executionTarget, function (this: TestWindowContext, intentRemote: IntentType): number {
+export async function addIntentListener(executionTarget: Identity, intent: IntentType, listener?: (context: Context) => any): Promise<RemoteIntentListener> {
+    const remoteFn = listener ? await ofBrowser.getOrMountRemoteFunction(executionTarget, listener) : undefined;
+    const id = await ofBrowser.executeOnWindow(executionTarget, function (this: TestWindowContext, intentRemote: IntentType, listenerRemote?: any): number {
```
bryangaleopenfin: `listenerRemote` could be typed as `(context: Context) => Promise<any>`, and then `remoteFn` cast to that as it's passed into `executeOnWindow`.

I guess behind the scenes a `JSHandle` is turned into an actual function, but that's not properly captured by the type system.

It's messy, so I'm marginal on this, fine with you saying you prefer as is

bryangaleopenfin:
Done, back to you 😛

bryangaleopenfin:
n/m, missed your last push, will do another resolving pass!

bryangaleopenfin (commented):
Flagging the last bit of this comment: https://github.com/HadoukenIO/fdc3-service/pull/165#discussion_r346311811. Also, not in favour of the line break changes in `internal.ts` and `helpers.ts`. Break-on-punctuation isn't really a thing we're doing elsewhere

test/demo/utils/fdc3RemoteExecution.ts
ottah: Typed `remoteFn` into a behemoth.

test/demo/raiseIntent/intentResolution.inttest.ts
ottah: My thoughts were that since intent resolution applies to both files but doesn't effect any of the tests it should be its own file. It also is testing a specific behaviour that isn't specced so will probably change in the future so it is easier to alter when split out.

ottah:
@bryangaleOF Changed the documentation to not break on punctuation, I do think some of the lines are a bit too long, I personally prefer docs to be 80-100 columns length but I can go either way on this.

test/demo/utils/fdc3RemoteExecution.ts
bryangaleopenfin: I think as is, the cast is premature, and should happen as `remoteFn` is passed into `executeOnWindow` on line 128, rather than on line 106. Until it's gone through `executeOnWindow`, `remoteFn` is still unambiguously a `JSHandler` and not a callable function, so the cast should go later.

bryangaleopenfin (approved):
On the line break - I'm saying I'd prefer it like the comments on `rasieIntent` and ` broadcast` in `main.ts`, where there's a line break as soon as a word won't fit within a line length limit, like how line-breaking works in any normal text editor. It looks like you're still doing on punctuation, just doing it on full-stops rather than full-stops and commas.

On the separate test file, aside from the data stuff not being in the FDC3 spec, I think the arguments for it being separate would apply to some tests already in `withTarget.ts` or `withoutTarget.ts`. I think as is, things will end up growing messily, with it being arbitrary if someone puts new tests in `intentResolution` or split across `with/withoutTarget`. I can live with it being a separate file, but in that case would much prefer it be renamed to something like  `data.inttest.ts`, so it's clear what it's purpose is, and doesn't have a name that seems redundant given the folder it's in.

PR #166, opened 2019-10-23
[SERVICE-737] Support for non-JavaScript applications

pjbroadbent:
Follow-up to #154. 

Provider supports connections from external adapter connections, and logs them in the model in the same way as windows. Model now tracks the entity type of each registered entity. Renamed some internal types/vars to remove window-specific names from things that cover both windows and external connections.

Adds some unit tests to cover basic external app use-cases.


test/provider/ChannelHandler.unittest.ts
```diff
@@ -144,12 +144,12 @@ describe('When setting the last broadcast context for a channel', () => {
     const testChannel = createMockChannel();
     const testContext = {type: 'test'};
 
-    let testWindow: AppWindow;
+    let testWindow: AppConnection;
```
bryangaleopenfin: As general policy, should this be `testWindow`, or `testConnection`?

test/provider/AppWindow.unittest.ts
```diff
@@ -4,15 +4,23 @@ import {Identity} from 'openfin/_v2/main';
...
 import {createMockChannel} from '../mocks';
 import {Application} from '../../src/client/main';
-import {AbstractAppWindow} from '../../src/provider/model/AppWindow';
+import {AppConnectionBase} from '../../src/provider/model/AppConnection';
 import {ContextChannel} from '../../src/provider/model/ContextChannel';
 import {useMockTime, unmockTime, advanceTime, resolvePromiseChain} from '../utils/unit/time';
+import {EntityType} from '../../src/provider/model/Environment';
 
-class TestAppWindow extends AbstractAppWindow {
+class TestAppWindow extends AppConnectionBase {
```
bryangaleopenfin, line 14: Rename to `TestAppConnection`, or have it always pass `EntityType.WINDOW` to `super`?

src/provider/model/FinAppConnection.ts
```diff
@@ -0,0 +1,38 @@
+import {Identity, Window} from 'openfin/_v2/main';
```
bryangaleopenfin: Window import is unused

src/provider/model/Environment.ts
```diff
@@ -3,7 +3,7 @@ import {Signal} from 'openfin-service-signal';
 
 import {Application} from '../../client/main';
 
-import {AppWindow} from './AppWindow';
+import {AppConnection} from './AppConnection';
 import {ContextChannel} from './ContextChannel';
 
 export enum EntityType {
```
bryangaleopenfin: Should we be using the term "Entity" at all? Couldn't we use "Connection" everywhere we're using "Entity"?

src/provider/controller/ContextHandler.ts
```diff
@@ -29,7 +29,7 @@ export class ContextHandler {
      * @param window Window to send the context to
      * @param context Context to be sent
      */
-    public async send(window: AppWindow, context: Context): Promise<void> {
+    public async send(window: AppConnection, context: Context): Promise<void> {
```
bryangaleopenfin: Are you updating parameter names/variable names/comments?

src/provider/model/FinEnvironment.ts
```diff
@@ -60,25 +64,31 @@ export class FinEnvironment extends AsyncInit implements Environment {
         }
     }
 
-    public wrapApplication(appInfo: Application, identity: Identity, channel: ContextChannel): AppWindow {
+    public wrapApplication(appInfo: Application, identity: Identity, entityType: EntityType, channel: ContextChannel): AppConnection {
```
bryangaleopenfin: My PR renames this to `wrapWindow`. Should this be `wrapConnection`?

test/provider/expect.unittest.ts
```diff
@@ -238,13 +238,81 @@ describe('When creating a non-directory FDC3 app', () => {
     });
 });
 
+describe('When creating an external connection', () => {
+    // External connections won't have a corresponding window. These "windows" MUST have a createdTime of undefined to produce accurate scenerios.
```
bryangaleopenfin: Spelling of "scenerios"

test/provider/expect.unittest.ts
```diff
@@ -238,13 +238,81 @@ describe('When creating a non-directory FDC3 app', () => {
...
+    // Connections from external windows must connect within the WINDOW_EXPECT_TO_CREATED timeout, which is a much stricter requirement than normal OF windows
```
bryangaleopenfin, line 242: Hmm, don't hate me, but did I get this wrong? Maybe we do want to give a more generous `WINDOW_EXPECT_TO_CREATED ` timeout. (Also, should the name of `WINDOW_EXPECT_TO_CREATED` change?

bryangaleopenfin (commented):
First pass - need to spend some more time staring at the logic in FinEnvironment/Model.

How far are you going with eliminating the term window?

src/provider/model/Model.ts
```diff
@@ -252,121 +267,135 @@ export class Model {
...
             }));
 
             // Create a promise that resolves once the window has been created
-            const created = untilTrue(this._environment.windowCreated, () => {
-                return this._environment.isWindowCreated(identity);
+            const windowCreated = untilTrue(this._environment.onWindowCreated, () => {
+                return this._environment.isKnownEntity(identity);
+            });
+            const connectionCreated = untilTrue(this.onConnectionAdded, () => {
+                return this._apiHandler.isClientConnection(identity);
```
bryangaleopenfin, line 455: Think `this._apiHandler.onConnection` would be the correct signal to use here

src/provider/model/Environment.ts
pjbroadbent: I've been taking entity to mean the thing "on the other end of" the connection. As in, the entity connects to the service, and we track the active connections.

It's possibly a rather meaningless distinction though. Open to using connection everywhere if that makes more sense.

In this particular case, the `EntityType` name is because these values come directly from the runtime, which refers to these constants as "entity types". I'm (re-)declaring here for consistency - to use an enum rather than a type union, as we do elsewhere.

src/provider/model/FinEnvironment.ts
pjbroadbent: Yeah, `wrapConnection` IMO. It can remain as `wrapWindow` if it specifically only ever deals with windows, but this method can be used for any type of connection.

test/provider/expect.unittest.ts
pjbroadbent: Ahh, yeah - should rename.

Should maybe increase a little - but depends to what extent. It's currently 200ms for the window creation, followed by 5000ms for the subsequent IAB connection. Bumping the former to 500-1000 ish seems fine, if it's any longer than that then it seems a bit pointless having the two distinct timeouts.

src/provider/model/Model.ts
pjbroadbent: Was using this before, but was causing issues due to the delay between `ApiHandler.onConnection` and `Model.onConnectionAdded` - since the handler in between those signals has async stuff in it.

Have switched back to the ApiHandler signal and all seems fine now though.

test/provider/ChannelHandler.unittest.ts
pjbroadbent: IMO, should only use "window" if it's strictly only ever going to be an OpenFin window.

In this case, the test will only ever use `EntityType.WINDOW` - but the underlying code is intended to be used with connections as well as windows. A bit arbitrary, maybe - but updated this to `testConnection`.

test/provider/AppWindow.unittest.ts
pjbroadbent: Went with the latter

pjbroadbent:
> How far are you going with eliminating the term window?

It's not so much "eliminating", but only using window when that var/function/etc will only ever deal with an OpenFin window, rather than something that's "generic" and deals with any type of connection. For instance, `Environment`/`FinEnvironment` signals are still `onWindow*`.

For this PR, I've made everything within `src/model` consistent. Changes in `src/controller` are only as a result of the renames of things within `model`, but could probably do with a similar review. Can add to this PR, or leave until other PR's are all merged - already feel there's going to be a lot of conflicts here. Latter is my preference, but happy with either.

bryangaleopenfin:
> Can add to this PR, or leave until other PR's are all merged - already feel there's going to be a lot of conflicts here. Latter is my preference, but happy with either.

If it's saved till later, will it actually get done?

src/provider/model/FinEnvironment.ts
bryangaleopenfin: Think do in this PR

src/provider/model/Environment.ts
bryangaleopenfin: When you've got `registerConnection` commented with " Registers an entity in the model" it feels kinda muddled. Enity everywhere might be better, then no confusion with FDC3 connections, and it's in line with runtime terminology

pjbroadbent:
@bryangaleOF Have made some extra renames, use of 'window' etc should now be consistent. Let's discuss if there's any further naming/terminology questions.

src/provider/controller/ContextHandler.ts
```diff
@@ -30,24 +30,24 @@ export class ContextHandler {
     }
 
     /**
-     * Send a context to a specific window. The promise returned may reject if provided window throws an error, so caller should take
+     * Send a context to a specific connection. The promise returned may reject if provided client throws an error, so caller should take
```
bryangaleopenfin: Would change `client` to `entity` or `connection`

src/provider/controller/EventHandler.ts
```diff
@@ -74,8 +74,14 @@ export class EventHandler {
...
-    private dispatchEvent<T extends Events>(targetWindow: AppConnection, eventTransport: Targeted<Transport<T>>): Promise<void> {
-        return this._apiHandler.dispatch(targetWindow.identity, 'event', eventTransport);
+    private dispatchEvent<T extends Events>(target: AppConnection, eventTransport: Targeted<Transport<T>>): Promise<void> {
+        return collateClientCalls([this._apiHandler.dispatch(target.identity, 'event', eventTransport)]).then(([result]) => {
+            if (result === ClientCallsResult.ALL_FAILURE) {
+                console.warn(`Error thrown by client attempting to handle event ${eventTransport.type}, swallowing error`);
+            } else if (result === ClientCallsResult.TIMEOUT) {
+                console.warn(`Timeout waiting for client to handle event ${eventTransport.type}, swallowing error`);
+            }
+        });
```
bryangaleopenfin: Is this a diff-ing error? Surprised this is showing as new

src/provider/model/AppConnection.ts
```diff
@@ -8,63 +8,46 @@ import {IntentType} from '../intents';
...
 /**
- * Model interface, representing a window that has connected to the service.
+ * Model interface, representing an entity that has connected to the service.
  *
- * Only windows that have created intent or context listeners will be represented in this model. If any non-registered
- * window.
- *
- * TODO [SERVICE-737] Review naming of this interface, due to these objects now representing (potentially window-less)
- * external connections. Likewise for "downstream" types/functions/etc.
+ * Since an IAB connection is required for an `AppConnection` object to be created, these will only ever refer to FDC3-enabled windows/applications/etc.
```
bryangaleopenfin: Would use `entity` instead of `windows/applications/etc.`, since I'd say this definitely isn't an application

src/provider/model/AppConnection.ts
```diff
@@ -89,14 +73,22 @@ export abstract class AbstractAppWindow implements AppWindow {
...
     private readonly _onChannelContextListenerAdded: Signal<[ContextChannel]> = new Signal();
 
-    constructor(identity: Identity, appInfo: Application, maturePromise: Promise<void>, channel: ContextChannel, appWindowNumber: number) {
+    constructor(
+        identity: Identity,
+        entityType: EntityType,
+        appInfo: Application,
+        maturePromise: Promise<void>,
+        channel: ContextChannel,
+        appWindowNumber: number
```
bryangaleopenfin: We're using the `appWindowNumber` name in a bunch of places which I feels wrong now

src/provider/model/FinEnvironment.ts
```diff
@@ -64,39 +74,53 @@ export class FinEnvironment extends AsyncInit implements Environment {
...
-    public wrapWindow(liveApp: LiveApp, identity: Identity, channel: ContextChannel): AppWindow {
+    public wrapConnection(liveApp: LiveApp, identity: Identity, entityType: EntityType, channel: ContextChannel): AppConnection {
         identity = parseIdentity(identity);
         const id = getId(identity);
 
-        // If `identity` is an adapter connection, there will not be any `_windows` entry for this identity
-        const window = this._windows.get(id) || {index: this._windowsCreated++};
+        // If `identity` is an adapter connection that hasn't yet connected to the service, there will not be a KnownEntity for this identity
+        // In these cases, we will register the entity now
+        const knownEntity: KnownEntity|undefined = this._knownEntities.get(id) || this.registerEntity(identity, entityType);
```
bryangaleopenfin: Think we're generally putting spaces around `|`

test/provider/expect.unittest.ts
bryangaleopenfin: Nudge on this, doesn't seem to have changed

src/provider/model/Model.ts
```diff
@@ -316,133 +332,154 @@ export class Model {
...
+            delete this._connectionsById[id];
+            this.onConnectionRemoved.emit(connection);
+        } else if (this._expectedConnectionsById[id]) {
+            delete this._expectedConnectionsById[id];
+        }
+    }
+
+    // private findConnectionsByAppId(appId: AppId): AppConnection[] {
+    //     return this.findConnections((connection) => connection.appInfo.appId === appId);
+    // }
```
bryangaleopenfin: Remove

src/provider/model/Model.ts
```diff
@@ -316,133 +332,154 @@ export class Model {
...
+
+    private findConnections(predicate: (connection: AppConnection) => boolean): AppConnection[] {
+        return this.connections.filter(predicate);
+    }
```
bryangaleopenfin: Unused function

src/provider/model/Model.ts
```diff
@@ -316,133 +332,154 @@ export class Model {
...
-            appWindow.removeAllListeners();
+        if (connection) {
+            // Assume client disconnected due to a page reload, or some other change that resets its listener state
+            connection.removeAllListeners();
+
+            // For non-window connections, also treat this as the entity being destroyed
+            // There is no `window-closed` equivilant for external connections, so we will do our full clean-up of state here
+            if (connection.entityType !== EntityType.WINDOW) {
+                this.removeConnection(connection.identity);
+            }
```
bryangaleopenfin: Do you also want to call `setClosed()` on the `LiveApp` and delete it from `this._liveAppsByUuid[uuid]` (same as we do in `onApplicationClosed`)?

src/provider/model/Model.ts
```diff
@@ -316,133 +332,154 @@ export class Model {
...
 
+        if (entityType === EntityType.EXTERNAL_CONNECTION) {
             // Any connections to the service from adapters should be immediately registered
             const appInfo = await this._environment.inferApplication(identity);
-            const liveApp = new LiveApp(Promise.resolve());
+            const liveApp = new LiveApp(undefined);
 
             liveApp.setAppInfo(appInfo);
-            this._liveAppsByUuid[uuid] = liveApp;
+            this._liveAppsByUuid[identity.uuid] = liveApp;
```
bryangaleopenfin: Is this safe? Any danger of multiple external connections having the same UUID?

bryangaleopenfin:
@pjbroadbent  Reviewed. Also think it's worth Ctrl+Shift+F-ing for "client window". Should any of those be changed to entity/connection?

src/provider/model/Model.ts
pjbroadbent: Not quite, as there could be multiple connections from the one UUID.

But added that in with an "if there's no more connections left". Also amended `onApiHandlerConnection`, in case there is already a LiveApp with the current UUID.

src/provider/model/AppConnection.ts
pjbroadbent: Going with `entityIndex`, can discuss naming if you have other suggestions.

bryangaleopenfin: Would still go with number - index suggests literal position in an array

src/provider/model/Model.ts
```diff
@@ -351,7 +351,7 @@ export class Model {
         if (entityType === EntityType.EXTERNAL_CONNECTION) {
             // Any connections to the service from adapters should be immediately registered
             const appInfo = await this._environment.inferApplication(identity);
-            const liveApp = new LiveApp(undefined);
+            const liveApp = this._liveAppsByUuid[identity.uuid] || new LiveApp(undefined);
 
             liveApp.setAppInfo(appInfo);
             this._liveAppsByUuid[identity.uuid] = liveApp;
```
bryangaleopenfin: Suggest check for existing live app first, and only if there isn't one, do `inferApplication`/`setAppInfo`

src/provider/model/Model.ts
```diff
@@ -379,6 +379,12 @@ export class Model {
             // There is no `window-closed` equivilant for external connections, so we will do our full clean-up of state here
             if (connection.entityType !== EntityType.WINDOW) {
                 this.removeConnection(connection.identity);
+
+                const liveApp: LiveApp|undefined = this._liveAppsByUuid[identity.uuid];
```
bryangaleopenfin: ` | `

src/provider/model/Model.ts
```diff
@@ -379,6 +379,12 @@ export class Model {
             // There is no `window-closed` equivilant for external connections, so we will do our full clean-up of state here
             if (connection.entityType !== EntityType.WINDOW) {
```
bryangaleopenfin: Why `connection.entityType !== EntityType.WINDOW` here, but `entityType === EntityType.EXTERNAL_CONNECTION` in `onApiHandlerConnection`?

bryangaleopenfin: And similar in `FinEnvironment`, going with `entityType === EntityType.EXTERNAL_CONNECTION || entityType === EntityType.IFRAME`

pjbroadbent: Standardised to just `EXTERNAL_CONNECTION` using `FinConnection`, with all other types using `FinAppWindow`. May tweak that when we add the `VIEW` type.

src/provider/model/AppConnection.ts
bryangaleopenfin: Would still go with number. 'number' - the number of this connect. 'counter' - the variable that counts total connections

bryangaleopenfin (approved):

tim-dinsdale (approved):
I never want to see another window->connection again

PR #167, opened 2019-10-25
[N/A] Test fixes

bryangaleopenfin:
Fixes for some occasional test failures

### DIFF
--- a/test/demo/contextChannels/broadcastOnChannel.inttest.ts
+++ b/test/demo/contextChannels/broadcastOnChannel.inttest.ts
@@ -34,7 +34,7 @@ describe('When attempting to broadcast on a channel object', () => {
     });
 
     test('If a valid context is provided, the broadcast() resolves successfully', async () => {
-        await expect(channel.broadcast(testContext)).resolves;
+        await channel.broadcast(testContext);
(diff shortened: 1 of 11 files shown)


test/demo/contextChannels/broadcastOnChannel.inttest.ts
```diff
@@ -34,7 +34,7 @@ describe('When attempting to broadcast on a channel object', () => {
     });
 
     test('If a valid context is provided, the broadcast() resolves successfully', async () => {
-        await expect(channel.broadcast(testContext)).resolves;
+        await channel.broadcast(testContext);
```
bryangaleopenfin, line 37: TURNS OUT - `await expect(somePromise).resolves;` doesn't do what I thought it did!

The .resolves doesn't in itself return a promise, you need to call a matcher to get an actual promise, or in this case if you don't care about the result, you can just await the promise

tomrobertsOF: Huh, TIL

tomrobertsOF (approved):

ottah (approved):

PR #169, opened 2019-10-28
[N/A] Resolver fix

bryangaleopenfin:
This restores the original resolver queuing behaviour that got lost with an earlier PR. Of mine 😐. Also adds tests around this behaviour and the small improvement of not ending up with an ever-growing chain of resolved promises.

### DIFF
--- a/src/provider/controller/IntentHandler.ts
+++ b/src/provider/controller/IntentHandler.ts
@@ -45,7 +45,7 @@ export class IntentHandler {
         if (hasTarget(intent)) {
             return this.raiseWithTarget(intent);
         } else {
-            return this.startResolve(intent);
+            return this.startResolve(intent, this.queueResolve.bind(this));
         }
     }
 
@@ -75,7 +75,10 @@ export class IntentHandler {
         }
     }
 
-    private async startResolve(intent: Intent): Promise<IntentResolution> {
+    private async startResolve(
+        intent: Intent,
+        handleAppChoice: (intent: Intent, apps: Application[]) => Promise<IntentResolution>
+    ): Promise<IntentResolution> {
         const apps: Application[] = await this._model.getApplicationsForIntent(intent.type, intent.context.type);
 
         if (apps.length === 0) {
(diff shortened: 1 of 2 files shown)


test/demo/raiseIntent/withoutTarget.inttest.ts
```diff
@@ -323,14 +391,33 @@ async function raiseIntentAndExpectResolverToShow(intent: Intent): Promise<Boxed
 }
 
 /**
- * Closes the resolver by remotely clicking the Cancel button in it
+ * Rremotely clicks the cancel button on the resolver, and checks the resolver closes
```
tomrobertsOF: `Rremotely`. Also on line 405 below.

src/provider/controller/IntentHandler.ts
```diff
@@ -86,34 +89,51 @@ export class IntentHandler {
...
     }
 
     private async queueResolve(intent: Intent, applications: Application[]): Promise<IntentResolution> {
         if (this._resolvePromise) {
             console.log(`Resolver showing, re-resolving intent '${intent.type}' when resolver closes'`);
 
-            this._resolvePromise = this._resolvePromise.catch(() => {}).then(() => this.startResolve(intent));
-
-            return this._resolvePromise;
+            this._resolvePromise = this._resolvePromise.catch(() => {}).then(() => this.startResolve(intent, this.showResolver.bind(this)));
```
tomrobertsOF, line 102: Don't we have an `allowReject` async util somewhere to avoid the `catch(()=>{})`?

test/demo/raiseIntent/withoutTarget.inttest.ts
bryangaleopenfin: ✅

src/provider/controller/IntentHandler.ts
bryangaleopenfin: We do but it's not quite the same
`allowReject(promise)` - a promise that resolves or rejects when the promise resolves or rejects respectively (but rejection won't print an error to the console)
`promise.catch(() => {})` - a promise that resolves when the promise either resolves or rejects

tomrobertsOF: Hmmm, ok, that makes sense.

tomrobertsOF (approved):

ottah (approved):

PR #174, opened 2019-11-12
[SERVICE-555] Ensure provider API calls to client are robust against a misbehaving client

bryangaleopenfin:
This PR
- Modifies error codes to match actual failure cases, rather than FDC3 spec (which kinda didn't make sense). Any reconciliation is saved for SERVICE-561. We new have a group of call-specific error codes, and a group of "generic" error codes
- Uses new error codes in `open` and `raiseIntent`
- Guards against calls from the service to the client erroring or failing to return quickly, generally taking the PoV that if one context/intent handler succeeds, even if others fail/timeout, that's enough to be a success
- Updates `ofPuppeteer.ts` from Notifications (I was going to use the new function mounting method, but then didn't, but it seemed worth keeping this change)

I'll need to re-work some things if this goes in after PR-165

### DIFF
--- a/src/client/errors.ts
+++ b/src/client/errors.ts
@@ -3,7 +3,8 @@
  */
 
 /**
- * Error codes specific to the [[open]] function.
+ * Error codes specific to the [[open]] function. Note that this is not an exhaustive list of error codes that can be returned by [[open]],
+ * which may return any of the generic error codes, such as ApplicationError.
  */
 export enum OpenError {
     /**
@@ -31,7 +32,8 @@ export enum OpenError {
 }
 
 /**
- * Error codes specific to the [[raiseIntent]] function.
+ * Error codes specific to the [[raiseIntent]] function. Note that this is not an exhaustive list of error codes that can be returned by
+ * [[raiseIntent]], which may return any of the generic error codes, such as ApplicationError.
  */
 export enum RaiseIntentError {
     /**
--- a/src/client/errors.ts
+++ b/src/client/errors.ts
@@ -22,6 +22,10 @@ export enum OpenError {
      * Currently unused.
      */
     ResolverUnavailable = 'ResolverUnavailable',
+
+    SendContextError = 'SendContextError',
+    SendContextTimeout = 'SendContextTimeout',
+    SendContextNoHandler = 'SendContextNoHandler'
 }
 
 /**
@@ -57,7 +61,7 @@ export enum ResolveError {
     /**
      * The intent resolver UI was dismissed by the user, so the intent has been cancelled.
      */
-    ResolverClosedOrCancelled = 'ResolverClosedOrCancelled',
+    ResolverClosedOrCancelled = 'ResolverClosedOrCancelled'
 }
 
 /**
--- a/src/client/errors.ts
+++ b/src/client/errors.ts
@@ -61,7 +61,11 @@ export enum ResolveError {
     /**
      * The intent resolver UI was dismissed by the user, so the intent has been cancelled.
      */
-    ResolverClosedOrCancelled = 'ResolverClosedOrCancelled'
+    ResolverClosedOrCancelled = 'ResolverClosedOrCancelled',
+
+    SendIntentError = 'SendIntentError',
+    SendIntentTimeout = 'SendIntentTimeout',
+    SendIntentNoHandler = 'SendContextNoHandler'
 }
 
 /**
@@ -72,6 +76,11 @@ export enum ChannelError {
      * Indicates that [[getChannelById]] has failed because no such channel exists with the given ID.
      */
     ChannelDoesNotExist = 'ChannelDoesNotExist'
+}
+
+export enum JoinError {
+    SendContextError = 'SendContextError',
+    SendContextTimeout = 'SendContextTimeout',
 }
 
 /**
--- a/src/client/errors.ts
+++ b/src/client/errors.ts
@@ -3,65 +3,72 @@
  */
 
 /**
- * Error codes returned by actions involving creating applications.
+ * Error codes specific to the [[open]] function
  */
 export enum OpenError {
     /**
-     * Indicates that an application cannot be found in the application directory.
+     * Indicates that the an application of the provided name cannot be found in the application directory.
      */
     AppNotFound = 'AppNotFound',
 
     /**
-     * Currently unused.
+     * In the case where the optional `context` argument is provided, indicates that the provided application was started, but threw an
+     * error when attempting to handle the provided context.
      */
-    ResolverUnavailable = 'ResolverUnavailable',
-
     SendContextError = 'SendContextError',
+
+    /**
+     * In the case where the optional `context` argument is provided, indicates that the given application was started, but a timeout was
+     * reached waiting for it to handle the provided context.
+     */
     SendContextTimeout = 'SendContextTimeout',
+
+    /**
+     * In the case where the optional `context` argument is provided, indicates that the given application was started, but did not add a
+     * context handler.
+     */
     SendContextNoHandler = 'SendContextNoHandler'
 }
 
 /**
- * Error codes returned after failure to fire intents to applications.
+ * Error codes specific to the [[raiseIntent]] function
  */
-export enum ResolveError {
+export enum RaiseIntentError {
     /**
-     * Indicates that no apps could be found for a particular intent.
+     * Indicates that no apps could be found to handle the provided intent and context.
      */
     NoAppsFound = 'NoAppsFound',
     /**
-     * Indicates that, in the case when a 'target' argument is passed to [[raiseIntent]], no
-     * such app either exists in the application directory or is currently running.
+     * In the case when a the optional 'target' argument is provided, no such app either exists in the application directory or is
+     * currently running.
      */
     TargetAppNotAvailable = 'TargetAppNotAvailable',
     /**
-     * Indicates that, in the case when a 'target' argument is passed to [[raiseIntent]], the app is not able to handle this intent.
+     * In the case when a the optional 'target' argument is provided, indicates that the app is not able to handle this intent and context.
      */
     TargetAppDoesNotHandleIntent = 'TargetAppDoesNotHandleIntent',
-    /**
-     * Indicates that the provider has started an app to receive an intent, and that it has timed out whilst waiting
-     * for that app to handle the intent.
-     */
-    IntentTimeout = 'IntentTimeout',
-    /**
-     * Currently unused.
-     */
-    ResolverUnavailable = 'ResolverUnavailable',
-    /**
-     * Currently unused.
-     */
-    ResolverTimeout = 'ResolverTimeout',
     /**
      * The intent resolver UI was dismissed by the user, so the intent has been cancelled.
      */
     ResolverClosedOrCancelled = 'ResolverClosedOrCancelled',
-
+    /**
+     * Indicates that an application was started, but threw an error when attempting to handle the provided intent and context.
+     */
     SendIntentError = 'SendIntentError',
+    /**
+     * Indicates that an application was started, but a timeout was reached waiting for it to handle the provided intent and context.
+     */
     SendIntentTimeout = 'SendIntentTimeout',
+    /**
+     * Indicates that an application was started, but did not add a handler for the provided intent.
+     */
     SendIntentNoHandler = 'SendContextNoHandler'
 }
 
-export enum LaunchError {
+/**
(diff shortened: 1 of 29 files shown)


src/provider/utils/async.ts
```diff
@@ -1,10 +1,19 @@
 import {Signal} from 'openfin-service-signal';
 
 import {DeferredPromise} from '../common/DeferredPromise';
+import {Timeouts} from '../constants';
+
+export enum CollateApiCallResultsResult {
```
pjbroadbent: Can we call this something more concise? `CollateResult`? `ApiResult`?..

src/provider/utils/async.ts
```diff
@@ -1,10 +1,19 @@
...
+    AnySuccess,
+    AllFailure,
+    Timeout,
+    NoCalls
```
pjbroadbent: Can we use `CONST_CASE` for these? All enums outside of error codes (which we should look at as part of the review story) use const case.

src/client/main.ts
```diff
@@ -420,24 +420,53 @@ function deserializeChannelChangedEvent(eventTransport: Transport<ChannelChanged
...
+            }));
+
+            if (failures > 0 && successes === 0) {
+                throw new Error(`All ${payload.intent} intent handlers failed`);
+            }
         });
 
         channelClient.register(APIToClientTopic.RECEIVE_CONTEXT, (payload: ReceiveContextPayload) => {
+            let successes = 0;
+            let failures = 0;
```
pjbroadbent: Can we avoid the duplication here? (and in contextChannels.ts too)

Should the `collateApiCallResults` util be in `client/internal.ts`, much like the error code utils?

src/client/main.ts
```diff
@@ -420,24 +420,53 @@ function deserializeChannelChangedEvent(eventTransport: Transport<ChannelChanged
...
 
         const eventHandler = getEventRouter();
 
         channelClient.register('event', (eventTransport: Targeted<Transport<Events>>) => {
-            eventHandler.dispatchEvent(eventTransport);
+            try {
+                eventHandler.dispatchEvent(eventTransport);
+            } catch (e) {
+                console.warn(`Error thrown dispatching ${eventTransport.type} event, rethrowing error. Error message: ${e.message}`);
+                throw e;
```
pjbroadbent, line 448: General point, but sticking it here.

When re-throwing an error like this, does it preserve the initial stack trace? IIRC, not preserving the stack trace is of those special JavaScript "features"...

If it's overwritten, we should likely trace the initial stack trace as well as the errors (maybe in a collapsible way - `console.group` and stuff).

src/provider/controller/ContextHandler.ts
```diff
@@ -29,20 +30,26 @@ export class ContextHandler {
     }
 
     /**
-     * Send a context to a specific app
+     * Send a context to a specific window. The promise returned may reject if provided window throws an error, so caller should take
```
pjbroadbent, line 33: > The promise returned may reject

Assuming this is referring to the `if (window.hasContextListener())` case, rather than the `else`? Don't see how the latter could throw.

Is it strange that some failures are handled, and some are propagated up the call stack?

src/provider/utils/async.ts
```diff
@@ -81,6 +91,39 @@ export function allowReject<T>(promise: Promise<T>): Promise<T> {
...
 }
 
+/**
+ * Takes multiple promises representing API calls, and reduces them to a single result. In the case that more than one promise resolves to
+ * a result, the first to result will be used. Intented to be used when calling a client from the provider, to protect against a
+ * misbehaving client.
+ *
+ * @param promises An array of promises
+ */
+export async function collateApiCallResults<T = void>(promises: Promise<T>[]): Promise<[CollateApiCallResultsResult, T | undefined]> {
```
pjbroadbent: Is resolving in the case of an error resulting in a bit of an anti-pattern here?..

I understand that the calling code will need to add context to any failures that occur, but would that be better done by this throwing an `FDC3Error` that can be (optionally) caught and decorated further up the stack?

I guess it's largely a question of semantics, would need to see which approach looks better when actually used in practice(?)

src/client/errors.ts
```diff
@@ -3,79 +3,98 @@
...
-    AppTimeout = 'AppTimeout',
+    SendContextTimeout = 'SendContextTimeout',
+
     /**
-     * Currently unused.
+     * In the case where the optional `context` argument is provided, indicates that the provided application was started, but did not add
+     * a context handler.
      */
-    ResolverUnavailable = 'ResolverUnavailable',
+    SendContextNoHandler = 'SendContextNoHandler'
```
pjbroadbent, line 31: This may be a question for the error code review story, but are these getting a bit too granular?

Is an application going to care about the distinction between these three codes? I definitely want us to be as detailed as possible in our error reporting, but the distinction between these three is something that could be conveyed in the message rather than by the code. The main purpose of the codes is to let applications handle errors programmatically, and I don't really see how an app would want/need to distinguish between these cases.

Feels like we're getting into "code and sub-code" territory here, which could be a decent compromise. Error code would be "problem starting app", sub-code would be "handler threw an error", "no handler registered", "handler timeout", etc.

src/client/errors.ts
```diff
@@ -3,79 +3,98 @@
...
+export enum RaiseIntentError {
     /**
-     * Indicates that no apps could be found for a particular intent.
+     * Indicates that no application could be found to handle the provided intent and context.
      */
     NoAppsFound = 'NoAppsFound',
     /**
-     * Indicates that, in the case when a 'target' argument is passed to [[raiseIntent]], no
-     * such app either exists in the application directory or is currently running.
+     * In the case when a the optional 'target' argument is provided, no such application either exists in the application directory or is
```
pjbroadbent: > when a the optional

(and again below)

src/client/errors.ts
```diff
@@ -3,79 +3,98 @@
...
     /**
-     * The intent resolver UI was dismissed by the user, so the intent has been cancelled.
+     * Indicates that an application was started, but did not add a handler for the provided intent.
      */
-    ResolverClosedOrCancelled = 'ResolverClosedOrCancelled',
+    SendIntentNoHandler = 'SendIntentNoHandler'
+}
+
+/**
+ * Error codes relating to launching applications.
```
pjbroadbent, line 71: Again, possibly one for another story, but bugs me that we're mixing our approaches here. Seems weird that some error enums are API-specific (but coverage is far smaller than the API as a whole), and some error enums are "generic".

For `ApplicationError` specifically, for example - it's not clear if `open` or `raiseIntent` can throw `ApplicationError`s. Both those APIs can start applications, but these errors don't exist in the API-specific enums above. Or, possibly worse, they have duplicate error codes for the same error - making it hard to handle errors in a somewhat generic way.

test/demo/raiseIntent/withTarget.inttest.ts
```diff
@@ -131,8 +132,8 @@ listener to be added', async () => {
                         );
 
                         await expect(raiseIntentPromise).toThrowFDC3Error(
-                            ResolveError.IntentTimeout,
-                            `Timeout waiting for intent listener to be added for intent: ${validIntent.type}`
+                            RaiseIntentError.SendIntentNoHandler,
+                            `No intent handler added for intent: ${validIntent.type}`
```
pjbroadbent, line 136: Do the usual "maturity" rules apply here?

Will reject instantly with "no handler" error if a mature app has no listeners, if an immature app has no listeners, it's given the remainder of it's "startup" period in which to add one?

src/provider/utils/async.ts
bryangaleopenfin: ✅

src/provider/controller/ContextHandler.ts
bryangaleopenfin: It's not ideal, hence it being flagged specifically in the comment. But I'm not seeing a nicer way of handling this, given the different callers of send, without a lot of duplication

src/provider/utils/async.ts
bryangaleopenfin: With this I'd rather force the caller to consider failure, which they'd be able to just ignore if this throws exceptions. And I don't think they'd be any case where would could just let the error bubble up

src/client/errors.ts
bryangaleopenfin: ✅

src/client/errors.ts
bryangaleopenfin: Some of the errors very much are API call specific and make sense to present as such, but if we were to go along with that with full consistency, then we're getting into making call-specific TypeErrors, for instance. And I think duplicate error codes are easy enough to avoid

test/demo/utils/delay.ts
```diff
@@ -13,6 +13,7 @@ export enum Duration {
     PAGE_NAVIGATE = 500,
     SHORTER_THAN_APP_MATURITY = 2000,
     LONGER_THAN_APP_MATURITY = 6000,
+    LONGER_THAN_SERVICE_TO_CLIENT_API_CALL_TIMEOUT = 6000,
```
tomrobertsOF, line 16: 🤣

tomrobertsOF: Can we just combine some of these into a single "long delay"

test/demo/raiseIntent/withTarget.inttest.ts
bryangaleopenfin: Maturity rules still apply, I just decided not to phrase this as a timeout, given we're using that term for timeouts on client calls, and it doesn't quite fit so well when we're making a call to a long-running app rather than one that's just started

src/provider/utils/async.ts
tomrobertsOF: If this is going to be in the async utils, can it be a bit more generic. I.e. not explicitly referencing API calls or externally defined timeouts. I expect we'll be making this a standalone module soon and would prefer to keep it as independent as possible for when that day comes.

bryangaleopenfin: In that case I'd rather go the other direction - keep this specific to API calls, but move it somewhere other than `async`. (And if it makes sense, pull out a generic part of the implementation into `async`). Happy for me to take that approach?

test/demo/utils/delay.ts
bryangaleopenfin: As is, this communicates something about why we're delaying, which a single "long delay" wouldn't. Would prefer to leave as is

tomrobertsOF: Hmmm. I just seems unnecessarily verbose, but not worth arguing over.

src/provider/utils/async.ts
tomrobertsOF: Yea, that's fine. My metric would be anything which would be useful to any random block of promises should stay here, and anything that is only really useful in the context of the services belongs elsewhere.

src/client/main.ts
bryangaleopenfin: Removed the client side duplication, but still distinct from the "collate" call on the provider side. Think they are sufficiently different for it to not make sense to unify the two

✅

src/provider/utils/async.ts
bryangaleopenfin: ✅

Moved into `provider/utils/helpers.ts`

src/provider/utils/async.ts
bryangaleopenfin: Renamed the function to `CollateClientCalls`, so this is naturally a bit shorter

tomrobertsOF (approved):

src/client/main.ts
bryangaleopenfin: This is the only place we're rethrowing in this PR, but I think it's kinda irrelevant to your point - in this case it gets thrown within the API handler, so bubbles up to the provider, where it then gets captured and logged. It's not possible for a client developer to get direct access to this error object anyway

What might be worthwhile is always logging the stack trace on the client side? But I think that's a separate question from whether stack traces are overwritten on re-throw

pjbroadbent: Didn't realise this was in the client - true, a bit more specific than I was thinking.

Makes sense as-is.

src/client/errors.ts
pjbroadbent: Will address in SERVICE-561

src/client/errors.ts
pjbroadbent: Not completely sure I agree - but yeah, the answer could just be some clarifying comments. Can decide as part of SERVICE-561.

pjbroadbent (approved):

PR #179, opened 2019-11-14
[SERVICE-791] Revisit return type of subscription methods

bryangaleopenfin:
Modifies `addContextListener`/`addEventListener` on our ChannelObject to return synchronously, and changes `broadcast` on our Channel object to return asynchronously, to match equivalent calls in `main`

Also does some tidying of how intent handlers are invoked in demo apps

### DIFF
--- a/src/demo/apps/NewsApp.tsx
+++ b/src/demo/apps/NewsApp.tsx
@@ -33,19 +33,19 @@ export function NewsApp(): React.ReactElement {
             }
         });
 
-        const intentListenerPromise = fdc3.addIntentListener(fdc3.Intents.VIEW_NEWS, (context: Context): void => {
+        const intentListener = fdc3.addIntentListener(fdc3.Intents.VIEW_NEWS, (context: Context): void => {
             handleIntent(context as InstrumentContext);
         });
 
-        const contextListenerPromise = fdc3.addContextListener((context: Context): void => {
+        const contextListener = fdc3.addContextListener((context: Context): void => {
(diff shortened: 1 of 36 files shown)


src/demo/apps/NewsApp.tsx
```diff
@@ -33,7 +33,7 @@ export function NewsApp(): React.ReactElement {
             }
         });
 
-        const intentListener = fdc3.addIntentListener(fdc3.Intents.VIEW_NEWS, (context: Context): Promise<void> => {
+        const intentListenerPromise = fdc3.addIntentListener(fdc3.Intents.VIEW_NEWS, (context: Context): Promise<void> => {
             return new Promise((resolve, reject) => {
                 try {
                     handleIntent(context as InstrumentContext);
```
ottah: Side note but can't all of these listeners in the component just be:
```ts
async (context: Context) => {
	await handleIntent(context as InstrumentContext)
}
```

ottah (approved):
Looks good.

src/demo/apps/NewsApp.tsx
bryangaleopenfin: Hmm...

Yeah, not sure why this is like this. Will tidy and see if anything breaks

bryangaleopenfin: Sorted in News, Dialer, and Charts.

ottah (approved):

tomrobertsOF (approved):

PR #180, opened 2019-11-15
[SERVICE-561] Review error codes

bryangaleopenfin:


src/client/errors.ts
```diff
@@ -3,82 +3,57 @@
  */
 
 /**
- * Error codes specific to the [[open]] function. Note that this is not an exhaustive list of error codes that can be returned by [[open]],
- * which may return any of the generic error codes, such as ApplicationError.
- */
```
pjbroadbent, line 8: Extra space before the `*`s

src/client/errors.ts
```diff
@@ -3,82 +3,57 @@
...
-     * Indicates that an application could not be started from an OpenFin manifest.
+     * Indicates that the target application has no windows that have a relevant handler for the given context.
      */
-    ErrorOnLaunch = 'ErrorOnLaunch',
+    NoHandler = 'SendContextError:SendContextError:NoHandler',
     /**
-     * Indicates that a timeout was reached before the application was started.
+     * Indicates that all handlers for the given context threw an error when invoked.
+     */
+    HandlerError = 'SendContextError:SendContextError:HandlerError',
```
pjbroadbent: Double prefix on this and `NoHandler`

src/client/errors.ts
```diff
@@ -88,17 +63,17 @@ export enum ChannelError {
...
 /**
  * Error codes relating to OpenFin windows.
  */
 export enum IdentityError {
     /**
-     * Indicates that a window with a provided OpenFin Identity cannot be found.
+     * Indicates that no window with a provided OpenFin Identity is registered with the FDC3 service.
      */
-    WindowWithIdentityNotFound = 'WindowWithIdentityNotFound'
+    WindowWithIdentityNotFound = 'IdentityError:WindowWithIdentityNotFound'
```
pjbroadbent: Make this less window-specific, eg `ConnectionNotFound`

src/client/main.ts
```diff
@@ -158,6 +158,9 @@ const contextListeners: ContextListener[] = [];
  * ```
  * @param name The [[AppName]] to launch.
  * @param context A context to pass to the app post-launch.
+ * @throws [[FDC3Error]] with an [[ApplicationError]] code.
+ * @throws If `context` is passed, [[FDC3Error]] with a [[SendContextError]] code.
```
pjbroadbent, line 162: Does this line-up better in the docs if it's flipped round?

```
@throws [[FDC3Error]] with a [[SendContextError]] code, if `context` is passed.
```

src/client/errors.ts
bryangaleopenfin: ✅

src/client/errors.ts
bryangaleopenfin: ✅

src/client/errors.ts
bryangaleopenfin: ✅

src/client/main.ts
bryangaleopenfin: I think visually that's better, but as a sentence, it's more natural to put that first. It also gets trickier with the `TypeErrors` - since that isn't our class to document, I've put more in the `@throws` line, and I'm following the pattern:

If [circumstances this error is relevant] [error type], if [specific cause of error if not documented in linked error class]

bryangaleopenfin:
@pjbroadbent Done with first round of feedback

src/client/errors.ts
pjbroadbent: Based on outcomes of SERVICE-737, kinda changed my mind about this one. For the client API, I think it makes more sense if we just leave everything in terms of 'windows'.

Sorry :-)

ngunday (approved):

src/client/errors.ts
bryangaleopenfin: On intuition I've left the enum as `ConnectionError`, but then the specific error code is about a windows

bryangaleopenfin:
@pjbroadbent Boop

pjbroadbent (approved):

PR #182, opened 2019-11-18
[OC-1990] launch FDC3 provider and inject client API from fdc3.asar

wenjunche:
[Core PR](https://gitlab.com/openfin/core/merge_requests/1124)

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -53,10 +53,6 @@ export function getServicePromise(): Promise<ChannelClient> {
     if (!channelPromise) {
         if (typeof fin === 'undefined') {
             channelPromise = Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.'));
-        } else if (fin.Window.me.uuid === getServiceIdentity().uuid && fin.Window.me.name === getServiceIdentity().name) {
-            // Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
-            // that includes this, but for now it is easier to put a guard in place.
-            channelPromise = Promise.reject<ChannelClient>(new Error('Trying to connect to provider from provider'));
         } else {
             channelPromise = new Promise<ChannelClient>((resolve, reject) => {
                 // TODO: just use RuntimeInfo once its type is updated from js v2 API
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -15,7 +15,7 @@ import {EventEmitter} from 'events';
 
 import {ChannelClient} from 'openfin/_v2/api/interappbus/channel/client';
 
-import {APIFromClientTopic, SERVICE_CHANNEL, setServiceChannel, SERVICE_IDENTITY, APIFromClient, deserializeError, Events} from './internal';
+import {APIFromClientTopic, getServiceChannel, setServiceChannel, getServiceIdentity, setServiceIdentity, APIFromClient, deserializeError, Events} from './internal';
 import {EventRouter} from './EventRouter';
 
 /**
@@ -52,7 +52,7 @@ export function getServicePromise(): Promise<ChannelClient> {
     if (!channelPromise) {
         if (typeof fin === 'undefined') {
             channelPromise = Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.'));
-        } else if (fin.Window.me.uuid === SERVICE_IDENTITY.uuid && fin.Window.me.name === SERVICE_IDENTITY.name) {
+        } else if (fin.Window.me.uuid === getServiceIdentity().uuid && fin.Window.me.name === getServiceIdentity().name) {
             // Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
             // that includes this, but for now it is easier to put a guard in place.
             channelPromise = Promise.reject<ChannelClient>(new Error('Trying to connect to provider from provider'));
@@ -60,14 +60,13 @@ export function getServicePromise(): Promise<ChannelClient> {
             channelPromise = new Promise<ChannelClient>((resolve, reject) => {
                 fin.System.getRuntimeInfo().then((info: any) => {
                     if (info.fdc3AppUuid && info.fdc3ChannelName) {
-                        SERVICE_IDENTITY.uuid = info.fdc3AppUuid;
-                        SERVICE_IDENTITY.name = info.fdc3AppUuid;
+                        setServiceIdentity(info.fdc3AppUuid);
                         setServiceChannel(info.fdc3ChannelName);
                     }
-                    if (fin.Window.me.uuid === SERVICE_IDENTITY.uuid && fin.Window.me.name === SERVICE_IDENTITY.name) {
+                    if (fin.Window.me.uuid === getServiceIdentity().uuid && fin.Window.me.name === getServiceIdentity().name) {
                         reject(new Error('Trying to connect to provider from provider'));
                     } else {
-                        fin.InterApplicationBus.Channel.connect(SERVICE_CHANNEL, {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
+                        fin.InterApplicationBus.Channel.connect(getServiceChannel(), {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
                             // Register service listeners
                             channel.register('WARN', (payload: unknown) => console.warn(payload));  // tslint:disable-line:no-any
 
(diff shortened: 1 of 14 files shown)


src/provider/model/Model.ts
```diff
@@ -53,7 +53,7 @@ export class Model {
     private readonly _onWindowRegisteredInternal: Signal<[AppWindow]> = new Signal();
 
     constructor(
-        @inject(Inject.APP_DIRECTORY) directory: AppDirectory,
+    @inject(Inject.APP_DIRECTORY) directory: AppDirectory,
```
bryangaleopenfin: Indentation

src/provider/index.ts
```diff
@@ -37,7 +37,7 @@ export class Main {
     private readonly _model: Model;
 
     constructor(
-        @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>,
+    @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>,
```
bryangaleopenfin: Indentation

src/provider/controller/IntentHandler.ts
```diff
@@ -20,7 +20,7 @@ export class IntentHandler {
     private _resolvePromise: Promise<IntentResolution> | null;
 
     constructor(
-        @inject(Inject.MODEL) model: Model,
+    @inject(Inject.MODEL) model: Model,
```
bryangaleopenfin: Indentation

src/provider/controller/EventHandler.ts
```diff
@@ -17,7 +17,7 @@ export class EventHandler {
     private readonly _apiHandler: APIHandler<APIFromClientTopic>;
 
     constructor(
-        @inject(Inject.CHANNEL_HANDLER) channelHandler: ChannelHandler,
+    @inject(Inject.CHANNEL_HANDLER) channelHandler: ChannelHandler,
```
bryangaleopenfin: Indentation

src/provider/controller/ContextHandler.ts
```diff
@@ -20,7 +20,7 @@ export class ContextHandler {
     private readonly _model: Model;
 
     constructor(
-        @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>,
+    @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>,
```
bryangaleopenfin: Indentation

src/client/internal.ts
```diff
@@ -19,14 +19,17 @@ import {FDC3Error} from './errors';
...
-    uuid: 'fdc3-service',
+    uuid: 'fdc3-service',  // default value
     name: 'fdc3-service'
 };
 
 /**
  * Name of the IAB channel use to communicate between client and provider
  */
-export const SERVICE_CHANNEL = 'of-fdc3-service-v1';
+export let SERVICE_CHANNEL: string = 'of-fdc3-service-v1';
```
bryangaleopenfin: Would prefer for consistency if we had
`getServiceChannel`, `setServiceChannel`, `getServiceIdentity`, `setServiceIdentity`

Would also change `SERVICE_CHANNEL` and `SERVICE_IDENTITY` to camcelCase since they're not constant, and not export them 

Would also prefer to keep functions together at end of file

src/client/connection.ts
```diff
@@ -57,14 +57,27 @@ export function getServicePromise(): Promise<ChannelClient> {
...
-                // Register service listeners
-                channel.register('WARN', (payload: unknown) => console.warn(payload));  // tslint:disable-line:no-any
+            channelPromise = new Promise<ChannelClient>((resolve, reject) => {
+                fin.System.getRuntimeInfo().then((info: any) => {
+                    if (info.fdc3AppUuid && info.fdc3ChannelName) {
+                        SERVICE_IDENTITY.uuid = info.fdc3AppUuid;
+                        SERVICE_IDENTITY.name = info.fdc3AppUuid;
+                        setServiceChannel(info.fdc3ChannelName);
+                    }
+                    if (fin.Window.me.uuid === SERVICE_IDENTITY.uuid && fin.Window.me.name === SERVICE_IDENTITY.name) {
```
bryangaleopenfin: Do we definitely still want the check on line 55? Would get rid of that branch if it's now redundant

src/client/connection.ts
```diff
@@ -57,14 +57,27 @@ export function getServicePromise(): Promise<ChannelClient> {
             // that includes this, but for now it is easier to put a guard in place.
             channelPromise = Promise.reject<ChannelClient>(new Error('Trying to connect to provider from provider'));
         } else {
-            channelPromise = fin.InterApplicationBus.Channel.connect(SERVICE_CHANNEL, {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
-                // Register service listeners
-                channel.register('WARN', (payload: unknown) => console.warn(payload));  // tslint:disable-line:no-any
+            channelPromise = new Promise<ChannelClient>((resolve, reject) => {
+                fin.System.getRuntimeInfo().then((info: any) => {
```
bryangaleopenfin: Would make the typing more explicit - `(info: RuntimeInfo & {fdc3AppUuid?: string, fdc3ChannelName?: string})` instead of `(info: any)`.

We should use just `RuntimeInfo` once we're able, probably as part of SERVICE-840. Would add a line `// TODO: Change to use just RuntimeInfo when we have updated types [SERVICE-840]` above this

src/provider/model/FinEnvironment.ts
```diff
@@ -93,10 +93,10 @@ export class FinEnvironment extends AsyncInit implements Environment {
             const application = fin.Application.wrapSync(identity);
             const applicationInfo = await application.getInfo();
 
-            const {shortcut, startup_app} = applicationInfo.manifest as OFManifest;
+            const {shortcut, startup_app} = (applicationInfo.manifest || {}) as OFManifest;
```
bryangaleopenfin: If the `ApplicationInfo` interface is inaccurate, and `manifest` can be null/undefined, would comment that here (with another TODO referencing 840 if we expect the types to be updated soon).

Would also change the `OFManifest` interface, so we have `startup_app?` rather than `startup_app`, otherwise casting `{}` to `OFManifest` is potentially misleading

src/client/internal.ts
```diff
@@ -19,14 +19,17 @@ import {FDC3Error} from './errors';
  * The identity of the main application window of the service provider
  */
 export const SERVICE_IDENTITY = {
-    uuid: 'fdc3-service',
+    uuid: 'fdc3-service',  // default value
```
bryangaleopenfin: Wonder if it would be more helpful to slightly change the block comment above this, rather than have this line comment. Suggests only `uuid` can change, when `name` can also change

bryangaleopenfin (commented):
A few comments, but mostly stray indentation changes, and minor fixes to bring more in line with project conventions

bryangaleopenfin:
@wenjunche Flagging that @tim-dinsdale has made roughly the same changes in [his branch](https://github.com/HadoukenIO/fdc3-service/pull/189), and to add the confusion, I've resolved comments on PR-182 as though they were on PR-189

src/client/connection.ts
```diff
@@ -52,19 +53,32 @@ export function getServicePromise(): Promise<ChannelClient> {
     if (!channelPromise) {
         if (typeof fin === 'undefined') {
             channelPromise = Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.'));
-        } else if (fin.Window.me.uuid === SERVICE_IDENTITY.uuid && fin.Window.me.name === SERVICE_IDENTITY.name) {
+        } else if (fin.Window.me.uuid === getServiceIdentity().uuid && fin.Window.me.name === getServiceIdentity().name) {
```
bryangaleopenfin: I don't think we need this check any more? We do this check again further down

src/provider/model/Model.ts
```diff
@@ -66,7 +66,7 @@ export class Model {
     private readonly _onConnectionRegisteredInternal: Signal<[AppConnection]> = new Signal();
 
     constructor(
-        @inject(Inject.APP_DIRECTORY) directory: AppDirectory,
+        @inject(Inject.APP_DIRECTORY) directory: AppDirectory, // eslint-disable-line @typescript-eslint/indent
```
bryangaleopenfin: All other places are doing this with `// eslint-disable-next-line`, would prefer consistency

src/client/connection.ts
wenjunche: removed

src/provider/model/Model.ts
wenjunche: updated

tim-dinsdale (approved):

bryangaleopenfin (approved):

PR #184, opened 2019-11-25
[SERVICE-782] Move FDC3 and Notifications to latest runtime

bryangaleopenfin:
Discovered bug [SERVICE-835] during manual testing, but verified was not introduced with the runtime upgrade


tomrobertsOF:
Not sure if this was discussed elsewhere, but I think we should also upgrade the `@types/openfin` package when we bump the runtime to keep them in sync.

tomrobertsOF (approved):

rypete (approved):

PR #185, opened 2019-12-02
[N/A] Distributed app directory groundwork

bryangaleopenfin:
This PR is the first tracer bullet into the "distributed app directory" feature. This is not intentionally written to complete any particular story, though I claim it completes [SERVICE-821], so please review with that in mind.

With this PR, a client may set a domain's app directory, and have that app directory used by the service.


src/provider/model/Model.ts
```diff
@@ -308,6 +310,19 @@ export class Model {
         }
     }
 
+    public async onDirectoryChanged(): Promise<void> {
+        await parallelForEach(Object.entries(this._liveAppsByUuid).filter((entry) => !entry[1].mature && !entry[1].hasFinalAppInfo()), async (entry) => {
```
tomrobertsOF: very minor, but this line is really long. Can it be formatted a bit more readably somehow?

test/demo/appDirectory.inttest.ts
```diff
@@ -0,0 +1,126 @@
...
+});
+
+setupTeardown();
+
+describe('When running a directory app', () => {
+    setupOpenDirectoryAppBookends(testAppInDirectory1);
+
+    afterEach(async () => {
+        await clearDirectoryShard();
+        await quitApps(testAppNotInDirectory1);
```
tomrobertsOF: Do these need to be closed sequentially, or could you just pass both to one call of `quitApps`?

test/provider/AppDirectory-get.unittest.ts
```diff
@@ -0,0 +1,474 @@
+import 'jest';
```
tomrobertsOF: We have a convention in the integration tests of putting "categoryies" of tests under a subdirectory. I think it would make sense to do that here too, i.e. `/provider/AppDirectory/get.unittest.ts`.

bryangaleopenfin: I'd wanted with the unit tests to match the structure of the files under test (same as in Notifications) - I've moved the unit tests to actually do that.

Open to still grouping by folder, but would prefer to keep it that for unit tests, folders are there to match the structure we're mirroring

test/demo/appDirectory.inttest.ts
bryangaleopenfin: Can be one call. Done.

src/provider/model/Model.ts
bryangaleopenfin: Done, in part by changing `hasFinalAppInfo`

test/provider/AppDirectory-get.unittest.ts
tomrobertsOF: Ok, that's fine. Forgot that's how we were structuring for the unit tests.

tomrobertsOF (approved):

ngunday (approved):
LGTM after changes mentioned in #187

PR #186, opened 2019-12-04
[SERVICE-847] Added support for 'view' entities

pjbroadbent:
View-based applications, including applications using Layouts v2, can now connect to the service and make API calls.

### DIFF
--- a/src/provider/model/Environment.ts
+++ b/src/provider/model/Environment.ts
@@ -7,34 +7,61 @@ import {AppConnection} from './AppConnection';
 import {ContextChannel} from './ContextChannel';
 import {LiveApp} from './LiveApp';
 
+/**
+ * The runtime-defined list of possible OpenFin entity types.
+ */
 export enum EntityType {
-    WINDOW = 'window',
-    IFRAME = 'iframe',
+    /**
+     * Connection from a non-DOM source. Typically from a .NET, Java, or other adapter-based app.
+     */
     EXTERNAL_CONNECTION = 'external connection',
+    /**
+     * Connection from an app running in a `fin.View` container. Typically, a Layouts2-enabled app.
+     */
+    VIEW = 'view',
+    /**
+     * Connection from an app running in a `fin.Window` container. Typically a non-Layouts2 app, or the frame of a
+     * Layouts2-based application.
+     */
+    WINDOW = 'window',
+
+    /**
+     * Connection from an app running in an iframe within a `view` or `window` entity.
+     *
+     * Not supported. Entities of this type may not be able to use FDC3 fully.
+     */
+    IFRAME = 'iframe',
+    /**
+     * Runtime couldn't determine entity type. Could be an external connection from an outdated adapter version.
+     *
+     * Not supported.
+     */
     UNKNOWN = 'unknown'
 }
 
 export interface Environment {
     /**
-     * Indicates that a window has been created by the service.
+     * Indicates that a window-like entity has been created, and registered with the service. Fired only for entity types where
+     * this is possible (`VIEW` and `WINDOW`).
      *
-     * Will fire for all OpenFin windows that get created, and that existed before the service started running. These windows may or may not be FDC3-aware.
+     * Will fire for all window-like entities that get created, and that existed before the service started running.
+     * These entities may or may not be FDC3-aware.
      *
-     * Arguments: (identity: Identity)
+     * Arguments: (identity: Identity, entityType: EntityType)
      */
-    onWindowCreated: Signal<[Identity]>;
+    onWindowCreated: Signal<[Identity, EntityType]>;
 
     /**
-     * Indicates that a window has been closed.
+     * Indicates that an existing window-like entity has been closed.
      *
-     * Arguments: (identity: Identity)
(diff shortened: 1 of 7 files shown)


src/provider/model/FinEnvironment.ts
```diff
@@ -262,4 +265,16 @@ export class FinEnvironment extends AsyncInit implements Environment {
             }
         }
     }
+
+    /**
+     * Checks if an entity is of a "page-based" entity type. These are entities that are built using a
+     * webpage/browser/DOM/etc. This only includes entity types that both meet this definition, AND are supported by
+     * the service.
+     *
+     * These entitity types have a more finely controlled handshake process, as the fin API provides better eventing
```
bryangaleopenfin: "entitity

src/provider/model/FinEnvironment.ts
```diff
@@ -262,4 +265,16 @@ export class FinEnvironment extends AsyncInit implements Environment {
...
+     * and querying of these entity types.
+     */
+    private isPagedEntity(entityType: EntityType): boolean {
```
bryangaleopenfin: Make static or just pull out into a non-class function

bryangaleopenfin (commented):
Looks good, just need to stare at `expect.unittest.ts` a bit longer

test/provider/expect.unittest.ts
```diff
@@ -308,6 +308,16 @@ describe('When creating an external connection', () => {
 function expectTest(testWindow: TestWindow, appDirectoryResultTime: number, resultParams: ResultParam[]): void {
     const testParams = buildTestParams(testWindow, resultParams);
 
+    function getEntityType(entityIdentity: Identity, testWindowIdentity: Identity): EntityType {
```
bryangaleopenfin: This feels a bit round-a-bout

Would do:

- Put `const testWindowEntityType = testWindow.appType === 'non-directory-external' ?  EntityType.EXTERNAL_CONNECTION : return EntityType.WINDOW` around line 324.

- Use that const in the `emit`s

- Put the extracted function back in-line, and use that const

bryangaleopenfin (commented):
Any way to integration test this?

bryangaleopenfin:
Back to you @pjbroadbent

pjbroadbent:
> Any way to integration test this?

Requires creating windows with views, which would need some updates to the window creation and spawn utils. Worth considering once layouts 2 is more stable, but may just stick with some unit testing for the time being.

src/provider/model/FinEnvironment.ts
```diff
@@ -28,6 +28,18 @@ interface KnownEntity {
...
 
+/**
+ * Checks if an entity is of a "page-based" entity type. These are entities that are built using a
+ * webpage/browser/DOM/etc. This only includes entity types that both meet this definition, AND are supported by
+ * the service.
+ *
+ * These entity types have a more finely controlled handshake process, as the fin API provides better eventing
+ * and querying of these entity types.
+ */
+function isPagedEntity(entityType: EntityType): boolean {
```
bryangaleopenfin: Think this more naturally belongs at the bottom, being an implementation detail of the class

src/provider/model/Environment.ts
```diff
@@ -7,34 +7,61 @@ import {AppConnection} from './AppConnection';
...
 
     /**
-     * Indicates that a window has been closed.
+     * Indicates that an existing window-like entity has been closed.
      *
-     * Arguments: (identity: Identity)
+     * Arguments: (identity: Identity, entityType: EntityType)
      */
-    onWindowClosed: Signal<[Identity]>;
+    onWindowClosed: Signal<[Identity, EntityType]>;
```
ottah, line 59: Do we want to start referring to windows as entities now? Rename to onEntity*?

pjbroadbent: This only fires for window-like entities, so I think `onEntityClosed` would be misleading. It should really be something along the lines of `onWindowOrViewClosed`.

Discussed this briefly with Bryan, our view was that in the interests of conciseness, views should just be thought of as windows.

test/provider/expect.unittest.ts
```diff
@@ -374,10 +401,28 @@ function expectTest(testWindow: TestWindow, appDirectoryResultTime: number, resu
...
+                };
+                mockEnvironment.onWindowCreated.emit(windowIdentity, EntityType.WINDOW);
+            }
             mockEnvironment.onWindowCreated.emit(identity, testWindowEntityType);
         }, testWindow.createdTime);
         maybeSetTimeout(() => mockApiHandler.onConnection.emit(identity), testWindow.connectionTime);
-        maybeSetTimeout(() => mockEnvironment.onWindowClosed.emit(identity, testWindowEntityType), testWindow.closeTime);
+        maybeSetTimeout(() => {
+            if (testWindow.windowType === 'view') {
+                // There will also be an event for the window creation at around the same time as the view creation
```
bryangaleopenfin: Change comment for closed event

bryangaleopenfin (commented):
Looks good, just the comment comment

bryangaleopenfin (approved):

tomrobertsOF (approved):

PR #187, opened 2019-12-04
[SERVICE-822] Validate app-contributed directory data

bryangaleopenfin:


src/client/validation.ts
```diff
@@ -81,6 +82,28 @@ export function parseAppChannelName(name: string): ChannelId {
     return name;
 }
 
+/**
+ * Validates the provided number is an integer
+ */
+export function parseInteger(number: number): number {
```
ngunday: I'd prefer to use something like `value` to resolve the ambiguity. can't tell if you mean the type name or argument name when you use `number` in the body.

src/client/validation.ts
```diff
@@ -81,6 +82,28 @@ export function parseAppChannelName(name: string): ChannelId {
...
+        throw new TypeError(`${safeStringify(number, 'The provided value')} is not a valid integer`);
+    }
+
+    return number;
+}
+
+/**
+ * Validates the provided app directory data - can be either a URL or an array of Applications
+ */
+export function parseAppDirectoryData(data: string | Application[]): string | Application[] {
```
ngunday: I'd prefix these functions with `is` or `validate` instead of `parse`.

src/provider/utils/helpers.ts
```diff
@@ -52,3 +52,13 @@ export async function collateClientCalls<T = void>(promises: Promise<T>[]): Prom
         return [ClientCallsResult.TIMEOUT, undefined];
     }
 }
+
+/**
+ * Takes an array and a comparison fuction, and returns a deduplicated equivalent, taking the first of any duplicate elements
```
ngunday: typo: fuction => function

bryangaleopenfin: ✅

src/client/validation.ts
bryangaleopenfin: ✅

src/client/validation.ts
bryangaleopenfin: Yeah, parse isn't the right word, but wanted to communicate that these actually return something.

Have gone withe `sanitize`, which I think better capture "unsafe data in, safe data out"

ngunday (approved):

src/provider/model/AppDirectory.ts
```diff
@@ -20,6 +21,22 @@ interface CacheEntry {
...
+
+interface DomainShardScope {
+    type: 'domain';
+    domain: string;
+}
+
+type ShardScope = GlobalShardScope | DomainShardScope;
+
+interface ScopedAppDirectoryShard {
+    scope: ShardScope;
```
pjbroadbent: Is the naming of these intended to align with the `service-config` definition of scopes? As it's the same name and use-case, but using a different syntax (using `type` instead of `level`). Also, defining a new `"global"` scope for something that is `"desktop"` scope in the `service-config` world.

src/provider/model/AppDirectory.ts
```diff
@@ -117,16 +134,27 @@ export class AppDirectory extends AsyncInit {
...
             {
-                urls: configUrl ? [configUrl] : [],
-                applications: []
+                scope: {
+                    type: 'global'
+                },
+                shard: {
+                    urls: configUrl ? [configUrl] : [],
+                    applications: []
+                }
```
pjbroadbent: If there's no `configUrl`, should this shard just be omitted? Rather than creating an empty shard.

src/provider/model/AppDirectoryStorage.ts
```diff
@@ -2,8 +2,13 @@ import {Signal} from 'openfin-service-signal';
...
+export interface DomainAppDirectoryShard {
+    domain: string;
+    shard: StoredAppDirectoryShard;
+}
+
 export interface AppDirectoryStorage {
     readonly changed: Signal<[]>;
 
-    getStoredDirectoryShards(): StoredAppDirectoryShard[];
+    getDirectoryShards(): DomainAppDirectoryShard[];
```
pjbroadbent: Can this return `DomainAppDirectoryShard[]`? Otherwise there's a lot of mapping going on here?:

1. Client stores `{version, urls, applications}` objects in the store
2. FinAppDirectoryStorage transforms these into `{domain: string, shard: {version, urls, applications}}`
3. AppDirectory transforms these into `{scope: DomainShardScope, shard: {version, urls, applications}}`

pjbroadbent: _Might_ also make sense for `FinAppDirectoryStorage` to query and append the desktop-level shard?..

Then the `AppDirectoryStorage` is what collects all directory configuration, and `AppDirectory` is what handles all data fetching and caching? Seems like a cleaner separation?

src/provider/model/AppDirectory.ts
```diff
@@ -231,3 +263,50 @@ export class AppDirectory extends AsyncInit {
...
+
+            return false;
+        }
+    });
+}
+
+function isUrlValidForScope(scope: ShardScope, url: string): boolean {
+    if (scope.type === 'global') {
+        return true;
+    } else if (scope.type === 'domain') {
```
pjbroadbent, line 260: Rather than filtering needing to be aware of both shard types, would it make sense to do the filtering before combining all the shards together? Since we're only filtering domain-based shards, and those come from a separate source to the global shard?

src/provider/model/AppDirectory.ts
```diff
@@ -117,16 +134,27 @@ export class AppDirectory extends AsyncInit {
...
+                    domain: shard.domain
+                } as ShardScope,
+                shard: shard.shard
+            }))
         ];
 
-        const remoteDirectorySnippets = await parallelMap(directoryShards, async (shard) => {
-            return parallelMap(shard.urls, async (url) => {
+        const remoteDirectorySnippets = await parallelMap(scopedShards, async (scopedShard) => {
+            return parallelMap(filterUrlsByScope(scopedShard.scope, scopedShard.shard.urls), async (url) => {
```
pjbroadbent: Having some trouble understanding the layers here. `remoteDirectorySnippets` has type `Application[][][]`, and not clear what each level of the array relates to. Some intermediate type definitions might make this more understandable.

Is using `parallelForEach` to push into an array simpler than the mapping and reducing?
```
const applications: Application[] = [];
await parallelForEach(scopedShards, async (scopedShard) => {
    const {scope, shard} = scopedShard;

    applications.push(...filterAppsByScope(scope, shard.applications));
    await parallelForEach(filterUrlsByScope(scope, shard.urls), async (url) => {
        // TODO: URLs will be fetched once per service run. Improve this logic [SERVICE-841]
        let snippet: Application[] | null = this.fetchCachedSnippet(url);

        if (!snippet) {
            snippet = await this.fetchRemoteSnippet(url);

            if (snippet) {
                this._fetchedUrls.add(url);
                this.updateCache(url, snippet);
            }
        }

        applications.push(...(snippet || []));
    });
});
```
(Above assumes the `this._fetchedUrls.has` check moves into `fetchCachedSnippet`)

test/provider/model/AppDirectory-get.unittest.ts
```diff
@@ -292,55 +330,126 @@ describe('When fetching initial data', () => {
...
+            setupDirectoryStorage([
+                {domain: insideDomain, shard: {urls: [], applications: [app1, app2, app3]}}
+            ]);
+
+            await createAppDirectory();
+
+            await expect(appDirectory.getAllApps()).resolves.toEqual([app1, app3]);
+        });
+
+        test.only('When a stored shard references a remote snippet from outside its domain, that snippet is not used by the directory', async () => {
```
pjbroadbent: Remove `.only`s (here and below)

bryangaleopenfin: ✅

src/provider/model/AppDirectoryStorage.ts
bryangaleopenfin: First comment, do you mean "Can this return `ScopedAppDirectoryShard[]`"?

bryangaleopenfin: Moved into `FinAppDirectoryStorage `

src/provider/model/AppDirectory.ts
bryangaleopenfin: Done, in `FinAppDirectoryStorage`

src/provider/model/AppDirectory.ts
bryangaleopenfin: In new version they're already combined when they come out of the AppDirectoryStore, and even if that wasn't the case I'd end up repeating myself, performing the same step twice, first without the filter then without

src/provider/model/AppDirectory.ts
bryangaleopenfin: Not intended at all, and not sure I agree it's the same use case. Prefer if I use different naming entirely?

src/provider/model/AppDirectory.ts
bryangaleopenfin: Doing like this the final applications order becomes non-deterministic, and moving the `this._fetchedUrls.has` check into `fetchCachedSnippet` changes the behaviour in the case that we tried to fetch the URL and it failed

bryangaleopenfin: I tried creating some intermediate types and using those in the arrow function declarations but it kinda felt like noise, so I've just changed the naming of `remoteDirectorySnippets` to hopefully make the naming clearer

bryangaleopenfin:
@pjbroadbent Back to you

src/provider/model/AppDirectoryStorage.ts
pjbroadbent: > First comment, do you mean "Can this return `ScopedAppDirectoryShard[]`"?

Oops, yeah - that's what I meant :-)

src/provider/model/AppDirectory.ts
pjbroadbent: Should validation be inside `AppDirectoryStore` then?.. To remove invalid data as early as possible, rather than (potentially, if other parts of the service needed to query the directory store) needing to duplicate the validation elsewhere?

bryangaleopenfin: The point of AppDirectoryStore was to be a light wrapper around the use of `fin`, so we could avoid referencing fin in tests. Moving the config store stuff into AppDirectoryStore I can juuuust about be ok with, but this feels like to much. I also would argue that anywhere using AppDirectoryStore directly is doing something wrong.

This could instead be 3 layers, the fin wrapper layer, the filter and combine with config store layer, and the actual AppDirectory layer, but that feels like overkill

pjbroadbent: Agree three layers is a bit much, makes sense keeping `ADS` as more of a fin-abstraction.

src/provider/model/AppDirectory.ts
pjbroadbent: Meh - as discussed, they're sufficiently distinct in usage.

pjbroadbent (approved):

PR #188, opened 2019-12-05
[N/A] Prevent context listener breaking change

pjbroadbent:
Provider now tracks client version of each connection. Will assume that any pre-0.2.1 client connections always have context listeners - this prevents the recent addition of context listener handshakes from being a breaking change.

### DIFF
--- a/src/provider/APIHandler.ts
+++ b/src/provider/APIHandler.ts
@@ -139,12 +139,12 @@ export class APIHandler<T extends Enum> {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     private onConnectionHandler(identity: Identity, payload?: any): void {
         const id = getId(identity);
-        const semver: SemVer = SemVer.parse(payload && payload.version);
+        const semVer: SemVer = SemVer.parse(payload && payload.version);
 
-        this._clientVersions[id] = semver;
+        this._clientVersions[id] = semVer;
 
-        if (semver.isValid) {
-            console.log(`connection from client: ${identity.name}, version: ${semver.version}`);
+        if (semVer.valid) {
+            console.log(`connection from client: ${identity.name}, version: ${semVer.version}`);
         } else {
             console.log(`connection from client: ${identity.name}, unable to determine version`);
         }
--- a/src/provider/APIHandler.ts
+++ b/src/provider/APIHandler.ts
@@ -6,6 +6,9 @@ import {Signal} from 'openfin-service-signal';
 
 import {SERVICE_CHANNEL, serializeError} from '../client/internal';
 
+import {getId} from './utils/getId';
+import {SemVer} from './utils/SemVer';
+
 /**
  * Semantic type definition.
  *
@@ -66,6 +69,11 @@ export class APIHandler<T extends Enum> {
     public readonly onDisconnection: Signal<[Identity]> = new Signal();
 
     private _providerChannel!: ChannelProvider;
+    private readonly _clientVersions: {[key: string]: SemVer};
+
+    constructor() {
+        this._clientVersions = {};
+    }
 
     public isClientConnection(identity: Identity): boolean {
         return !!this._providerChannel && this._providerChannel.connections.some((conn: Identity) => {
@@ -75,6 +83,10 @@ export class APIHandler<T extends Enum> {
 
     public getClientConnections(): Identity[] {
         return this._providerChannel.connections;
+    }
+
+    public getClientVersion(id: string): SemVer {
+        return this._clientVersions[id] || SemVer.parse('');
     }
 
     public dispatch(to: Identity, action: string, payload: any): Promise<any> {
@@ -126,8 +138,13 @@ export class APIHandler<T extends Enum> {
     // TODO?: Remove the need for this any by defining connection payload type?
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     private onConnectionHandler(identity: Identity, payload?: any): void {
-        if (payload && payload.version && payload.version.length > 0) {
-            console.log(`connection from client: ${identity.name}, version: ${payload.version}`);
+        const id = getId(identity);
+        const semver: SemVer = SemVer.parse(payload && payload.version);
+
+        this._clientVersions[id] = semver;
+
+        if (semver.isValid) {
+            console.log(`connection from client: ${identity.name}, version: ${semver.version}`);
         } else {
             console.log(`connection from client: ${identity.name}, unable to determine version`);
         }
@@ -139,7 +156,10 @@ export class APIHandler<T extends Enum> {
         });
     }
 
-    private onDisconnectionHandler(app: Identity): void {
-        this.onDisconnection.emit(app);
+    private onDisconnectionHandler(identity: Identity): void {
+        const id = getId(identity);
+        delete this._clientVersions[id];
+
+        this.onDisconnection.emit(identity);
     }
 }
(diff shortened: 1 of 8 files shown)


src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+    private static readonly REGEX_SEMVER: RegExp = /(\d+)\.(\d+)\.(\d+)(?:-([^-+]+))?/;
+    private static readonly REGEX_INTEGER: RegExp = /^\d+$/;
+    private static readonly CACHE: {[key: string]: SemVer} = {};
+
+    public static compare(a: string | SemVer, operator: Operator, b: string | SemVer): boolean {
+        return SemVer.parse(a).compare(operator, b);
+    }
+
+    public static parse(version: string | SemVer): SemVer {
+        if (typeof version === 'string' || version === null || version === undefined) {
```
bryangaleopenfin: Not seeing that this can do anything meaningful with `null` or `undefined`, and if we're expecting those, can they be in the type?

src/provider/APIHandler.ts
```diff
@@ -126,8 +138,13 @@ export class APIHandler<T extends Enum> {
     // TODO?: Remove the need for this any by defining connection payload type?
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     private onConnectionHandler(identity: Identity, payload?: any): void {
-        if (payload && payload.version && payload.version.length > 0) {
-            console.log(`connection from client: ${identity.name}, version: ${payload.version}`);
+        const id = getId(identity);
+        const semver: SemVer = SemVer.parse(payload && payload.version);
```
bryangaleopenfin: If the class is `SemVer`, should the variable be `semVer`?

src/provider/model/FinEnvironment.ts
```diff
@@ -116,19 +122,21 @@ export class FinEnvironment extends AsyncInit implements Environment {
 
             const application = fin.Application.wrapSync(identity);
             const applicationInfo = await application.getInfo();
+            let {name: title, icon} = applicationInfo.initialOptions as ApplicationOption;
 
-            const {shortcut, startup_app: startupApp} = applicationInfo.manifest as OFManifest;
-
-            const title = (shortcut && shortcut.name) || startupApp.name || startupApp.uuid;
-            const icon = (shortcut && shortcut.icon) || startupApp.icon;
+            if (applicationInfo.manifest) {
```
bryangaleopenfin, line 128: If the type is wrong and `manifest` can be undefined/null, please comment

src/provider/model/FinEnvironment.ts
```diff
@@ -116,19 +122,21 @@ export class FinEnvironment extends AsyncInit implements Environment {
...
             return {
-                appId: application.identity.uuid,
-                name: application.identity.uuid,
+                appId: identity.uuid,
+                name: identity.uuid,
                 title,
                 icons: icon ? [{icon}] : undefined,
                 manifestType: 'openfin',
-                manifest: applicationInfo.manifestUrl
+                manifest: applicationInfo.manifestUrl || ''
```
bryangaleopenfin, line 141: Ditto

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+            return semver;
+        } else {
+            return version;
+        }
+    }
+
+    public readonly version: string;
+    public readonly type: SemVerType;
+    public readonly components: Readonly<[number, number, number, ...(string|number)[]]>;
```
bryangaleopenfin: Space around `|`

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+            this.components = [parseInt(major), parseInt(minor), parseInt(patch), ...preReleaseComponents];
+            this.type = SemVerType.VALID;
+        } else {
+            this.components = [0, 0, 0];
+            this.type = version ? SemVerType.INVALID : SemVerType.UNKNOWN;
+            this.version = this.type;
+        }
+    }
+
+    public get isValid(): boolean {
```
bryangaleopenfin: I think of "is" as like "get but for a boolean", so it feels weird this is a getter rather than just a method.

Would make this `public get valid()` or `public isValid()`

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+        return this.type === SemVerType.VALID;
+    }
+
+    public compare(operator: Operator, other: string|SemVer, defaultIfInvalid: boolean = false): boolean {
```
bryangaleopenfin: Space around `|`

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Util for comparing semantic version numbers.
+ *
+ * NOTE: Only parses the major, minor, patch and pre-release parts of the semver string - build metadata is ignored.
+ *
+ * Spec: https://semver.org/
+ */
+export class SemVer {
```
bryangaleopenfin, line 21: This is very unit testable

bryangaleopenfin (commented):
This seems like it *should* be integration testable? Like, easy enough to have a test html that imports a pre-0.2.1 client using a script tag?

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+    private static readonly REGEX_SEMVER: RegExp = /(\d+)\.(\d+)\.(\d+)(?:-([^-+]+))?/;
+    private static readonly REGEX_INTEGER: RegExp = /^\d+$/;
+    private static readonly CACHE: {[key: string]: SemVer} = {};
+
+    public static compare(a: string | SemVer, operator: Operator, b: string | SemVer): boolean {
```
tim-dinsdale, line 26: why the union type? Don't we just want SemVers, and not strings?

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,114 @@
...
+        return SemVer.parse(a).compare(operator, b);
+    }
+
+    public static parse(version: string | SemVer): SemVer {
+        if (typeof version === 'string' || version === null || version === undefined) {
+            let semver = SemVer.CACHE[version];
```
tim-dinsdale: This cache looks like a premature optimisation - do we really think that it's worth it?

src/provider/utils/SemVer.ts
pjbroadbent: The `null`/`undefined` checks were sanity-checks for clients passing in weird/missing version numbers at runtime (for instance, the adapters don't currently pass a version number in the connection payload). You _shouldn't_ call this with null/undefined, but it _could_ happen at runtime.

Flipped the check to still handle these cases, but in a less misleading way.

src/provider/utils/SemVer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    public static compare(a: string | SemVer, operator: Operator, b: string | SemVer): boolean {
+        return SemVer.parse(a).compare(operator, b);
+    }
+
+    public static parse(version: string | SemVer): SemVer {
+        if (version instanceof SemVer) {
+            return version;
+        } else {
+            let semver = SemVer.CACHE[version];
```
bryangaleopenfin: Same comment as before about `semVer` vs. `semver`

bryangaleopenfin (commented):
Worth testing semver comparisons?

(I feel bad leaning on you and would probably accept 'no' as an answer)

src/provider/utils/SemVer.ts
pjbroadbent: As an "ease of use" thing, will implicitly create a `SemVer` object for any incoming strings. This also couples with the caching, which is less effective if done outside of the util.

src/provider/utils/SemVer.ts
pjbroadbent: Could be... Was motivated by the size of the regex, and how we're basically only ever going to need to check a rather small set of semver strings over and over again.

bryangaleopenfin (approved):

tim-dinsdale (approved):

PR #189, opened 2019-12-09
Feature/fdc3 asar

tim-dinsdale:
Re-merge of [Wenjun's branch](https://github.com/HadoukenIO/fdc3-service/pull/182/) with fixes applied (and some random spelling fixes while I was in there)


src/provider/controller/ContextHandler.ts
```diff
@@ -20,7 +20,7 @@ export class ContextHandler {
     private readonly _model: Model;
 
     constructor(
-        @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>,
+        @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>, // eslint-disable-line @typescript-eslint/indent
```
bryangaleopenfin, line 23: Huh. Is this linting rule changes with the updated service-tooling? Why was the eslint exception not needed before?

src/provider/model/FinEnvironment.ts
```diff
@@ -111,16 +111,16 @@ export class FinEnvironment extends AsyncInit implements Environment {
...
                 // eslint-disable-next-line camelcase
-                startup_app: {uuid: string; name?: string; icon?: string};
+                startup_app?: {uuid: string; name?: string; icon?: string};
             }
 
             const application = fin.Application.wrapSync(identity);
             const applicationInfo = await application.getInfo();
 
-            const {shortcut, startup_app: startupApp} = applicationInfo.manifest as OFManifest;
+            const {shortcut, startup_app: startupApp} = (applicationInfo.manifest || {}) as OFManifest;
```
bryangaleopenfin, line 120: Would still comment that we don't trust the `ApplicationInfo` interface that `manifest` will be non-null. Seems like the kind of thing where someone would otherwise look at the type and decide `||` check isn't necessary

bryangaleopenfin (commented):
Looks good, just want an extra comment, and some leaning on to convince me that we're doing the right thing wrt the eslint-disable stuff :)

bryangaleopenfin:
@tim-dinsdale Flagging that it looks like @wenjunche has made roughly the same changes in his PR. Though to add to the confusion, I missed this and resolved the comments on that PR as though they were on this branch

PR #195, opened 2019-12-16
[SERVICE-823] Integrate "meta directory" with the rest of the service

bryangaleopenfin:
This PR changes how we access the App Directory, so we now almost always go through Model, and do so using higher level functions than before. This is so we can resolve API calls without needing to make multiple queries to the directory, reducing the scope for things to go wonky should the app directory change at just the wrong time.

I claim once this is in, the feature branch is almost ready to merge into develop, and the only work remaining is blocked on an actual release of a runtime with the Storage API


src/provider/model/Model.ts
```diff
@@ -407,6 +411,61 @@ export class Model {
...
+     *
+     * @param name The name of the application to use in any error messages
+     * @param getAppInfo A function that returns an [[Application]], if a matching application cannot be found among running apps
+     * @param identifyAppInfo A function to test if the passed [[Application]] is the one we want to find. Should returns true if a match, false otherwise
+     * @param testAppInfo A function to test if the passed [[Application]] fulfils our criteria. Should return undefined if criteria is satisfied, otherwise
+     * T describing the failure
+     * @param testLiveApp A function to test if the passed [[LiveApp]] fulfils our criteria. should return undefined if criteria is satisfied, otherise T
+     * describing the failure
+     * @typeparam T Failure type, returned if one of ours tests for fails for the identified [[Application]]/[[LiveApp]]
+     */
```
tomrobertsOF: As discussed, think this could use a bit more high-level explanation of why these are functions instead of just passing the values

bryangaleopenfin: Expanded comments, should explain things better

tomrobertsOF (approved):
Much clearer now with the comments. 👍

src/provider/model/Model.ts
```diff
@@ -407,6 +420,77 @@ export class Model {
...
+     * true if this is the app we are looking for, false otherwise
+     *
+     * @param getAppInfo Will be called if we're not able to find the desired application among currently running
+     * applications. Should return the [[Application]] of an app we want to start
+     *
+     * @param testLiveApp In the case where the application we are looking for is already running (i.e.
+     * `identifyAppInfo` has found a match), this will be called to test if the [[LiveApp]] fulfils our desired
+     * criteria. This should return undefined if our criteria is fulfiled, otherwise T describing the failure. In the
+     * case where this is called, `getOrCreateLiveApp` will return either the [[LiveApp]] or the failure value
+     *
```
tim-dinsdale: returning 'undefined' on success is kind of weird - I would expect 'true'. But I mean, it works, and it's not like it's a public api

src/provider/model/Model.ts
```diff
@@ -407,6 +420,77 @@ export class Model {
...
+     * @param testAppInfo In the case where the application we are looking for is not already running (i.e.
+     * `identifyAppInfo` has not found a match and we have an [[Application]] from `getAppInfo`), this will be called
+     * to test if the [[Application]] we want to start fulfils our desired criteria. This should return undefined if
+     * our criteria is fulfiled, otherwise T describing the failure. In the case where this is called,
+     * `getOrCreateLiveApp` will either start the [[Application]] and return the corresponding [[LiveApp]], or return
+     * the failure value
+     *
+     * @typeparam T Failure type, returned if one of ours tests for fails for the identified [[Application]]/[[LiveApp]]
+     */
```
tim-dinsdale, line 455: I think this is correct but I am surprised at the complexity of the api - essentially we have two callbacks that are then fed into two other callbacks, where those two other callbacks are only used once. I would perhaps naively imagined this could be simplified into a function that simply accepts two callbacks, and in the getOrCreateLiveAppByNameForIntent case, roll testLiveApp into identifyAppInfo and testAppInfo into getAppInfo. 

I suspect you have good reasons for this, though, so maybe I am wrong

src/provider/model/Model.ts
```diff
@@ -166,33 +165,63 @@ export class Model {
...
-                if (identity.uuid === uuid) {
-                    slot.remove();
-                    deferredPromise.resolve(liveApp);
+                if (!liveApp.mature && AppDirectory.shouldAppSupportIntent(liveApp.appInfo!, intentType, contextType)) {
+                    return undefined;
                 }
-            });
 
-            this._environment.createApplication(appInfo);
+                return 'does-not-support-intent';
```
tim-dinsdale: Does this want to be a const?

src/provider/model/Model.ts
bryangaleopenfin: I'm sorta thinking of these functions as "tell me why this *doesn't* match our criteria", so `undefined` for success makes sense. Feels like it would be easy to get something wrong if both the success and failure cases are truthy

bryangaleopenfin:
Can get it down to two callbacks if I push some complexity into their return types. Checking tests still pass then will update comments

src/provider/model/Model.ts
bryangaleopenfin: My feeling is no. the type system enforces that I can't type some arbitrary string here, and with stuff like `Promise<LiveApp | 'does-not-support-intent'>`, you have to use a string literal anyway, you can't extract that into a const

bryangaleopenfin:
@tim-dinsdale Ready for another look

tim-dinsdale (approved):

PR #196, opened 2019-12-19
Dev/tim/noodles

tim-dinsdale:
noodly noodly


src/client/connection.ts
```diff
@@ -52,19 +53,29 @@ export function getServicePromise(): Promise<ChannelClient> {
...
             channelPromise = Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.'));
-        } else if (fin.Window.me.uuid === SERVICE_IDENTITY.uuid && fin.Window.me.name === SERVICE_IDENTITY.name) {
-            // Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
-            // that includes this, but for now it is easier to put a guard in place.
-            channelPromise = Promise.reject<ChannelClient>(new Error('Trying to connect to provider from provider'));
         } else {
-            channelPromise = fin.InterApplicationBus.Channel.connect(SERVICE_CHANNEL, {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
-                // Register service listeners
-                channel.register('WARN', (payload: unknown) => console.warn(payload));  // tslint:disable-line:no-any
+            channelPromise = new Promise<ChannelClient>((resolve, reject) => {
```
tomrobertsOF, line 57: Not sure what we gain from wrapping this in another promise, since the `.then` will return a promise anyway. You can just make the inner function async and use return/throw instead of resolve/reject.

I know this is more about content than style at the moment, but figure it's worth mentioning.

src/client/connection.ts
```diff
@@ -52,19 +53,29 @@ export function getServicePromise(): Promise<ChannelClient> {
...
+                fin.System.getRuntimeInfo().then((info: RuntimeInfo & {fdc3AppUuid?: string; fdc3ChannelName?: string}) => {
+                    if (info.fdc3AppUuid && info.fdc3ChannelName) {
```
tomrobertsOF, line 59: Does this actually work right now, or are you adding these in anticipation of the runtime doing it in a later build?

src/client/internal.ts
```diff
@@ -15,18 +15,24 @@ import {AppIntent, Context, IntentResolution, Listener} from './main';
 import {ChannelId, DefaultChannel, SystemChannel, DisplayMetadata, ChannelWindowAddedEvent, ChannelWindowRemovedEvent, ChannelChangedEvent, ChannelBase, AppChannel} from './contextChannels';
 import {FDC3Error} from './errors';
 
+export interface ServiceIdentity{
```
tomrobertsOF, line 18: Why do we need another identity type just for the service? UUID/Name pairs are used all over the place so we should keep it consistent.

src/client/internal.ts
```diff
@@ -326,3 +332,28 @@ export function deserializeError(error: Error): Error | FDC3Error {
...
+    return serviceIdentity;
+}
+
+export function getServiceIdentityUUID(): string {
+    return serviceIdentity.uuid;
+}
+
+export function getServiceIdentityName(): string {
+    return serviceIdentity.name;
+}
```
tomrobertsOF, line 354: These don't really seem necessary to me. How is `getServiceIdentityUUID()` any better than `getServiceIdentity().uuid`?

src/provider/controller/ChannelHandler.ts
```diff
@@ -18,9 +20,14 @@ export class ChannelHandler {
     public readonly onChannelChanged: Signal<[AppConnection, ContextChannel | null, ContextChannel | null], Promise<void>>;
 
     private readonly _model: Model;
+    private readonly _mrh: MultiRuntimeHandler;
```
tomrobertsOF, line 23: This isn't rust; you can afford to write out an actual variable name.

src/provider/controller/ContextHandler.ts
```diff
@@ -18,15 +19,18 @@ export class ContextHandler {
     private readonly _apiHandler: APIHandler<APIFromClientTopic>;
     private readonly _channelHandler: ChannelHandler;
     private readonly _model: Model;
+    private readonly _mrh: MultiRuntimeHandler;
 
     constructor(
-        @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>,
+        @inject(Inject.API_HANDLER) apiHandler: APIHandler<APIFromClientTopic>, // eslint-disable-line @typescript-eslint/indent
```
tomrobertsOF, line 25: Why did we not need this eslint comment before, but we do now?

src/provider/controller/ContextHandler.ts
```diff
@@ -74,15 +78,23 @@ export class ContextHandler {
      * @param context Context to send
      * @param source Entity sending the context. It won't receive the broadcast
      * @param channel ContextChannel to broadcast on
+     * @param foreign whether the message came from a foreign runtime or not
      */
-    public broadcastOnChannel(context: Context, source: AppConnection, channel: ContextChannel): Promise<void> {
+    public broadcastOnChannel(context: Context, source: AppConnection | null, channel: ContextChannel, foreign: boolean): Promise<void> {
```
tomrobertsOF, line 83: Is the point here that `foreign: false` always corresponds to `source: null`? If so, it should be possible to type this more strongly to only allow those options, but not super vital right now.

src/provider/controller/EventHandler.ts
```diff
@@ -17,7 +17,7 @@ export class EventHandler {
     private readonly _apiHandler: APIHandler<APIFromClientTopic>;
 
     constructor(
-        @inject(Inject.CHANNEL_HANDLER) channelHandler: ChannelHandler,
+        @inject(Inject.CHANNEL_HANDLER) channelHandler: ChannelHandler, // eslint-disable-line @typescript-eslint/indent
```
tomrobertsOF, line 20: As above: why did we not need this eslint comment before, but we do now?

src/provider/controller/MultiRuntimeHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+import {getServiceIdentity} from '../../client/internal';
+import {Model} from '../model/Model';
+import {AppConnection} from '../model/AppConnection';
+
+import {AsyncInit} from './AsyncInit';
+import {ContextHandler} from './ContextHandler';
+import {ChannelHandler} from './ChannelHandler';
+
+const IAB_TOPIC = 'fdc3-cabal';
+const REQUEST_GET_ALL = 'request-get-all';
```
tomrobertsOF, line 16: We've generally been using enums for these kinds of const string lists. Might be worth using one here too if possible

src/provider/controller/MultiRuntimeHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+    ) {
+        super();
+        this._contextHandler = contextHandler;
+        this._channelHandler = channelHandler;
+        this._model = model;
+
+        this._channelHandler.onChannelChanged.add(this.onChannelChangedHandler, this);
+    }
+
+    private async onChannelChangedHandler(connection: AppConnection, channel: ContextChannel | null, previousChannel: ContextChannel | null): Promise<void> {
```
tomrobertsOF, line 74: `on[...]Handler` seems a bit redundant. Would make it `channelChangeHandler`, `onChannelChanged`, or better yet `handleChannelChanged` to match the other methods in this class.

src/provider/controller/MultiRuntimeHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    private setSetColorChannel(channelState: ChannelState) {
+        if (channelState.context) {
+            const channel = this._channelHandler.getChannelById(channelState.channelId);
+            this._contextHandler.broadcastOnChannel(channelState.context, null, channel, true);
+        }
+    }
+
+    protected async init(): Promise<void> {
+        const subMsg = (msg: Msg, sourceUUID: string, sourceName: string) => {
```
tomrobertsOF, line 116: Any particular reason to have this assigned? Could it not just be passed directly into the subscribe call?

src/provider/common/Injectables.ts
```diff
@@ -14,6 +14,7 @@ enum Injectable {
     EVENT_HANDLER,
     INTENT_HANDLER,
     MODEL,
+    MULTI_RUNTIME_HANDLER,
```
tomrobertsOF, line 17: You've added the injectable key, but also need to add a type and a binding for it in `injector.ts`

src/provider/controller/MultiRuntimeHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+interface MsgUnregisterAppFromChannel {
+    kind: typeof UNREGISTER_APP_FROM_CHANNEL;
+    channelId: ChannelId;
+    identity: Identity;
+}
+
+type Msg = MsgRequestGetAll | MsgRespondGetAll | MsgSetColorChannel | MsgRegisterAppInChannel | MsgUnregisterAppFromChannel;
+
+@injectable()
+export class MultiRuntimeHandler extends AsyncInit {
```
tomrobertsOF, line 57: I know we're just in POC for now, but when we do it for real this should not have any direct fin dependencies, and should instead use the injected `Environment` to do what it needs to do.

src/provider/model/FinEnvironment.ts
```diff
@@ -117,13 +117,13 @@ export class FinEnvironment extends AsyncInit implements Environment {
             interface OFManifest {
                 shortcut?: {name?: string; icon: string};
                 // eslint-disable-next-line camelcase
-                startup_app: {uuid: string; name?: string; icon?: string};
+                startup_app?: {uuid: string; name?: string; icon?: string};
```
tomrobertsOF, line 120: Is this really optional? Why did this need changing?

package.json
```diff
@@ -52,7 +53,7 @@
     "openfin-service-async": "^1.0.1",
     "openfin-service-config": "1.0.2",
     "openfin-service-signal": "^1.0.0",
-    "openfin-service-tooling": "1.0.21",
+    "openfin-service-tooling": "1.0.23",
```
rypete, line 56: The latest `openfin-service-tooling` is version `1.0.25`. Can we bring this in as part of this PR (or separate)?

src/client/connection.ts
```diff
@@ -52,19 +53,29 @@ export function getServicePromise(): Promise<ChannelClient> {
...
+                        setServiceChannel(info.fdc3ChannelName);
+                    }
+                    if (fin.Window.me.uuid === getServiceIdentityUUID() && fin.Window.me.name === getServiceIdentityName()) {
+                        // Currently a runtime bug when provider connects to itself. Ideally the provider would never import a file
+                        // that includes this, but for now it is easier to put a guard in place.
+                        reject(new Error('Trying to connect to provider from provider'));
+                    } else {
+                        fin.InterApplicationBus.Channel.connect(getServiceChannel(), {payload: {version: PACKAGE_VERSION}}).then((channel: ChannelClient) => {
+                            // Register service listeners
+                            channel.register('WARN', (payload: unknown) => console.warn(payload));  // tslint:disable-line:no-any
```
rypete, line 70: Still need the `// tslint:disable-line:no-any`?

src/client/internal.ts
rypete: If you do keep it for any reason, brace formatting

src/provider/controller/MultiRuntimeHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+const SET_COLOR_CHANNEL = 'set-color-channel';
+const REGISTER_APP_IN_CHANNEL = 'register-app-in-channel';
+const UNREGISTER_APP_FROM_CHANNEL = 'unregister-app-from-channel';
+
+interface MsgRequestGetAll {
+    kind: typeof REQUEST_GET_ALL;
+    source: Identity;
+}
+
+interface ChannelState{
```
rypete, line 27: Brace formatting

src/provider/controller/ResolverHandler.ts
```diff
@@ -18,7 +18,7 @@ const RESOLVER_URL = (() => {
     }
 
     // Locate the default resolver HTML page, relative to the location of the provider
-    return providerLocation.replace('provider.html', 'ui/resolver');
+    return providerLocation.replace('provider.html', 'ui/resolver/index.html');
```
rypete, line 21: Maybe a silly question, but do we not have `index.*` be the default document for the server?

rypete (commented):
@tomrobertsOF marked a few but, what overall what is the reason for the multiple ` // eslint-disable-line @typescript-eslint/indent`? There are a few formatting errors. I'm wondering if a lint will fix both? Ignore me if linting has changed :)

Can we also have a proper title for this PR? It will look nicer in the history when merged

PR #197, opened 2020-01-06
[SERVICE-794][SERVICE-793] Reconnect to service on disconnect

ottah:
This PR adds reconnect logic to the client so for when the provider reloads or disconnects.

- Await for DOMContentLoaded event before initializing.
- Wait for IAB channel.
- On disconnect rejoin channel.
- Readd context/intent listeners on reconnect.
- Rejoin channels.

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -49,24 +49,20 @@ const hasDOMContentLoaded = new DeferredPromise<void>();
 
 if (typeof document !== 'undefined') {
     document.addEventListener('DOMContentLoaded', () => {
-        initialize();
+        hasDOMContentLoaded.resolve();
+        if (typeof fin !== 'undefined') {
+            getServicePromise();
+
+            fin.InterApplicationBus.Channel.onChannelDisconnect((event: OpenFinChannelConnectionEvent) => {
+                const {uuid, name, channelName} = event;
+                if (uuid === getServiceIdentity().uuid && name === getServiceIdentity().name && channelName === getServiceChannel()) {
+                    channelPromise = null;
+                }
+            });
+        } else {
+            channelPromise = Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.'));
+        }
     });
-}
-
-function initialize() {
-    hasDOMContentLoaded.resolve();
-    if (typeof fin !== 'undefined') {
-        getServicePromise();
-
-        fin.InterApplicationBus.Channel.onChannelDisconnect((event: OpenFinChannelConnectionEvent) => {
-            const {uuid, name, channelName} = event;
-            if (uuid === getServiceIdentity().uuid && name === getServiceIdentity().name && channelName === getServiceChannel()) {
-                channelPromise = null;
-            }
-        });
-    } else {
-        channelPromise = Promise.reject(new Error('fin is not defined. The openfin-fdc3 module is only intended for use in an OpenFin application.'));
-    }
 }
 
 export async function getServicePromise(): Promise<ChannelClient> {
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -17,7 +17,7 @@ import {DeferredPromise} from 'openfin-service-async';
 import {ChannelClient} from 'openfin/_v2/api/interappbus/channel/client';
 import {RuntimeInfo} from 'openfin/_v2/api/system/runtime-info';
 
-import {APIFromClientTopic, getServiceChannel, setServiceChannel, getServiceIdentity, setServiceIdentity, APIFromClient, deserializeError, Events, OpenFinChannelConnectionEvent} from './internal';
+import {APIFromClientTopic, getServiceChannel, setServiceChannel, getServiceIdentity, setServiceIdentity, APIFromClient, deserializeError, Events, onReconnect} from './internal';
 import {EventRouter} from './EventRouter';
 
 /**
@@ -47,6 +47,7 @@ export function getEventRouter(): EventRouter<Events> {
 let channelPromise: Promise<ChannelClient> | null = null;
 const hasDOMContentLoaded = new DeferredPromise<void>();
 let hasDisconnectListener = false;
+let reconnect = false;
 
 if (typeof fin !== 'undefined') {
     if (getServiceIdentity().name !== fin.Window.me.name && getServiceIdentity().uuid !== fin.Window.me.uuid) {
@@ -89,10 +90,14 @@ export async function getServicePromise(): Promise<ChannelClient> {
                     channel.setDefaultAction(() => false);
                     if (!hasDisconnectListener) {
                         channel.onDisconnection(() => {
+                            reconnect = true;
                             channelPromise = null;
                             setTimeout(getServicePromise, 300);
                         });
                         hasDisconnectListener = true;
+                    }
+                    if (reconnect) {
+                        onReconnect.emit();
                     }
                     resolve(channel);
                 }
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -13,9 +13,10 @@
  */
 import {EventEmitter} from 'events';
 
+import {DeferredPromise} from 'openfin-service-async';
 import {ChannelClient} from 'openfin/_v2/api/interappbus/channel/client';
 
-import {APIFromClientTopic, SERVICE_CHANNEL, SERVICE_IDENTITY, APIFromClient, deserializeError, Events} from './internal';
+import {APIFromClientTopic, SERVICE_CHANNEL, SERVICE_IDENTITY, APIFromClient, deserializeError, Events, OpenFinChannelConnectionEvent} from './internal';
 import {EventRouter} from './EventRouter';
 
 /**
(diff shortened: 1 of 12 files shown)


bryangaleopenfin:
Is there a Notifications equivalent in progress? SERVICE-793 was meant to cover the stuff generic to both FDC3 and Notifications

test/demo/contextChannels/join.inttest.ts
```diff
@@ -563,3 +563,16 @@ describe('When using a non-directory app', () => {
         }]);
     });
 });
+
+describe.only('When the provider is reloaded', () => {
```
bryangaleopenfin: The `.only` is leftover from testing?

test/demo/utils/common.ts
```diff
@@ -195,3 +195,10 @@ async function isServiceClear(): Promise<boolean> {
         }, testManagerIdentity
     );
 }
+
+export async function reloadProvider(): Promise<void> {
```
bryangaleopenfin: Would move up with the other exported functions

ottah:
@bryangaleOF Yeah that is just the DOMContentLoaded with the deferred promise from this (lines 47-54 connection.ts). (Wont let me link the lines)

src/client/connection.ts
```diff
@@ -43,40 +44,56 @@ export function getEventRouter(): EventRouter<Events> {
...
+                }
+                if (fin.Window.me.uuid === getServiceIdentity().uuid && fin.Window.me.name === getServiceIdentity().name) {
+                    reject(new Error('Trying to connect to provider from provider'));
+                } else {
+                    fin.InterApplicationBus.Channel.connect(getServiceChannel(), {
+                        wait: true,
+                        payload: {version: PACKAGE_VERSION}
+                    }).then((channel: ChannelClient) => {
+                        // @ts-ignore Timestamp channel creation time for debugging
+                        channel['timestamp'] = (new Date()).toUTCString();
```
bryangaleopenfin: Intentionally left in?

src/client/connection.ts
```diff
@@ -43,40 +44,56 @@ export function getEventRouter(): EventRouter<Events> {
...
  * Promise to the channel object that allows us to connect to the client
  */
-let channelPromise: Promise<ChannelClient>|null = null;
+let channelPromise: Promise<ChannelClient> | null = null;
+const hasDOMContentLoaded = new DeferredPromise<void>();
 
-if (typeof fin !== 'undefined') {
-    getServicePromise();
-}
+if (typeof document !== 'undefined') {
```
bryangaleopenfin: When would `document` be undefined, and should we set a rejected value for `channelPromise` in that case?

src/client/connection.ts
```diff
@@ -43,40 +44,56 @@ export function getEventRouter(): EventRouter<Events> {
...
+    document.addEventListener('DOMContentLoaded', () => {
+        hasDOMContentLoaded.resolve();
```
bryangaleopenfin, line 56: Suggest put this beneath the below block, so it's 100% clear that by the time the line `await hasDOMContentLoaded.promise`, `channelPromise` may be set

src/client/main.ts
```diff
@@ -416,16 +416,18 @@ function hasIntentListener(intent: string): boolean {
...
-function deserializeChannelChangedEvent(eventTransport: Transport<ChannelChangedEvent>): ChannelChangedEvent {
-    const type = eventTransport.type;
-    const identity = eventTransport.identity;
-    const channel = eventTransport.channel ? getChannelObject(eventTransport.channel) : null;
-    const previousChannel = eventTransport.previousChannel ? getChannelObject(eventTransport.previousChannel) : null;
+if (typeof fin !== 'undefined') {
+    initialize();
 
-    return {type, identity, channel, previousChannel};
+    fin.InterApplicationBus.Channel.onChannelConnect((event: OpenFinChannelConnectionEvent) => {
```
bryangaleopenfin: Is this going to lead to us doing a double-initialize on startup in most cases?

bryangaleopenfin: WE do almost the exact some thing in `connection`. Worth pulling out a shared function? Even if it's only somrthing that takes an event and returns true/false for whether it's about the service channel

src/client/internal.ts
```diff
@@ -243,6 +243,16 @@ export interface ChannelReceiveContextPayload {
     context: Context;
 }
 
+export interface OpenFinChannelConnectionEvent {
```
bryangaleopenfin: Deeeeefinitely nothing in `@types/openfin` that just gives us this type, or at least something we can extend?

test/demo/raiseIntent/withTarget.inttest.ts
```diff
@@ -176,6 +176,26 @@ listener to be added', async () => {
     });
 });
 
+describe('When reloading the provider', () => {
+    setupOpenDirectoryAppBookends(testAppWithPreregisteredListeners1);
+
+    test('Apps with a preregistered intent listener recieve intents', async () => {
```
bryangaleopenfin: receive

test/demo/raiseIntent/withTarget.inttest.ts
```diff
@@ -176,6 +176,26 @@ listener to be added', async () => {
...
+        await reloadProvider();
+
+        await raiseIntent(preregisteredIntent, testAppWithPreregisteredListeners1);
+        const listener = await fdc3Remote.getRemoteIntentListener(testAppWithPreregisteredListeners1, preregisteredIntent.type);
+        await expect(listener).toHaveReceivedContexts([preregisteredIntent.context]);
+    });
+
+    test('Apps that registered an intent listener recieve intents', async () => {
```
bryangaleopenfin: receive

test/demo/contextListener.inttest.ts
```diff
@@ -191,4 +191,19 @@ window, the listener is triggered exactly once with the correct context', async
...
+
+    describe('When the provider is reloaded', () => {
+        let listener: fdc3Remote.RemoteContextListener;
+        setupOpenDirectoryAppBookends(testAppInDirectory1);
+
+        beforeEach(async () => {
+            listener = await fdc3Remote.addContextListener(testAppInDirectory1);
+        });
+
+        test('Contexts listeners are readded and contexts are recieved', async () => {
```
bryangaleopenfin: received

bryangaleopenfin (commented):
Some surface level stuff - need to stare at `contextChannels.ts` a bit longer

src/client/contextChannels.ts
```diff
@@ -585,7 +615,25 @@ if (typeof fin !== 'undefined') {
...
         eventHandler.registerDeserializer('window-removed', deserializeWindowRemovedEvent);
+        eventHandler.registerDeserializer('channel-changed', deserializeChannelChangedEvent);
+
+        rehydrate();
     }, (reason) => {
         console.warn('Unable to register client channel context handlers. getServicePromise() rejected with reason:', reason);
     });
 }
+
+async function rehydrate(): Promise<void> {
```
bryangaleopenfin, line 618: This will likely not play nice with `getOrCreateAppChannel`. Probably an in person conversation

bryangaleopenfin:
Did my staring at `contextChannels`, looks good, no further feedback

src/client/contextChannels.ts
ottah: This now works with app channels.

src/client/connection.ts
```diff
@@ -43,13 +44,28 @@ export function getEventRouter(): EventRouter<Events> {
 /**
  * Promise to the channel object that allows us to connect to the client
  */
-let channelPromise: Promise<ChannelClient>|null = null;
+let channelPromise: Promise<ChannelClient> | null = null;
+const hasDOMContentLoaded = new DeferredPromise<void>();
 
 if (typeof fin !== 'undefined') {
-    getServicePromise();
+    if (getServiceIdentity().name !== fin.Window.me.name && getServiceIdentity().uuid !== fin.Window.me.uuid) {
```
bryangaleopenfin: Is this safe? We don't know we have a fixed service identity until after `getServicePromise` is called

src/client/internal.ts
bryangaleopenfin: Same comment as for the notifications equivalent:

"Cool. For now, would move into connection.ts, unexported since anywhere else using this is probably doing something wrong. And consider extending WindowEvent<,>, which will give you the last 4 fields here"

src/client/contextChannels.ts
```diff
@@ -554,38 +554,81 @@ function deserializeWindowAddedEvent(eventTransport: Transport<ChannelWindowAdde
...
+    let channelToJoin: DefaultChannel | SystemChannel | AppChannel | null = currentChannel;
+    // Check if the client reloaded and was already in a channel
+    if (!channelToJoin) {
+        const previousChannel = await getCurrentChannel();
+        channelToJoin = previousChannel ? previousChannel : defaultChannel;
```
bryangaleopenfin: Does this work? When would `previousChannel` be false-y?

test/demo/contextChannels/join.inttest.ts
```diff
@@ -563,3 +564,24 @@ describe('When using a non-directory app', () => {
...
+        await expect(fdc3Remote.getCurrentChannel(testAppNotInDirectory1)).resolves.toHaveProperty('channel', greenChannel.channel);
+        await expect(greenChannel.getMembers()).resolves.toEqual([{uuid: testAppNotInDirectory1.uuid, name: testAppNotInDirectory1.name}]);
+        await reloadProvider();
+        await expect(fdc3Remote.getCurrentChannel(testAppNotInDirectory1)).resolves.toHaveProperty('channel', greenChannel.channel);
+        await expect(greenChannel.getMembers()).resolves.toEqual([{uuid: testAppNotInDirectory1.uuid, name: testAppNotInDirectory1.name}]);
+        await reloadProvider();
+        await expect(fdc3Remote.getCurrentChannel(testAppNotInDirectory1)).resolves.toHaveProperty('channel', greenChannel.channel);
+        await expect(greenChannel.getMembers()).resolves.toEqual([{uuid: testAppNotInDirectory1.uuid, name: testAppNotInDirectory1.name}]);
+
+        await delay(1000);
```
bryangaleopenfin: Either comment or use one of the/add a `Duration` value

test/demo/contextChannels/join.inttest.ts
```diff
@@ -563,3 +564,24 @@ describe('When using a non-directory app', () => {
...
+
+        await reloadProvider();
+        await expect(fdc3Remote.getCurrentChannel(testAppNotInDirectory1)).resolves.toHaveProperty('channel', greenChannel.channel);
+        await expect(greenChannel.getMembers()).resolves.toEqual([{uuid: testAppNotInDirectory1.uuid, name: testAppNotInDirectory1.name}]);
+        await reloadProvider();
+        await expect(fdc3Remote.getCurrentChannel(testAppNotInDirectory1)).resolves.toHaveProperty('channel', greenChannel.channel);
+        await expect(greenChannel.getMembers()).resolves.toEqual([{uuid: testAppNotInDirectory1.uuid, name: testAppNotInDirectory1.name}]);
+        await reloadProvider();
+        await expect(fdc3Remote.getCurrentChannel(testAppNotInDirectory1)).resolves.toHaveProperty('channel', greenChannel.channel);
+        await expect(greenChannel.getMembers()).resolves.toEqual([{uuid: testAppNotInDirectory1.uuid, name: testAppNotInDirectory1.name}]);
```
bryangaleopenfin: If this is just doing the same thing 3 times, would put it in a for loop to make that obvious

src/client/contextChannels.ts
```diff
@@ -554,38 +554,81 @@ function deserializeWindowAddedEvent(eventTransport: Transport<ChannelWindowAdde
...
+                    payload.context,
+                    (e) => console.warn(`Error thrown by channel context handler, swallowing error. Error message: ${e.message}`),
+                    () => new Error('All channel context handlers failed')
+                );
+            });
+        } catch (e) {
+            // Trying to resubscribe to the same channel
+        }
+
+        await rehydrate();
```
bryangaleopenfin: Should this be on reconnect, rather than first initialization? If you're client is refreshing, I feel like that's on you to handle

bryangaleopenfin:
Re-review done

test/demo/contextChannels/join.inttest.ts
```diff
@@ -563,3 +563,19 @@ describe('When using a non-directory app', () => {
         }]);
     });
 });
+
+describe('When the provider is reloaded', () => {
```
bryangaleopenfin, line 567: I feel more confused now that it's just the reload repeating, sorry. What do we get from doing the reload 4 times? If that's really what we want, suggest comment at the start of the for loop

src/client/internal.ts
ottah: Switched to using a signal to fire this event as its more reliable and doesn't get lost in the runtime when there is multiple windows trying to connect at the same time.

src/client/connection.ts
ottah: Removed the checks.

src/client/contextChannels.ts
```diff
@@ -554,38 +554,74 @@ function deserializeWindowAddedEvent(eventTransport: Transport<ChannelWindowAdde
...
+            // Trying to resubscribe to the same channel
+        }
     }, (reason) => {
         console.warn('Unable to register client channel context handlers. getServicePromise() rejected with reason:', reason);
     });
 }
+
+async function rehydrate(): Promise<void> {
+    let channelToJoin: Channel = currentChannel || defaultChannel;
+    // Check if the client reloaded and was already in a channel
```
bryangaleopenfin: Does this comment still apply?

src/client/internal.ts
bryangaleopenfin: Cool. Then this can be deleted?

src/client/contextChannels.ts
ottah: Removed

test/demo/contextChannels/join.inttest.ts
ottah: I've added a timeout to the reconnect but without the timeout theres a (random) chance that the that client doesn't reconnect or doesn't register listeners on the channel. Doing it 4 times is just to be sure its working correctly.

bryangaleopenfin (approved):

rypete (approved):

PR #205, opened 2020-01-24
[N/A] Update service tooling

bryangaleopenfin:


hooks.ts
```diff
@@ -1,10 +1,12 @@
-const fs = require('fs');
-const path = require('path');
+import fs = require('fs');
```
tomrobertsOF: I've never seen this syntax before. Does this actually work?

hooks.ts
```diff
@@ -1,10 +1,12 @@
...
+import path = require('path');
 
-module.exports = (app) => {
+import {Hook, registerHook} from 'openfin-service-tooling/utils/allowHook';
+
+registerHook(Hook.TEST_MIDDLEWARE, (app) => {
     // Sneakily return the test directory instead of the default one
     app.get('/provider/sample-app-directory.json', (req, res) => {
-        const testDirectory = JSON.parse(fs.readFileSync(path.join('res', 'test', 'sample-app-directory.json')));
+        const testDirectory = JSON.parse(fs.readFileSync(path.join('res', 'test', 'sample-app-directory.json')).toString());
```
tomrobertsOF, line 9: Why is this `toString` here? Wouldn't that just return `[Object object]` or something like that?

bryangaleopenfin: Without it I get "Argument of type 'Buffer' is not assignable to parameter of type 'string'.". Will double check it's doing what I assume

hooks.ts
bryangaleopenfin: It seems?! Copied from the `os-launchpad` `hooks.ts` file

bryangaleOF: Yep, does what I expected. Which makes it werid that it worked in the first place. I think JSON.parse must specifically check for buffer objects and handle appropriately, but that's not reflected in the types

hooks.ts
bryangaleopenfin: Yep, does what I expected. Which makes it werid that it worked in the first place. I think JSON.parse must specifically check for buffer objects and handle appropriately, but that's not reflected in the types

tomrobertsOF: Hmm, fair enough.

hooks.ts
tomrobertsOF: Hmmm. I really don't like it, but if it works then ok I guess.

tomrobertsOF (approved):

ottah (approved):

PR #215, opened 2020-04-08
Use the test directory in the test provider

ottah:


tomrobertsOF (approved):

res/test/provider.json
```diff
@@ -14,6 +14,6 @@
         "version": "14.78.45.31"
     },
     "serviceConfiguration": {
-        "applicationDirectory": "http://localhost:3923/provider/sample-app-directory.json"
+        "applicationDirectory": "http://localhost:3923/test/sample-app-directory.json"
```
pjbroadbent, line 17: If making this change, the hook that overrides the previous URL should no longer be required - can we remove it?

pjbroadbent (approved):

PR #217, opened 2020-04-17
[SERVICE-989] Fixed paths within manifest/config files

pjbroadbent:
Fixed paths within manifest/config files to point to the correct CDN base location.

I missed this change when bumping to the service-tooling version that included this change: https://github.com/HadoukenIO/service-tooling/pull/127. CDN paths are now constructed using string templating, rather than assuming all projects are versioned (as some, such as os-launchpad, aren't).

### DIFF
--- a/res/provider/app.json
+++ b/res/provider/app.json
@@ -4,7 +4,7 @@
     "startup_app": {
         "name": "FDC3-Service",
         "description": "OpenFin FDC3 Desktop Service",
-        "url": "https://cdn.openfin.co/services/openfin/fdc3/provider.html",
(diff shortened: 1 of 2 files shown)


tomrobertsOF (approved):

pjbroadbent:
Added an extra change, to account for the differences in how the provider window is shown/hidden that was in the same tooling PR: https://github.com/HadoukenIO/service-tooling/pull/127/files#diff-70741c6c6fdaf584a42e5e47970770d1L157-R160

Provider window should be configured to be hidden within the project, and the tooling will override this flag when running locally. Was previously the other way round.

PR #219, opened 2020-04-27
[N/A] Import tweaks

pjbroadbent:
Two small tweaks/fixes:
* Prevented harmless-but-annoying "Trying to connect to provider from provider" errors in provider console, which happens when the provider ends up referencing `connection.ts` in a "Javascript-y" way.
* Fixed the detection of non-OpenFin environments so that calls will reject with error, rather than hang indefinitely. This was because the client was waiting on a `DOMContentLoaded` event, but that listener wasn't being added.

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -45,19 +45,23 @@ export function getEventRouter(): EventRouter<Events> {
  * Promise to the channel object that allows us to connect to the client
  */
 let channelPromise: Promise<ChannelClient> | null = null;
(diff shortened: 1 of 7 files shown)


pjbroadbent:
Turns out the "Trying to connect to provider from provider" error wasn't harmless, it was actually _required_ for the service to work properly... Reworked the `setServiceX` stuff a bit, to work without requiring the self-connection attempt.

ottah (approved):

rypete (approved):

